PR #23, opened 2023-03-07
Add codec

rbair23:
This is a rather large change, far beyond what we started out looking to do. Initially we intended to add the "Codec" type. But we ended up restructuring the classes in `com.hedera.pbj.runtime.io`. The new structure we like very much, and have a large number of tests to go along with it.


pbj-core/gradle.properties
```diff
@@ -1,5 +1,5 @@
 # Version number
-version=0.1.0-SNAPSHOT
+version=0.3.3.11
```
nathanklick, line 2: Not a valid semantic version and will fail on Maven Central and Gradle Plugins release.

jasperpotts (approved):

PR #30, opened 2023-04-10
Replaced `/` with OS-dependent path separator

imalygin:
**Description**:

Due to the fact that the compiler expect a path separator to be `/`,  PBJ gradle task fails on Windows.

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
@@ -193,7 +193,7 @@ public final class Common {
 	 * @return File object for java file
 	 */
 	public static File getJavaFile(File srcDir, String javaPackage, String className) {
-		File packagePath = new File(srcDir.getPath() + File.separatorChar + javaPackage.replaceAll("\\.",File.separator));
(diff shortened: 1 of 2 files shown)


jasperpotts (dismissed):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -193,7 +193,7 @@ public static String removingLeadingDot(String text) {
 	 * @return File object for java file
 	 */
 	public static File getJavaFile(File srcDir, String javaPackage, String className) {
-		File packagePath = new File(srcDir.getPath() + File.separatorChar + javaPackage.replaceAll("\\.",File.separator));
+		File packagePath = new File(srcDir.getPath() + File.separatorChar + javaPackage.replaceAll("\\.","\\" + File.separator));
```
jasperpotts, line 196: I don't see why this has "\\"+File.seperator

imalygin: The second argument in `replaceAll` is also a regex. In case of `/` it works just fine, but if it's `\` the method call fails with `IllegalArgumentException` because it interprets it as a missing character that was supposed to be escaped.

jasperpotts (approved):

Issue #35, opened 2023-04-28
Update the `hashCode` function to avoid collisions


Neeharika-Sompalli:
Change generated record files for messages with a single number field to not use default generated dashcode() function but use the one below. 
This is done for performance reasons so that we have better balanced hashing buckets.

```
public static long perm64(long x) {
        // Shifts: {30, 27, 16, 20, 5, 18, 10, 24, 30}
        x += x << 30;
        x ^= x >>> 27;
        x += x << 16;
        x ^= x >>> 20;
        x += x << 5;
        x ^= x >>> 18;
        x += x << 10;
        x ^= x >>> 24;
        x += x << 30;
        return x;
    }

```

LLITCHEV:
PR https://github.com/hashgraph/pbj/pull/38

PR #38, opened 2023-05-09
Add better hashCode() function for single int field objects.

jasperpotts:
Work on issue #35 to add advanced hash code function to generated code.

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
@@ -255,6 +255,37 @@ public final class ModelGenerator implements Generator {
 			).replaceAll("\n","\n"+FIELD_INDENT);
 		}
 
+		// Add here hashCode() for object with a single int field.
+		if (fields.size() == 1) {
+			FieldType fieldType = fields.get(0).type();
+			switch (fieldType) {
+				case ENUM, INT32, UINT32, SINT32, FIXED32, SFIXED32 -> {
+					bodyContent += """
+					\n    /**
+					     * Override the default hashCode method for 
+					     * single field int objects.
+					     */\n""";
+					bodyContent += "    @Override public int hashCode() {\n" +
+							"            // Shifts: {30, 27, 16, 20, 5, 18, 10, 24, 30}\n" +
+							"            int x = " + fields.get(0).name() + ";\n" +
+							"            x += x << 30;\n" +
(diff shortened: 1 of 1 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -255,6 +255,37 @@ public void generate(final Protobuf3Parser.MessageDefContext msgDef,
 			).replaceAll("\n","\n"+FIELD_INDENT);
 		}
 
+		// Add here hashCode() for object with a single int field.
+		if (fields.size() == 1) {
+			FieldType fieldType = fields.get(0).type();
+			switch (fieldType) {
+				case ENUM, INT32, UINT32, SINT32, FIXED32, SFIXED32 -> {
```
jasperpotts, line 262: I don't think we need this for enums. We do need to add it for 64bit integers as well though.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -255,6 +255,37 @@ public void generate(final Protobuf3Parser.MessageDefContext msgDef,
...
+							"            x ^= x >>> 27;\n" +
+							"            x += x << 16;\n" +
+							"            x ^= x >>> 20;\n" +
+							"            x += x << 5;\n" +
+							"            x ^= x >>> 18;\n" +
+							"            x += x << 10;\n" +
+							"            x ^= x >>> 24;\n" +
+							"            x += x << 30;\n" +
+							"            return x;\n" +
+							"        }\n\n";
```
jasperpotts, line 281: Think I prefer staying with multi-line block strings with variable substitution using replace() like we have everywhere else. Mainly just for constancy. There is still some old style with string formatting but we are almost all moved over to blocks with "$abc" style variables with replace()

LLITCHEV: Addressed.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
LLITCHEV: Addressed.

LLITCHEV (approved):
Approving this per talk with Jasper. I need to do this since he created the PR.

Issue #36, opened 2023-05-04
Let Bytes be Comparable


rbair23:
### Problem

I have the need to sort Bytes objects. I don't want to have to convert them to byte arrays.

### Solution

 It would be nice if they were either `Comparable`, or there were some built in `Comparators` that could sort. One sorting strategy would be shortest to longest, and lowest bytes to highest bytes. Such that:

```
{ 7 }
{ 1, 2 }
{ 1, 2, 3 }
{ 2, 1, 1 }
{ 5, 6, 7, 8 }
{ 0, 1, 0, 1, 0 }
```

In the above example, we can see that shorter rows are sorted higher than longer rows, and lower bytes are sorted before higher bytes when given the same length.

rbair23:
If we do this, there is some code in `TransactionChecker` in services on the consensus node that could be improved.

jasperpotts:
@rbair23 This feels odd, you would have expected default sort order on value not just length?

rbair23:
I'm sure there are several interesting ways to sort (in my case, I want length and then by order). Rather than Bytes being comparable itself, we could just have Bytes have some built in comparators. You could have SORT_BY_LENGTH and SORT_BY_VALUE and then combine them Comparator.thenComparing, like `SORT_BY_LENGTH.thenComparing(SORT_BY_VALUE)`. That way we don't have to pick just one way.

LLITCHEV:
Fixed with https://github.com/hashgraph/pbj/pull/41

PR #41, opened 2023-05-10
Make the Bytes object Comparable.

LLITCHEV:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


LLITCHEV:
This resolve Issue #41

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -353,4 +353,31 @@ private void validateOffset(long offset) {
...
+            InputStream thisBytes = this.toInputStream();
+            InputStream otherBytes = other.toInputStream();
+            for (int i = 0; i < lThis; i++) {
+                int thisVal = thisBytes.read();
+                int otherVal = otherBytes.read();
+                if (thisVal != otherVal) {
+                    return thisVal > otherVal ? 1 : -1;
+                }
+            }
+        }
```
jasperpotts, line 376: I am not sure about the secondary sort order. We should check with @rbair23

LLITCHEV: Yes! I thought comparing bytes will be fine, since it sounds like it is important to get the order right. But, yes asking for the thoughts of @rbair23 is a great idea.

rbair23: I added some comment to the issue https://github.com/hashgraph/pbj/issues/36#issuecomment-1546363015. Reposting here:

>I'm sure there are several interesting ways to sort (in my case, I want length and then by order). Rather than Bytes being comparable itself, we could just have Bytes have some built in comparators. You could have SORT_BY_LENGTH and SORT_BY_VALUE and then combine them Comparator.thenComparing, like SORT_BY_LENGTH.thenComparing(SORT_BY_VALUE). That way we don't have to pick just one way.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -353,4 +353,31 @@ private void validateOffset(long offset) {
...
+        try {
+            InputStream thisBytes = this.toInputStream();
+            InputStream otherBytes = other.toInputStream();
+            for (int i = 0; i < lThis; i++) {
+                int thisVal = thisBytes.read();
+                int otherVal = otherBytes.read();
+                if (thisVal != otherVal) {
+                    return thisVal > otherVal ? 1 : -1;
+                }
+            }
```
rbair23, line 375: Wouldn't walking over the byte arrays be more efficient?

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/BytesTest.java
```diff
@@ -407,4 +407,48 @@ void matchesPrefixBytesTest() {
...
+        assertEquals(1, new Bytes.SORT_BY_VALUE().compare(bytes1, bytes2));
+        bytes1 = Bytes.wrap(new byte[]{1, 2});
+        bytes2 = Bytes.wrap(new byte[]{7, 1});
+        assertEquals(0, new Bytes.SORT_BY_LENGTH().compare(bytes1, bytes2));
+        assertEquals(-1, new Bytes.SORT_BY_VALUE().compare(bytes1, bytes2));
+        bytes1 = Bytes.wrap(new byte[]{7, 2});
+        bytes2 = Bytes.wrap(new byte[]{7, 2});
+        assertEquals(0, new Bytes.SORT_BY_LENGTH().compare(bytes1, bytes2));
+        assertEquals(0, new Bytes.SORT_BY_VALUE().compare(bytes1, bytes2));
+    }
```
rbair23, line 453: A great way to improve this code would be to break it into two parameterized tests (well, three tests if we have SORT_BY_LENGTH, SORT_BY_SIGNED_VALUE, SORT_BY_UNSIGNED_VALUE). JUnit5 has some great ways to do parameterized tests.

These below examples are just examples. They compile but not all the test pass (probably because the test is wrong).

This first example shows using the CSV Source parameterized test. It is easy to use when converting to certain primitive types:
```
    @ParameterizedTest
    @CsvSource(textBlock = """
            "", "", 0
            "a", "", 1
            "", "a", -1
            "a", "a", 0
            "aa", "a", 1
            "a", "aa", -1
            "ab", "ba", 0
            "abc", "cab", 0
            """)
    @DisplayName("Comparing Bytes by length")
    void compareByLength(String arr1, String arr2, int expected) {
        var bytes1 = Bytes.wrap(arr1);
        var bytes2 = Bytes.wrap(arr2);
        assertEquals(expected, Bytes.SORT_BY_LENGTH.compare(bytes1, bytes2));
    }
```

These next examples show using `MethodSource` style. Both CSVSource and MethodSource can be found online with lots of docs on how to use them:

```
    @ParameterizedTest
    @MethodSource
    @DisplayName("Comparing Bytes by signed value")
    void compareBySignedBytes(byte[] arr1, byte[] arr2, int expected) {
        var bytes1 = Bytes.wrap(arr1);
        var bytes2 = Bytes.wrap(arr2);
        assertEquals(expected, Bytes.SORT_BY_SIGNED_VALUE.compare(bytes1, bytes2));
    }

    static Stream<Arguments> compareBySignedBytes() {
        return Stream.of(
                Arguments.of(new byte[0], new byte[0], 0),
                Arguments.of(new byte[0], new byte[]{1}, -1),
                Arguments.of(new byte[]{1}, new byte[0], 1),
                Arguments.of(new byte[]{1}, new byte[]{2}, 1),
                Arguments.of(new byte[]{2}, new byte[]{1}, -1),
                Arguments.of(new byte[]{-1}, new byte[]{2}, 1),
                Arguments.of(new byte[]{2}, new byte[]{-1}, -1),
                Arguments.of(new byte[]{-1}, new byte[]{-2}, -1),
                Arguments.of(new byte[]{-2}, new byte[]{-1}, 1)
        );
    }

    @ParameterizedTest
    @MethodSource
    @DisplayName("Comparing Bytes by unsignedValue")
    void compareByUnsignedBytes(byte[] arr1, byte[] arr2, int expected) {
        var bytes1 = Bytes.wrap(arr1);
        var bytes2 = Bytes.wrap(arr2);
        assertEquals(expected, Bytes.SORT_BY_UNSIGNED_VALUE.compare(bytes1, bytes2));
    }

    static Stream<Arguments> compareByUnsignedBytes() {
        return Stream.of(
                Arguments.of(new byte[0], new byte[0], 0),
                Arguments.of(new byte[0], new byte[]{1}, -1),
                Arguments.of(new byte[]{1}, new byte[0], 1),
                Arguments.of(new byte[]{1}, new byte[]{2}, 1),
                Arguments.of(new byte[]{2}, new byte[]{1}, -1),
                Arguments.of(new byte[]{-1}, new byte[]{2}, 1),
                Arguments.of(new byte[]{2}, new byte[]{-1}, -1),
                Arguments.of(new byte[]{-1}, new byte[]{-2}, -1),
                Arguments.of(new byte[]{-2}, new byte[]{-1}, 1)
        );
    }
```

What is nice about using parameterized tests is that they all run individually and those that pass and fail are listed separately.

rbair23 (changes requested):

LLITCHEV:
@rbair23 All issues are supposed to be addressed. Thanks for the review!

rbair23 (approved):

PR #46, opened 2023-05-18
Added the toByteArray() API.

LLITCHEV:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


LLITCHEV:
@rbair23 The wrap method can take a null byte[]. I presumed this is OK since the parameter is tagged with @NonNull. The constructur later throws NPE (line 44) if one tries to do that. So, I coded the method as the buffer array is never null. Just making sure that is the intention.

jasperpotts (approved):

PR #47, opened 2023-05-19
Throw a new exception when reading empty stream.

LLITCHEV:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/EOFException.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/EOFException.java
@@ -2,9 +2,7 @@ package com.hedera.pbj.runtime.io.stream;
 
 import java.nio.BufferUnderflowException;
 
-/** This class is used as an exception to signal that the end of stream is reached
- * when reading.
- */
+/**  This class is used as an exception to signal that the end of stream is reached when reading.  */
 public class EOFException  extends BufferUnderflowException {
     private static final long serialVersionUID = 1799983599892333203L;
 }
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/EOFException.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/EOFException.java
@@ -1,8 +1,10 @@
 package com.hedera.pbj.runtime.io.stream;
 
+import java.nio.BufferUnderflowException;
+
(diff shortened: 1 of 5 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/EOFException.java
```diff
@@ -0,0 +1,10 @@
+package com.hedera.pbj.runtime.io.stream;
+
+import java.nio.BufferUnderflowException;
+
+/** This class is used as an exception to signal that the end of stream is reached
+ * when reading.
+ */
```
jasperpotts, line 7: minor ;-) would be nice to be one of these styles
```
/** 
  * This class is used as an exception to signal that the end of stream is reached
  * when reading.
  */
```
or
```
/**  This class is used as an exception to signal that the end of stream is reached when reading.  */
```

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingDataTest.java
```diff
@@ -258,6 +258,25 @@ public void close() throws IOException {
...
+        final var inputStream = new ByteArrayInputStream(new byte[0]) {
+            @Override
+            public void close() throws IOException {
+                throw new IOException("Failed");
+            }
+        };
+
+        byte[] read = new byte[5];
+        final var stream = new ReadableStreamingData(inputStream);
+        try {
```
jasperpotts, line 273: might be nice like 
```
assertThrows(EOFException.class, () -> {
      stream.readBytes(read);
});
```

jasperpotts (commented):
All good just a couple minor improvements

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingDataTest.java
LLITCHEV: Changed.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/EOFException.java
LLITCHEV: Changed the comment. It is less than 120 chars.

jasperpotts (approved):

PR #57, opened 2023-06-07
Fix all VarInt benchmarks.

LLITCHEV:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -556,4 +556,106 @@ private static Comparator<Bytes> valueSorter(@NonNull final Comparator<Byte> byt
...
+                if (buffer[tempPos++] < 0L) {
+                    return readVarIntLongSlow(pos, zigZag);
+                }
+            }
+        }
+        pos[0] = (int)tempPos;
+        return zigZag ? (x >>> 1) ^ -(x & 1) : x;
+    }
+
+    long readVarIntLongSlow(int[] posIn, final boolean zigZag) {
```
jasperpotts, line 648: This should not be committed I think into api

jasperpotts (changes requested):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -556,4 +556,106 @@ private static Comparator<Bytes> valueSorter(@NonNull final Comparator<Byte> byt
             return ((len > 0) ? 1 : -1);
         };
     }
+
+    public int readVarInt(int[] pos, final boolean zigZag) {
```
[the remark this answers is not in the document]
LLITCHEV: Yes, I just noticed that. It is different that ReadableSequentialData interface (readXXXXX() methods. Fixed.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
LLITCHEV: Yes! Fixed. Again, getting ready for the next PR. I'll cross that bridge when I get to it.

jasperpotts (approved):

jasperpotts (approved):

PR #69, opened 2023-06-29
Some optimizations of Bytes.

### LINKED ISSUE #59: Improve Bytes Performance
### Problem

Bytes is slow using the default read methods from the RandomAccessData interface.

### Solution

Implement all RandomAccessData methods directly in Bytes with direct access to array data. Either with normal array access or Misc.Unsafe if it is safe enough and much faster. Creating fast path implementations for all RandomAccessData methods.

LLITCHEV:
This optimizations get us about 4-5% speadup. (Enables us to do optimizations later as well.

The Bytes object is now a ReadableSequentialData object as well. Doing this optimization allows us to have private methods, instead of doing interface calls. That makes the Jit inline a lot of the code and gets us some perf.

**Description**:

**Related issue(s)**:

Fixes #59 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ReadableSequentialData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ReadableSequentialData.java
@@ -235,7 +235,7 @@ public interface ReadableSequentialData extends SequentialData {
             throw new BufferUnderflowException();
         }
 
-        return Bytes.wrap(bytes).toReadableSequentialData();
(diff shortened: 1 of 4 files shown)


jasperpotts (changes requested):
There is a key point missed here. There was a reason that Bytes did not implement the other interfaces. They have methods based on a position and limit state. For those to work bytes can not be immutable. The whole design reason for Bytes existing is it is immutable!

LLITCHEV:
New approach taken to address issues stated here. The PR that addresses this is #89

Issue #90, opened 2023-08-26
BufferedData slices are still broken


artemananiev:
The fix for https://github.com/hashgraph/pbj/issues/83 is incomplete. I use `BufferedData.readVarInt()` in MerkleDb, it was touched in that fix, but some pieces are still missing.

In particular, to read bytes BufferedData now takes wrapped buffer offset into consideration:
```java
        int tempPos = buffer.position() + buffer.arrayOffset();
```

However, when it needs to forward buffer position later, the offset is ignored:
```java
            buffer.position(tempPos);
```

Instead, it should be
```java
            buffer.position(tempPos - buffer.arrayOffset());
```

There may be more places broken like this.

Issue #83, opened 2023-08-11
BufferedData.view always reads from the start of the wrapped buffer


rbair23:
### Description

BufferedDate.view does not read the appropriate bytes from the underlying buffer. The problem is that in any method where BufferedData directly reads the underlying ByteBuffer array, it must *also* read the arrayOffset and use it. It does not do this correctly.

### Steps to reproduce

final var buf = BufferedData.wrap(new byte[] { 1, 2, 3, 4 });
buf.position(1);
buf.view(2);

assertThat(buf.getBytes()).isEqualTo(2); // <-- this test will fail. It will return 1 instead.

### Hedera network

other

### Version

v0.7.1

### Operating system

macOS

PR #91, opened 2023-08-28
90: BufferedData slices are still broken

artemananiev:
A follow-up fix for https://github.com/hashgraph/pbj/issues/83. In complex `BufferedData` methods like `readVarInt()` or `readVarLong()`, which are more than just propagate to the underlying buffer, reading takes array offset (for sliced buffers) into consideration after the fix above. However, when position is updated to skip the bytes read, the offset is still ignored. This PR changes that.

A couple new unit tests are provided.

Fixes: https://github.com/hashgraph/pbj/issues/90

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/BufferedData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/BufferedData.java
@@ -1,5 +1,7 @@
 package com.hedera.pbj.runtime.io.buffer;
 
+import static java.nio.ByteOrder.BIG_ENDIAN;
+
 import com.hedera.pbj.runtime.io.DataAccessException;
 import com.hedera.pbj.runtime.io.ReadableSequentialData;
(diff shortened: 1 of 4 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/BufferedData.java
```diff
@@ -664,20 +665,21 @@ public double readDouble(@NonNull final ByteOrder byteOrder) {
...
+        int lastPos = buffer.limit() + arrayOffset;
 
         if (lastPos == tempPos) {
             throw new BufferUnderflowException();
         }
 
         int x;
         if ((x = buff[tempPos++]) >= 0) {
-            buffer.position(tempPos);
+            buffer.position(tempPos - arrayOffset);
```
artemananiev, line 679: This is the core of the fix. Same change is in `readVarLong()`

rbair23 (approved):

imalygin (approved):

PR #92, opened 2023-09-05
Gradle updates: use module dependencies / cleanup config / update version

jjohannes:
**Description**:
Following similar setup as done for: https://github.com/hashgraph/hedera-services

In particular, this fixes the ANTLR transitive dependency declaration, so that the `pbj-runtime` now only depends on the `antlr4-runtime` (and not ANTLR itself). This is also important to no longer pull unnecessary ANTLR Jars into the hedera runtime.


jasperpotts (approved):
Looks good

PR #98, opened 2023-09-07
Change the Bytes.ytoString¸sxdef4p[]

### LINKED ISSUE #97: Change Bytes toString to just output HEX
### Problem

When `Bytes.toString()` is called it outputs like `Bytes[0,0,255]` which gets really long fast and hard to read.

### Solution

Change `toString()` output to just hex so above would be `0000FF` which is much nicer to read in logs etc.

### Alternatives

_No response_

LLITCHEV:
**Description**:

**Related issue(s)**:

Fixes #97 

**Notes for reviewer**:

**Checklist**


LLITCHEV:
Duplicate of #99. I don't know how this happened. Leave the #99 only.

PR #99, opened 2023-09-07
Change the Bytes.toString to match the required format by #97

### LINKED ISSUE #97: Change Bytes toString to just output HEX

LLITCHEV:
**Description**:
<!--
Changes the toString string to be formatted as per #97 
Example:
For buffer with bytes [0, 0, 0xFF] teh print is:
Instead of Bytes[0, 0, 255]
to be 00FF

**Related issue(s)**:

Fixes #97


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -334,13 +334,10 @@ public int read() throws IOException {
     public String toString() {
         // build string
         StringBuilder sb = new StringBuilder();
-        sb.append("Bytes[");
         for (long i = 0; i < length(); i++) {
```
jasperpotts, line 337: Feels like this could be nicer to use HexFormat.
```
return hexFormat.of().formatHex(buffer, pos, length)
```

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/BytesTest.java
```diff
@@ -44,7 +44,7 @@ void emptyBytesThrowsOnGetByte() {
     @Test
     @DisplayName("Empty bytes String is not null")
     void toStringWorks() {
-        assertEquals("Bytes[]", Bytes.EMPTY.toString());
+        assertEquals("", Bytes.EMPTY.toString());
```
jasperpotts, line 47: Maybe add a unit tests for some other non empty values

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -334,13 +334,10 @@ public int read() throws IOException {
...
             int v = getByte(i) & 0xFF;
-            sb.append(v);
-            if (i < (length()-1)) sb.append(',');
+            sb.append(Integer.toHexString(v).toUpperCase());
         }
-        sb.append(']');
         return sb.toString();
```
jasperpotts, line 341: I am not sure on the raw hex format, here are some other options.
1. `Bytes[255, 255, 255, 255]`  - What we have today that I thing is too long winded
2. `FFFF` - just plain HEX
3. `0xFFFF`  - not sure where this style is from
4. `[FFFF]` in square brackets
5. `Bytes[FFFF]` with class name

Think I like (3) best, what do you think?

jasperpotts (changes requested):

jasperpotts (approved):

PR #100, opened 2023-09-08
Fix for  issue 96.

### LINKED ISSUE #73: ProtoWriterTools methods throw IOException for no reason
Many methods in `ProtoWriterTools` throw IO exceptions for no clear reason, e.g. `writeLong()` It results in depending classes / methods like `Codec.write()` to throw checked exceptions, too.

In my MerkleDb prototype, there are methods to serialize / deserialize objects to / from `WritableSequentialData` / `ReadableSequentialData`. These methods don't throw any checked exceptions. When I implemented these method for `OnDiskKey` and `OnDiskValue` classes, I found that the corresponding Codec method throw redundant exceptions. For now, I just try / catch / rethrow runtime exceptions, but in the future it would be great if codecs didn't throw anything.

### LINKED ISSUE #96: `parseStrict()` not throwing given partial message
### Description

The attached file is a partial serialization of a [`ServicesConfigurationList`](https://github.com/hashgraph/hedera-protobufs/blob/f36e05bd6bf3f572707ca9bb338f5ad6421a4241/services/basic_types.proto#L1458) message. When parsed using a `protoc`-generated parser, an exception is thrown. No exception is thrown when parsed using a PBJ-generated parser. I.e., this test does not pass:

```
    @Test
    void parseStrictNotThrowingOnIncompleteMessage() throws IOException {
        final var hexed = Files.readString(Paths.get("<path-to-file>")).trim();
        final var unhexed = CommonUtils.unhex(hexed);
        assertThrows(InvalidProtocolBufferException.class, () ->
                com.hederahashgraph.api.proto.java.ServicesConfigurationList.parseFrom(unhexed));
        assertThrows(Exception.class, () ->
                com.hedera.hapi.node.base.ServicesConfigurationList.PROTOBUF.parseStrict(
                        BufferedData.wrap(unhexed)));
    }
```

[hexedPartialServicesConfigurationList.txt](https://github.com/hashgraph/pbj/files/12541525/hexedPartialServicesConfigurationList.txt)

### Steps to reproduce

As above.

### Additional context

_No response_

### Hedera network

other

### Version

`develop`

### Operating system

None

LLITCHEV:
Added tests for all the types.

**Related issue(s)**:

Fixes #96 
Fixes #73


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -216,8 +217,11 @@ public static double readDouble(final ReadableSequentialData input) throws IOExc
      */
     public static String readString(final ReadableSequentialData input) throws IOException {
         final int length = input.readVarInt(false);
+        if (input.remaining() < length) {
+            throw new BufferOverflowException();
+        }
         byte[] bytes = new byte[length];
-        input.readBytes(bytes);
+        final var read = input.readBytes(bytes);
```
jasperpotts, line 224: We need to check the read here, it should match length otherwise we need to throw exception.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -216,8 +217,11 @@ public static double readDouble(final ReadableSequentialData input) throws IOExc
      */
     public static String readString(final ReadableSequentialData input) throws IOException {
         final int length = input.readVarInt(false);
+        if (input.remaining() < length) {
```
jasperpotts, line 220: I think this check is not needed as it as `readBytes` will only read remaining bytes and return a read number smaller than length which will be caught by second `if`

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/Issue96Test.java
```diff
@@ -0,0 +1,281 @@
...
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.hedera.pbj.test.proto.pbj.Everything;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import java.util.HexFormat;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+final class Issue96Test {
```
jasperpotts, line 15: Would be nice to give this a name that includes what issue 96 is about like TruncatedDataTests as issue 96 will not mean anything when coming back and looking at the code in the future.

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/Issue96Test.java
```diff
@@ -0,0 +1,281 @@
...
+    void issue96() {
+        final var unhexed = HexFormat.of().parseHex(testData);
+        assertThrows(InvalidProtocolBufferException.class, () ->
+                com.hederahashgraph.api.proto.java.ServicesConfigurationList.parseFrom(unhexed));
+        assertThrows(Exception.class, () ->
+                com.hedera.hapi.node.base.ServicesConfigurationList.PROTOBUF.parseStrict(
+                        BufferedData.wrap(unhexed)));
+
+    }
+
```
jasperpotts, line 30: I wonder if it is worth making ALL these tests parameterized with a integer passed in for truncationLength with values of `[-1,-2,-3,-4,-5,-6,-7,-8]` Just to make sure a random truncation will throw exception.

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/Issue96Test.java
```diff
@@ -0,0 +1,281 @@
...
+        seq.position(0);
+        seq.limit(7);
+
+        // When we parse that sequence, then we fail because there are not enough bytes to continue parsing.
+        assertThrows(Exception.class, () -> ProtoParserTools.readSignedInt64(seq));
+    }
+
+    @Test
+    @DisplayName("Test readFixedInt64")
+    void testFixedInt364() {
```
jasperpotts, line 270: typo testFixedInt64 ?

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/Issue96Test.java
```diff
@@ -0,0 +1,281 @@
...
+        Bytes b = bytes.slice(0, bytes.length() - 1);
+        assertThrows(Exception.class, () ->
+                Everything.PROTOBUF.parseStrict(
+                        BufferedData.wrap(b.toByteArray())));
+    }
+
+    @Test
+    @DisplayName("Test Number Int64")
+    void testNumberInt64() {
+        Everything ev = Everything.newBuilder().int32Number(12345678).build();
```
jasperpotts, line 67: It feels like it would be nice to add more tests of this style for all other field types possible. To make sure they are all good with various truncation lengths.

jasperpotts (changes requested):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
LLITCHEV: Actually, good catch. I had it implemented that way and decided later to do the check before reading. This way it will fail faster - and potentially work faster (same as all the rest implementations are done in BufferedData. It is faster too, to not have to read at all and just fail.
The check is done in the pre-read if.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
LLITCHEV: See the explanation above. I changed it to be a pre-read check and fast fail. This way the code runs faster too and is consistent with how exceptions are thrown in BufferedData and the rest of ReadableSequentialData.

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/Issue96Test.java
LLITCHEV: Yes... I'm really bad with names. It just started as a bug test and grown into more. Changed the name as suggested.

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/Issue96Test.java
LLITCHEV: Yes, I started with the Everything tests and had more, but then after debugging it turned out they all go through ProtoParserTools. All the types are handled by reads actually made in this class, so I thought that should cover everything. I can certainly add them back in. Just there was no real value added. Just let me know...

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/Issue96Test.java
LLITCHEV: Fixed.

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/Issue96Test.java
LLITCHEV: Added parameterized tests.

jasperpotts (changes requested):
Think we still need to check at the end of parsing a sub-object that all the expected bytes were read. Anywhere there is length encoded data with a length, we should enforce we never read more or less bytes than specified in length.

LLITCHEV:
@jasperpotts I aded the extra check. I added the check to make sure we have enough limit (which is a fast fail - without having the actual read). It does make sense to do the check after the read as well, so added code there as well. Thanks!

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -216,6 +217,9 @@ public static double readDouble(final ReadableSequentialData input) throws IOExc
...
+            throw new BufferOverflowException();
+        }
         byte[] bytes = new byte[length];
         input.readBytes(bytes);
```
jasperpotts, line 224: I think we need to check number of bytes read from readBytes as well. `(input.remaining() < length)` might well be true because we are reading from a stream where remaining is infinite but we still did not read all bytes. We can never safely ignore number of bytes read when using streams.

jasperpotts:
We also need to check that all cases are covered. In protobuf there are a few core data types, fixed length, variant and length encoded. All length encoded data we need to make sure all of the specified length bytes are read or throw exception. Otherwise we can start processing the remaining data as a different field and get corrupt data. Length encoded includes, strings, bytes and sub-messages as maybe others.

jasperpotts:
We should also check anywhere we do a bulk read of bytes with `int getBytes(byteArray)` style method that we check the returned number of bytes read matches expected. We have choice to loop and keep reading till all bytes are read or throw exception.

artemananiev (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -227,11 +224,15 @@ public static String readString(final ReadableSequentialData input) throws IOExc
      * @param input the input to read from
      * @return read Bytes object, this can be a copy or a direct reference to inputs data. So it has same life span
      * of InputData
-     * @throws IOException If there was a problem reading
      */
-    public static Bytes readBytes(final ReadableSequentialData input) throws IOException {
+    public static Bytes readBytes(final ReadableSequentialData input) {
         final int length = input.readVarInt(false);
-        return input.readBytes(length);
+        validateInput(input, length);
```
jasperpotts, line 230: Is it really worth a method call for 2 copies of one if? Or is it for testing? If for testing then maybe makes sense and add java doc to validate method

jasperpotts (changes requested):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
imalygin: Initially this method was intended to be used in multiple places but it turned out to be unnecessary. So, I inlined it as you suggested.

jasperpotts (approved):

artemananiev (approved):

PR #101, opened 2023-09-08
Fix issue 87.

### LINKED ISSUE #87: Two comments at top of file cause warning
### Description

When .proto file has two header comments, warnings are emitted when generating pbj source.

See [primitives.proto](https://github.com/hashgraph/hedera-protobufs/blob/b22750f23c4b4974caa33211a2120a2a9582ea49/services/state/primitives.proto) for an example which tickles this bug.

### Steps to reproduce

When running `gradlew generatePbjSource` from hedera-services, I see this warning repeated twice:
`line 27:0 no viable alternative at input '/**\n * This proto file contains primitive value messages.\n * These are intended only for situations where the entire value to be stored in state is a single\n * primitive.  These should never be used as components of another message; use the protobuf\n * type instead.\n */option'`

### Additional context

_No response_

### Hedera network

other

### Version

v0.7.1

### Operating system

None

LLITCHEV:
Couple of fixes to the grammar.

Made a couple of changes in the grammar to fix issues. Allowed DOC_COMMENTS on top level of the file.

The issue here was not an error, but a warning. The runtime was still working properly.

Fixes #87

### DIFF
--- a/pbj-core/pbj-compiler/src/main/antlr/com/hedera/hashgraph/protoparser/grammar/Protobuf3.g4
+++ b/pbj-core/pbj-compiler/src/main/antlr/com/hedera/hashgraph/protoparser/grammar/Protobuf3.g4
@@ -23,6 +23,7 @@ proto
       | topLevelDef
       | emptyStatement_
       | optionComment
+      | DOC_COMMENT
     )*
   ;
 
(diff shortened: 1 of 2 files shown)


jasperpotts (approved):
Looks Good

PR #102, opened 2023-09-19
Implement stable hashCode for records.

### LINKED ISSUE #95: Generated PBJ objects should have proper hashCode() implementation
I assume this issue belongs to PBJ project, not hedera-services.

Generated services objects like `AccountID` don't have `hashCode()` method implemented. It means, default JVM-specific implementation is used. Not only it makes generated records unusable in Java collections, but it will also lead to troubles when such objects (wrapped into `OnDiskKey`, which forward hash code requests to wrapped records) are saved on disk by MerkleDb.

This ticket is to provide reasonable implementation of `hashCode()` for all generated records. This implementation must not change over time. For example, when a new field is added, it must only be accounted in hash codes if its value is non-default.

LLITCHEV:
Fixes #95


jasperpotts:
Indenting for `equals()` is messed up
```
       /**
 * Override the default equals method for
 */
@Override
public boolean equals(Object that) {
	if (that == null || this.getClass() != that.getClass()) {
		return false;
	}

	Signature thatObj = (Signature)that;

        if (signature == null && thatObj.signature != null) {
             return false;
         }
         if (signature != null && !signature.equals(thatObj.signature)) {
	return false;
}
                   return true;
    }
```
It could also be simplified for this case in `Signature.java`
```
       /**
        * Override the default equals method for
        */
       @Override
       public boolean equals(Object that) {
	      if (that == null || this.getClass() != that.getClass()) {
		      return false;
	      }

	      Signature thatObj = (Signature)that;
              if (signature == null) {
                   return thatObj.signature == null;
              } else {
	           return signature.equals(thatObj.signature);
             }
    }
```

jasperpotts:
Looking at generated `Setting.java` the `hashCode()` method is missing. Also equals formatting is a mess:
```
          /**
 * Override the default equals method for
 */
@Override
public boolean equals(Object that) {
	if (that == null || this.getClass() != that.getClass()) {
		return false;
	}

	Setting thatObj = (Setting)that;

        if (name == null && thatObj.name != null) {
             return false;
         }
         if (name != null && !name.equals(thatObj.name)) {
	return false;
}
if (value == null && thatObj.value != null) {
             return false;
         }
         if (value != null && !value.equals(thatObj.value)) {
	return false;
}
if (data == null && thatObj.data != null) {
             return false;
         }
         if (data != null && !data.equals(thatObj.data)) {
	return false;
}
                   return true;
    }
```
Also can be simplified similar to above comment.

jasperpotts:
We have put a lot of effort into the code generated being nicely formatted and commented just like it was hand written. It makes it super nice to read when debugging on services code that uses these classes.

The comment for equals() method is missing a conclusion, end with "for". Also would be nice for the bonnet to say why `equals()` is overridden so that it logically matches `hashCode()`
```
           /**
 * Override the default equals method for
 */
```

jasperpotts:
`hashCode()` needs to ignore fields whose value is default value. If it includes them in hash then protobuf's way of adding new fields does not work and all objects schemas will not be extendable. You need to be able to add a new field to protobuf without effecting the hashCode or equals methods of existing objects. Otherwise you store an object in a Map then come back later to fetch it after a new field has been added and the Map will tell you that object is not contained in the map. Because the hashCode value will have changed while the object is the same object.

LLITCHEV:
@jasperpotts As we discussed, I need to return only if there is false equals. Otherwise it should fallthrough the rest of the fields. That is why the ifs are a bit more involved.
Fixed the indenting - thanks for letting me know how it works exactly in this project.
Added a benchmark to measure the JDK's hashCode() and equals() implementation and compare it with the stable implementation we have.
No differences in perf detected.
JDK...
Benchmark                           Mode  Cnt   Score    Error  Units
EqualsHashCodeBench.benchEquals     avgt    5  ≈ 10⁻³           ns/op
EqualsHashCodeBench.benchHashCode   avgt    avgt    5  ≈ 10⁻³           ns/op
EqualsHashCodeBench.benchNotEquals  avgt    5   0.001 ±  0.001  ns/op

Our stable implementation...
Benchmark                           Mode  Cnt   Score    Error  Units
EqualsHashCodeBench.benchEquals     avgt    5  ≈ 10⁻³           ns/op
EqualsHashCodeBench.benchHashCode   avgt    5  ≈ 10⁻³           ns/op
EqualsHashCodeBench.benchNotEquals  avgt    5   0.001 ±  0.001  ns/op

LLITCHEV:
@OlegMazurov per talking to @jasperpotts , could you please take a look. We just want to make sure there are no obvious issues with scalability here. Thanks!!

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -1,203 +1,688 @@
...
+    /**
+     * Make sure first character of a string is upper case
+     *
+     * @param name string input who's first character can be upper or lower case
+     * @return name with first character converted to upper case
+     */
+    public static String capitalizeFirstLetter(String name) {
+        if (name.length() > 0) {
+            if (name.chars().allMatch(Character::isUpperCase)) {
+                return Character.toUpperCase(name.charAt(0)) + name.substring(1).toLowerCase();
```
artemananiev, line 63: The first char is already capitalized, no need to call toUpperCase() for it again

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -1,203 +1,688 @@
...
+     * @param name the input name in camel case
+     * @return output name in upper snake case
+     */
+    public static String camelToUpperSnake(String name) {
+        // check if already camel upper
+        if (name.chars()
+                .allMatch(c -> Character.isUpperCase(c) || Character.isDigit(c) || c == '_'))
+            return name;
+        // check if already has underscores, then just capitalize
+        if (name.chars().anyMatch(c -> c == '_')) return name.toUpperCase();
```
artemananiev, line 103: `name.contains()` ?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -1,203 +1,688 @@
...
+    public static String getFieldsHashCode(final List<Field> fields, String generatedCodeSoFar)
+            throws RuntimeException {
+        for (Field f : fields) {
+            if (f.parent() != null) {
+                final OneOfField oneOfField = f.parent();
+                generatedCodeSoFar += getFieldsHashCode(oneOfField.fields(), generatedCodeSoFar);
+            }
+
+            if (f.optionalValueType()) {
+                generatedCodeSoFar = getOptionalHashCodeGeneration(generatedCodeSoFar, f);
```
artemananiev, line 222: This is questionable, to say the least. Key requirement to `hashCode()` implementation is forward compatibility. Whenever a new field is added to an entity class, it must not change its hash code. New fields are always optional, if they are included here, they would contribute to hash code value.

With that in mind, I can't say if all HS types we use today have some required fields (that can be used to calculate hash codes). If so, should all of them be accounted in `hashCode()`? If the answer to either of these questions is "no", there must be a different mechanism to specify the set of fields to be used for hash codes (annotations?)

LLITCHEV: After talking to Jasper before, adding a new field actually should change the hashcode, if the value of an optional field is not null. In general, field values of null (whether optional or not) do not alter the hash code - default values (they are not considered in the calculation, except null in a list - a list with value noll at a certain position should change the dashcode, I think). Again, fields with def values for the corresponding type are not included in the dashcode calculation.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
LLITCHEV: This is the existing code. It got modified by the spotless formatter, but the code is the same as it was before this PR.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
LLITCHEV: Again... Just spotless formatting modification. But yes, I think contains would be faster probably, instead of having a routine we call for each character.

PR #109, opened 2023-11-07
95 Add stable equals/hashCode for PBJ objects

### LINKED ISSUE #95: Generated PBJ objects should have proper hashCode() implementation

imalygin:
This pull request is a cleaner version of https://github.com/hashgraph/pbj/pull/102

I excluded formatting and fixed some minor warnings. 

**Related issue(s)**:

Fixes #95 

**Notes for reviewer**:

This implementation does not include optional fields with default values into `hashCode` / `equals` calculations, so changing the object definition is not going to affect how existing objects behave. 

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
@@ -447,14 +447,10 @@ public final class Common {
 
 				generatedCodeSoFar += (
                 """
-                if ($fieldName instanceof Object) {
-				    if (this.$fieldName == null && thatObj.$fieldName != null) {
-				        return false;
-				    }
-				    if (!$fieldName.equals(thatObj.$fieldName)) {
-				        return false;
-				    }
-				} else if ($fieldName != thatObj.$fieldName) {
+				if (this.$fieldName == null && thatObj.$fieldName != null) {
+				    return false;
+				}
+				if (this.$fieldName != null && !$fieldName.equals(thatObj.$fieldName)) {
 				    return false;
 				}
 				""").replace("$fieldName", f.nameCamelFirstLower());
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
@@ -67,7 +67,8 @@ public final class Common {
 	 * @param firstUpper if true then first char is upper case otherwise it is lower
 	 * @return out name in camel case
 	 */
-	public static String snakeToCamel(String name, boolean firstUpper) {
+	@NonNull
+	public static String snakeToCamel(@NonNull String name, boolean firstUpper) {
 		final String out =  Arrays.stream(name.split("_")).map(Common::capitalizeFirstLetter).collect(
 				Collectors.joining(""));
 		return (firstUpper ? Character.toUpperCase(out.charAt(0)) : Character.toLowerCase(out.charAt(0)) )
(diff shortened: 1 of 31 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/BufferedData.java
```diff
@@ -372,11 +372,13 @@ public long getBytes(final long offset, @NonNull final BufferedData dst) {
...
-        if(len < 0) throw new IllegalArgumentException("Length cannot be negative");
-        // It is vital that we always copy here, we can never assume ownership of the underlying buffer
-        final var copy = new byte[len];
-        buffer.get(Math.toIntExact(offset), copy, 0, len);
-        return Bytes.wrap(copy);
+        if (direct) {
+            final var copy = new byte[len];
+            buffer.get(Math.toIntExact(offset), copy, 0, len);
+            return Bytes.wrap(copy);
+        } else {
```
artemananiev, line 379: Why are heap byte buffers handled differently? According to the (removed) comment above, the returned Bytes object should always be a copy rather than a wrapper over the existing buffer.

imalygin: I don't have a good explanation, so I rolled back the change :)

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -172,14 +174,367 @@ public static String javaPrimitiveToObjectType(String primitiveFieldType) {
...
+					""").replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.FIXED64 ||
+						f.type() == Field.FieldType.INT64 ||
+						f.type() == Field.FieldType.SFIXED64 ||
+						f.type() == Field.FieldType.SINT64 ||
+						f.type() == Field.FieldType.UINT64) {
+					generatedCodeSoFar += (
+               		"""
+					if ($fieldName != DEFAULT.$fieldName) {
+						result = 31 * result + Long.hashCode($fieldName);
```
artemananiev, line 217: Should the multiplier be 31 for all field types, or something different?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -172,14 +174,367 @@ public static String javaPrimitiveToObjectType(String primitiveFieldType) {
...
+	 * @param generatedCodeSoFar The string that the codegen is generated into.
+	 * @param f The field for which to generate the hash code.
+	 * @return Updated codegen string.
+	 */
+	@NonNull
+	private static String getOptionalHashCodeGeneration(String generatedCodeSoFar, Field f) {
+		switch (f.messageType()) {
+			case "StringValue" -> generatedCodeSoFar += (
+     			"""
+				if ($fieldName != DEFAULT.$fieldName) {
```
artemananiev, line 290: Should it be equals() instead?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -172,14 +174,367 @@ public static String javaPrimitiveToObjectType(String primitiveFieldType) {
...
+					result = 31 * result + Integer.hashCode($fieldName);
+				}
+				""").replace("$fieldName", f.nameCamelFirstLower());
+			case "Int64Value", "UInt64Value" -> generatedCodeSoFar += (
+     			"""
+				if ($fieldName != DEFAULT.$fieldName) {
+					result = 31 * result + Long.hashCode($fieldName);
+				}
+				""").replace("$fieldName", f.nameCamelFirstLower());
+			case "FloatValue" -> generatedCodeSoFar += (
```
artemananiev, line 312: Floats and doubles are risky to compare using == or !=, it should probably be something like `Math.abs($fieldName - DEFAULT.$fieldName) < 1e-6` - the exact epsilon value is to be checked with Jasper / Leemon

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -172,14 +174,367 @@ public static String javaPrimitiveToObjectType(String primitiveFieldType) {
...
+               		"""
+					if ($fieldName != DEFAULT.$fieldName) {
+						result = 31 * result + $fieldName.hashCode();
+					}
+					""").replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.BYTES) {
+					generatedCodeSoFar += (
+               		"""
+					if ($fieldName != DEFAULT.$fieldName) {
+						result = 31 * result + ($fieldName == null ? 0 : $fieldName.hashCode());
```
artemananiev, line 259: Are BYTES fields of type Bytes or byte[]? I assume the former, but if for some reason it is the latter, then Arrays.hashCode() should be used instead.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -172,14 +174,367 @@ public static String javaPrimitiveToObjectType(String primitiveFieldType) {
...
+	 * @param f The field for which to generate the hash code.
+	 * @return Updated codegen string.
+	 */
+	@NonNull
+	private static String getRepeatedHashCodeGeneration(String generatedCodeSoFar, Field f) {
+		generatedCodeSoFar += (
+           """
+			java.util.List list$$fieldName = $fieldName;
+			for (Object o : list$$fieldName) {
+				if ($fieldName != DEFAULT.$fieldName) {
```
artemananiev, line 347: This one is questionable. I would assume that for repeated fields, it makes sense to take all values into consideration, even default values

artemananiev: Another thought... If a repeated field is empty, then it should probably be ignored

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -245,121 +279,206 @@ public void generate(final Protobuf3Parser.MessageDefContext msgDef,
...
+						$hashCodeManipulation
+							return (int)hashCode;
+						}
+						"""
+						.replace("$fieldName", fields.get(0).name())
+						.replace("$hashCodeManipulation", HASH_CODE_MANIPULATION)
+						.indent(DEFAULT_INDENT)
+					;
 				}
 				default -> {
```
artemananiev, line 355: What about boolean fields?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -332,7 +332,7 @@ public int read() throws IOException {
     @Override
     @NonNull
     public String toString() {
-        return HexFormat.of().formatHex(buffer, start, start + length);
+        return HexFormat.of().formatHex(buffer, start, length);
```
artemananiev, line 335: The last param is `toIndex`, so `start + length` seems to be the right value.

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/HashEqualsTest.java
```diff
@@ -0,0 +1,92 @@
...
+import com.hedera.pbj.test.proto.pbj.TimestampTest;
+import com.hedera.pbj.test.proto.pbj.TimestampTest2;
+
+class HashEqualsTest {
+    @Test
+    void differentObjectsWithDefaulEquals() {
+        TimestampTest tst = new TimestampTest(1, 2);
+        TimestampTest2 tst2 = new TimestampTest2(1, 2, 0);
+
+        assertFalse(tst.equals(tst2));
```
artemananiev, line 18: 0 is the default value for int32, why are these two objects considered not equal?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: As far as I can tell, the usage of 31 is universal in this particular context. As it turned out there is a good explanation for that:

Computational Efficiency: The choice of 31 can be attributed to the fact that it allows for an efficient implementation of multiplication. In binary form, 31 is 11111, and multiplying by 31 can be efficiently implemented as a shift and a subtraction: (x << 5) - x. This is much faster than a general multiplication operation.

Balancing Speed and Distribution: While there are other primes, 31 offers a good balance between computational efficiency (due to its binary representation) and the effectiveness in producing distinct hash codes for different inputs.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: In this particular context `equals` and reference equality should work exactly the same because the default value is singleton. However, for the sake of clean code I'll replace it with `equals`

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: but that is not comparison to a real value, likely it's comparison to zero and `equals` should be sufficiently robust for that

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: I checked - it's `com.hedera.pbj.runtime.io.buffer.Bytes`

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/HashEqualsTest.java
imalygin: It's a repeated field. If the number of entries is different, the object is not equal.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: I addressed both comments.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
imalygin: Fixed

artemananiev (dismissed):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -172,14 +174,365 @@ public static String javaPrimitiveToObjectType(String primitiveFieldType) {
...
+	 * @param fields The fields of this object.
+	 * @param generatedCodeSoFar The accumulated hash code so far.
+	 * @return The generated code for getting the hashCode value.
+	 */
+	public static String getFieldsHashCode(final List<Field> fields, String generatedCodeSoFar) {
+		for(Field f : fields) {
+			if (f.parent() != null) {
+				final OneOfField oneOfField = f.parent();
+				generatedCodeSoFar += getFieldsHashCode(oneOfField.fields(), generatedCodeSoFar);
+			}
```
artemananiev, line 189: I don't know much about OneOf fields, so the question is: can a OneOf field be optional/repeated? I mean, should line 191 below be just "if" or "else if"?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -172,14 +174,365 @@ public static String javaPrimitiveToObjectType(String primitiveFieldType) {
...
+				    result = 31 * result + Double.hashCode($fieldName);
+				}
+				""").replace("$fieldName", f.nameCamelFirstLower());
+			case "BytesValue" -> generatedCodeSoFar += (
+     			"""
+				if ($fieldName != null && !$fieldName.equals(DEFAULT.$fieldName)) {
+				    result = 31 * result + ($fieldName == null ? 0 : $fieldName.hashCode());
+				}
+				""").replace("$fieldName", f.nameCamelFirstLower());
+			default -> throw new UnsupportedOperationException("Unhandled optional message type:" + f.messageType());
```
artemananiev, line 330: What about other messages? I mean, if field type is another PBJ message class. An example is accountId field (of type AccountID) in the Account class.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -172,14 +174,365 @@ public static String javaPrimitiveToObjectType(String primitiveFieldType) {
...
+                    return false;
+                }
+                if (this.$fieldName != null && !$fieldName.equals(thatObj.$fieldName)) {
+				    return false;
+                }
+				""").replace("$fieldName", f.nameCamelFirstLower());
+			case "BoolValue" ->
+				generatedCodeSoFar += (
+                """
+                if ($fieldName instanceof Object) {
```
artemananiev, line 466: When can a field not be an instance of Object?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: Yes, you're right. It has to be mutually exclusive.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: This particular method is designed for handling specific set of classes. Take a look at `SingleField#optionalValueType` definition:
```
	public boolean optionalValueType() { // Move logic for checking built in types to common
		return type == SingleField.FieldType.MESSAGE && (
				messageType.equals("StringValue") ||
				messageType.equals("Int32Value") ||
				messageType.equals("UInt32Value") ||
				messageType.equals("Int64Value") ||
				messageType.equals("UInt64Value") ||
				messageType.equals("FloatValue") ||
				messageType.equals("DoubleValue") ||
				messageType.equals("BoolValue") ||
				messageType.equals("BytesValue")
		);
	}
```
As you can see, it includes only primitive value wrappers and doesn't include any value class in general. Therefore, this condition 
```
} else if (f.optionalValueType()) {
    generatedCodeSoFar = getOptionalHashCodeGeneration(generatedCodeSoFar, f);
} 
```
is not going to be true.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -172,14 +174,365 @@ public static String javaPrimitiveToObjectType(String primitiveFieldType) {
...
+				generatedCodeSoFar += getFieldsHashCode(oneOfField.fields(), generatedCodeSoFar);
+			}
+
+			if (f.optionalValueType()) {
+				generatedCodeSoFar = getOptionalHashCodeGeneration(generatedCodeSoFar, f);
+			}
+			else if (f.repeated()) {
+				generatedCodeSoFar = getRepeatedHashCodeGeneration(generatedCodeSoFar, f);
+			}
+			else if (f.nameCamelFirstLower() != null) {
```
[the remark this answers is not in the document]
imalygin: Yes, agreed.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: If it's a primitive.

artemananiev (approved):

PR #111, opened 2023-11-11
59: Improve Bytes Performance

artemananiev:
Fix summary:
* UnsafeUtils class is added to provide a set of classes on top of sun.misc.Unsafe
* Bytes.getInt() and getLong() are implemented using UnsafeUtils
* A few other Bytes methods are overridden to provide more efficient implementation than what's in RandomAccessData: contains(), asUtf8String()

Testing:
* Unit tests for UnsafeUtils.getInt() and getLong() are added
* Unit tests for overridden Bytes methods are added
* A JMH benchmark for the old and the new getInt() and getLong() implementations is provided

Fixes: https://github.com/hashgraph/pbj/issues/59

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/UnsafeUtils.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/UnsafeUtils.java
@@ -1,0 +1,69 @@
+package com.hedera.pbj.runtime.io;
+
+import java.lang.reflect.Field;
+import java.nio.Buffer;
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import sun.misc.Unsafe;
+
+/**
+ * A set of utility methods on top of sun.misc.Unsafe
+ */
+public class UnsafeUtils {
+
+    private static final Unsafe UNSAFE;
+
+    private static final boolean NEED_CHANGE_BYTE_ORDER;
+
+    private static final int BYTE_ARRAY_BASE_OFFSET;
+
+    static {
+        try {
+            Field f = Unsafe.class.getDeclaredField("theUnsafe");
+            f.setAccessible(true);
+            UNSAFE = (Unsafe) f.get(null);
+            NEED_CHANGE_BYTE_ORDER = ByteOrder.nativeOrder() == ByteOrder.LITTLE_ENDIAN;
+            BYTE_ARRAY_BASE_OFFSET = UNSAFE.arrayBaseOffset(byte[].class);
+        } catch (NoSuchFieldException | SecurityException | IllegalArgumentException | IllegalAccessException e) {
+            throw new InternalError(e);
+        }
+    }
+
+    private UnsafeUtils() {
+    }
+
+    /**
+     * Reads an integer from the given array starting at the given offset. Array bytes are
+     * interpreted in BIG_ENDIAN order.
+     *
+     * @param arr The byte array
+     * @param offset The offset to read an integer at
+     * @return The integer number
+     * @throws java.nio.BufferOverflowException If array length is less than offset + integer bytes
+     */
+    public static int getInt(final byte[] arr, final int offset) {
+        if (arr.length < offset + Integer.BYTES) {
+            throw new BufferUnderflowException();
+        }
+        final int value = UNSAFE.getInt(arr, BYTE_ARRAY_BASE_OFFSET + offset);
+        return NEED_CHANGE_BYTE_ORDER ? Integer.reverseBytes(value) : value;
+    }
+
+    /**
+     * Reads a long from the given array starting at the given offset. Array bytes are
+     * interpreted in BIG_ENDIAN order.
+     *
+     * @param arr The byte array
+     * @param offset The offset to read a long at
+     * @return The long number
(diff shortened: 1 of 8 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
```diff
@@ -455,15 +455,10 @@ default String asUtf8String() {
...
-     * @throws BufferUnderflowException if {@code len} is greater than {@link #length()} - {@code offset}.
      * @throws IndexOutOfBoundsException If the given {@code offset} is negative or not less than {@link #length()}
      */
     @NonNull
     default String asUtf8String(final long offset, final long len) {
-        if (len > length() - offset) {
-            throw new BufferUnderflowException();
-        }
-
         if (offset < 0 || offset + len > length()) {
```
rbair23, line 462: Why is it safe to remove this check? Can it never happen?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/UnsafeUtils.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private static final Unsafe UNSAFE;
+
+    private static final boolean NEED_CHANGE_BYTE_ORDER;
+
+    private static final int BYTE_ARRAY_BASE_OFFSET;
+
+    static {
+        try {
+            Field f = Unsafe.class.getDeclaredField("theUnsafe");
```
rbair23, line 21:
```suggestion
            final Field f = Unsafe.class.getDeclaredField("theUnsafe");
```

rbair23:
Do you have any data from before and after this change with JMH? I'm wondering what the raw improvement numbers are.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
artemananiev: `len > length() - offset` is the same as `offset + len > length()`, which is checked next.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/UnsafeUtils.java
artemananiev: OK, will make it final

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
rbair23: Oh, I see, it was moved to the subclass implementation in Bytes. But other subclasses of RandomAccessData may still want this check?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -490,11 +513,34 @@ public Bytes getBytes(long offset, long length) {
         return new Bytes(buffer, Math.toIntExact(start + offset), Math.toIntExact(length));
     }
 
+    /** {@inheritDoc} */
+    @NonNull
+    @Override
+    public String asUtf8String(long offset, long len) {
```
rbair23, line 519:
```suggestion
    public String asUtf8String(final long offset, final long len) {
```

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -490,11 +513,34 @@ public Bytes getBytes(long offset, long length) {
...
+        }
+        if (len == 0) {
+            return "";
+        }
+        return new String(buffer, Math.toIntExact(offset), length, StandardCharsets.UTF_8);
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public boolean contains(long offset, @NonNull byte[] bytes) {
```
rbair23, line 531:
```suggestion
    public boolean contains(final long offset, @NonNull final byte[] bytes) {
```

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
rbair23: Haha, indeed :-). Should we keep the first check and throw BufferUnderflowException in that case, and in the secon check just do offset < 0 for IndexOutOfBounds?

I don't have a real opinion here, but I think we should take a holistic view of which exceptions are thrown where to make sure they are consistent throughout all the API. Maybe that change should be done as a separate PR.

artemananiev: The reason I got rid of BufferUnderflowException here is this exception is specific to buffers, when they are used to read data sequentially, but there is no data to read. This interface is `RandomAccessData`, which is more like an array than like a buffer, so IndexOutOfBoundsException makes much more sense to me

rbair23 (approved):

Issue #27, opened 2023-03-31
ReadableStreamingData should implement bulk read methods


jasperpotts:
### Problem

ReadableStreamingData should implement efficient versions of the InputStream bulk read methods.
```
    public long readBytes(@NonNull byte[] dst) {
    public long readBytes(@NonNull final byte[] dst, final int offset, final int maxLength) {
```
As the default just call read() in a loop which is not idea. Most InputStream sub classes like FileInputStream have efficient implementations

### Solution

Something like this maybe
```
    /** {@inheritDoc} */
    @Override
    public long readBytes(@NonNull byte[] dst) {
        return readBytes(dst, 0, dst.length);
    }

    /** {@inheritDoc} */
    @SuppressWarnings("DuplicatedCode")
    public long readBytes(@NonNull final byte[] dst, final int offset, final int maxLength) {
        if (maxLength < 0) {
            throw new IllegalArgumentException("Negative maxLength not allowed");
        }
        // return 0 if there is no data remaining to be read
        if (!hasRemaining()) return 0;

        int maxReadable = Math.min((int)Math.min(Integer.MAX_VALUE,remaining()), maxLength);
        try {
            int readBytes =  in.read(dst,offset,maxReadable);
            // our spec differs from InputStream that we return 0 when no bytes available to read because we reached
            // end of the file.
            if (readBytes == -1) {
                eof == true;
                return 0;
            }
            return readBytes;
        } catch (IOException e) {
            throw new DataAccessException(e);
        }
    }
```

artemananiev:
It appeared to be slightly more complicated than the proposed code snippet above. The issue is with `eof`. The streaming object is expected to reach EOF even if it was able to read some bytes. That it, checking the result of `in.read()` against -1 is not enough. There is test for that: it creates a byte array input stream over an array of 10 elements, then tries to read 12 bytes from it. 10 bytes are expected to be read (this is fine), but in addition to that, hasRemaining() is expected to be false. With the proposed implementation above it will be true: `eof` is still false, and position (10) is less than limit (which is usually Long.MAX_VALUE for streaming data, but in this particular test is set to 12).

PR #129, opened 2023-12-01
fix: 27: ReadableStreamingData should implement bulk read methods

artemananiev:
Fix summary:

* `readBytes()` methods are implemented in `ReadableStreamingData` in a more efficient way than inherited from `ReadableSequentialData`
* Unit tests are refactored, so some tests previously run for `ReadableStreamingData` only now are also used for other implementations of `ReadableSequentialData` (e.g. `BufferedData`)

Fixes: https://github.com/hashgraph/pbj/issues/27

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
@@ -2,10 +2,15 @@ package com.hedera.pbj.runtime.io.stream;
 
 import com.hedera.pbj.runtime.io.DataAccessException;
 import com.hedera.pbj.runtime.io.ReadableSequentialData;
+import com.hedera.pbj.runtime.io.buffer.BufferedData;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -96,12 +101,9 @@ public class ReadableStreamingData implements ReadableSequentialData, AutoClosea
     /** {@inheritDoc} */
     @Override
     public byte readByte() {
-        // It should NEVER be possible for position to exceed limit, but being extra safe here
-        // using >= instead of just ==
-        if (eof || position >= limit) {
+        if (!hasRemaining()) {
             throw new BufferUnderflowException();
         }
-
         // We know result is a byte, because we've already checked for EOF, and we know that
         // it will only ever be a byte, unless it is EOF, in which case it is a -1 int.
         try {
@@ -112,7 +114,7 @@ public class ReadableStreamingData implements ReadableSequentialData, AutoClosea
             }
             position++;
             return (byte) result;
-        } catch (IOException e) {
+        } catch (final IOException e) {
             throw new DataAccessException(e);
         }
     }
@@ -130,8 +132,97 @@ public class ReadableStreamingData implements ReadableSequentialData, AutoClosea
             long numSkipped = in.skip(clamped);
             position += numSkipped;
             return numSkipped;
-        } catch (IOException e) {
+        } catch (final IOException e) {
             throw new DataAccessException(e);
+        }
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public long readBytes(@NonNull final byte[] dst, final int offset, final int maxLength) {
+        if (maxLength < 0) {
+            throw new IllegalArgumentException("Negative maxLength not allowed");
+        }
+        if (!hasRemaining()) {
+            return 0;
+        }
+        try {
+            int bytesRead = in.read(dst, offset, maxLength);
+            position += bytesRead;
+            if (bytesRead < maxLength) {
+                final int nextByte = tryReadNextByte();
+                if (nextByte != -1) {
+                    dst[offset + bytesRead] = (byte) nextByte;
+                    bytesRead++;
+                }
+            }
+            return bytesRead;
+        } catch (final IOException e) {
+            throw new DataAccessException(e);
+        }
+    }
+
+    @Override
+    public long readBytes(@NonNull final ByteBuffer dst) {
+        if (!dst.hasArray()) {
+            return ReadableSequentialData.super.readBytes(dst);
+        }
+        if (!hasRemaining()) {
+            return 0;
+        }
+        final byte[] dstArr = dst.array();
+        final int dstArrOffset = dst.arrayOffset();
+        final int dstPos = dst.position();
+        final long len = Math.min(remaining(), dst.remaining());
+        try {
+            int bytesRead = in.read(dstArr, dstPos + dstArrOffset, Math.toIntExact(len));
+            position += bytesRead;
+            if (bytesRead < len) {
+                final int nextByte = tryReadNextByte();
+                if (nextByte != -1) {
+                    dst.put((byte) nextByte);
+                    bytesRead++;
+                }
+            }
+            return bytesRead;
+        } catch (final IOException e) {
+            throw new DataAccessException(e);
+        }
+    }
+
+    @Override
+    public long readBytes(@NonNull final BufferedData dst) {
+        final long len = Math.min(remaining(), dst.remaining());
+        int bytesRead = dst.writeBytes(in, Math.toIntExact(len));
+        position += bytesRead;
+        if (position < len) {
+            final int nextByte = tryReadNextByte();
+            if (nextByte != -1) {
+                dst.writeByte((byte) nextByte);
+                bytesRead++;
+            }
+        }
+        return bytesRead;
+    }
+
+    // This method is used when less than requested bytes have been read from the input stream.
+    // It could happen for two reasons: end of stream or because no more bytes are immediately
+    // available. A simple check for in.available() can't be used to detect end of stream. The
+    // workaround is to try to read the next byte and check the result. If it is -1, then the
+    // stream has really reached the end, otherwise this method returns the read byte
+    private byte tryReadNextByte() {
+        try {
+            final int nextByte = in.read();
+            if (nextByte == -1) {
+                eof = true;
+                return -1;
+            } else {
+                position++;
+                return (byte) nextByte;
+            }
+        } catch (final IOException e) {
+            eof = true;
+            return -1;
         }
     }
 }
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
@@ -3,9 +3,7 @@ package com.hedera.pbj.runtime.io.stream;
 import com.hedera.pbj.runtime.io.DataAccessException;
 import com.hedera.pbj.runtime.io.ReadableSequentialData;
 import com.hedera.pbj.runtime.io.buffer.BufferedData;
-import com.hedera.pbj.runtime.io.buffer.Bytes;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.BufferUnderflowException;
@@ -143,18 +141,19 @@ public class ReadableStreamingData implements ReadableSequentialData, AutoClosea
         if (maxLength < 0) {
             throw new IllegalArgumentException("Negative maxLength not allowed");
         }
-        if (!hasRemaining()) {
+        if ((offset < 0) || (maxLength > dst.length - offset)) {
+            throw new IndexOutOfBoundsException("Illegal read offset / maxLength");
+        }
+        final int len = Math.min(dst.length - offset, maxLength);
+        if ((len == 0) || !hasRemaining()) {
+            // Nothing to do
             return 0;
         }
         try {
-            int bytesRead = in.read(dst, offset, maxLength);
+            int bytesRead = in.readNBytes(dst, offset, len);
             position += bytesRead;
-            if (bytesRead < maxLength) {
-                final int nextByte = tryReadNextByte();
-                if (nextByte != -1) {
-                    dst[offset + bytesRead] = (byte) nextByte;
-                    bytesRead++;
-                }
+            if (bytesRead < len) {
+                eof = true;
             }
             return bytesRead;
         } catch (final IOException e) {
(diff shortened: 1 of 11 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
```diff
@@ -130,8 +132,97 @@ public long skip(final long n) {
...
+        if (maxLength < 0) {
+            throw new IllegalArgumentException("Negative maxLength not allowed");
+        }
+        if (!hasRemaining()) {
+            return 0;
+        }
+        try {
+            int bytesRead = in.read(dst, offset, maxLength);
+            position += bytesRead;
+            if (bytesRead < maxLength) {
```
jsync-swirlds, line 152: This should also check that bytesRead is less than `dst.length` so we do not try to write past the end of the array.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
```diff
@@ -130,8 +132,97 @@ public long skip(final long n) {
...
+    @Override
+    public long readBytes(@NonNull final byte[] dst, final int offset, final int maxLength) {
+        if (maxLength < 0) {
+            throw new IllegalArgumentException("Negative maxLength not allowed");
+        }
+        if (!hasRemaining()) {
+            return 0;
+        }
+        try {
+            int bytesRead = in.read(dst, offset, maxLength);
```
jsync-swirlds, line 150: If we use `in.readNBytes` here (same parameters) then we don't need the if loop below as the read will block until the bytes are read or EOS is reached. [Docs here.](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/InputStream.html#readNBytes(byte%5B%5D,int,int))
>This method blocks until the requested number of bytes has been read, end of stream is detected, or an exception is thrown.

It is worth noting that the default implementation of `read` actually behaves the same, but `readNBytes` is _specified_ to behave in this manner.
>The default implementation of this method blocks until the requested amount of input data len has been read, end of file is detected, or an exception is thrown.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
```diff
@@ -130,8 +132,97 @@ public long skip(final long n) {
...
+        }
+        return bytesRead;
+    }
+
+    // This method is used when less than requested bytes have been read from the input stream.
+    // It could happen for two reasons: end of stream or because no more bytes are immediately
+    // available. A simple check for in.available() can't be used to detect end of stream. The
+    // workaround is to try to read the next byte and check the result. If it is -1, then the
+    // stream has really reached the end, otherwise this method returns the read byte
+    private byte tryReadNextByte() {
```
jsync-swirlds, line 213: Rather than this workaround, we should use `readNBytes` if the intent is to block until all requested data is read or we reach end of stream.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
```diff
@@ -130,8 +132,97 @@ public long skip(final long n) {
...
+            return bytesRead;
+        } catch (final IOException e) {
+            throw new DataAccessException(e);
+        }
+    }
+
+    @Override
+    public long readBytes(@NonNull final BufferedData dst) {
+        final long len = Math.min(remaining(), dst.remaining());
+        int bytesRead = dst.writeBytes(in, Math.toIntExact(len));
```
jsync-swirlds, line 196: We should have an alternative `writeBytes` (it is very odd that we use `writeBytes` to _read_ data), or better a `readNBytes`, method on BufferedData that uses `readNBytes` on the input stream so that we do not need these single-byte-reading workarounds for when we need to block efficiently until the full requested length is read (or EOS is reached).

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
```diff
@@ -130,8 +132,97 @@ public long skip(final long n) {
...
+        }
+        if (!hasRemaining()) {
+            return 0;
+        }
+        final byte[] dstArr = dst.array();
+        final int dstArrOffset = dst.arrayOffset();
+        final int dstPos = dst.position();
+        final long len = Math.min(remaining(), dst.remaining());
+        try {
+            int bytesRead = in.read(dstArr, dstPos + dstArrOffset, Math.toIntExact(len));
```
jsync-swirlds, line 178: If the intent is to read all the requested data if possible, and block when needed, we should use `readNBytes` instead of read.  The "read a single byte" workaround isn't needed if we do that.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialTestBase.java
```diff
@@ -0,0 +1,207 @@
...
+    @Override
+    protected abstract ReadableSequentialData fullyUsedSequence();
+
+    @Override
+    @NonNull
+    protected abstract ReadableSequentialData sequence(@NonNull byte [] arr);
+
+    @Test
+    @DisplayName("Stream with no data")
+    void noDataTest() throws Exception {
```
jsync-swirlds, line 27: Even in test methods, declaring `throws Exception` is not great, as it can hide unexpected changes in API.  Generally better to declare `throws` only as necessary.

jsync-swirlds (commented):
I believe we could simplify much of this with a slightly different approach.
Interested in your take on this.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
artemananiev: I was thinking about `readNBytes()`, too, but decided not to proceed with it. Here is a scenario: a request to read N bytes from an input stream, which is known to be much larger than N, but less than N bytes are immediately available. According to `InputStream.read()` spec, it may not read all requested N bytes, but only available bytes. Subsequent reads may read more bytes, as they become available.

PBJ interfaces like `ReadableSequentialData` try to be at the same level as `InputStream` and `ByteBuffer`. This is why semantics of `ReadableSequentialData.readBytes()` is "read as much as available and return the number of bytes read", but not "read exactly the specified number of bytes, or throw an exception" (similar to `readNBytes()`)

In MerkleDb code, we have utility methods `MerkleDbFileUtils.completelyRead()/completelyWrite()`, which are to help with such incomplete reads/writes. Similar methods will be provided at PBJ protobuf level (e.g. in `ProtoParserTools` and `ProtoWriterTools`). IO interfaces like `ReadableSequentialData` are too low-level for that, in my opinion.

@jasperpotts What was the original idea behind PBJ data I/O library APIs? Should they be as low-level as `InputStream` and `ByteBuffer`, or should work as higher level utility methods like `MerkleDbFileUtils.completelyRead()`?

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialTestBase.java
artemananiev: This "throws Exception" clause comes from `AutoCloseable` interface. Originally these tests were enabled only for `ReadableStreamingData`, which is known to implement `AutoCloseable`, so the tests just used try-with-resources. Now I'm trying to make the tests more generic, so they can be used for any `ReadableSequentialData`, e.g. `BufferedData`, but not all `ReadableSequentialData` are `AutoCloseable`.

The only (slightly ugly) option I see is to drop "throws Exception" from all test methods, but instead explicitly catch exceptions from `AutoCloseable.close()` and re-throw them as runtime exceptions. What do you think about it?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
artemananiev: See my comment above on `readNBytes()`. It's a design question, what behavior we'd expect from these `readBytes()` and similar methods.

Regarding reads vs writes, I agree it may look odd at a glance. After a while, it starts feeling quite natural, though. Reading from this to dst is the same as writing this to dst. Similar, writing src to this is the same as reading from src to this.

Now let's take a look at `BufferedData.writeBytes(InputStream)`. It's still single-byte writing code, unless the buffered data object is backed by a byte array. Instead of using `dst.writeBytes()` here, I could check if dst has a byte array and use `in.read(array)` as a shortcut. Unfortunately (or fortunately, actually), `BufferedData` provides no API to check if it's direct of byte array based, and no API to get access to the underlying byte array. This is why `dst.writeBytes()` is used instead, and let `BufferedData` subclasses handle it internally.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
jsync-swirlds: readNBytes doesn't throw an exception if it reads too few bytes, it just keeps calling for bytes until it gets the requested number or reaches end of stream.  It's doing _exactly_ what you're doing here, just built into InputStream.

Using readNBytes doesn't change the semantics of your PR, it just reduces the code we have to implement and maintain.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialTestBase.java
jsync-swirlds: I would actually recommend using the `Closeable` interface in the test here (which anything dealing with IO, such as PBJ code, should implement, rather than the parent interface AutoCloseable). That only throws IOException, which is pretty close to what we actually might throw from the code.

If the PBJ classes don't implement `Closeable`, we should change them to do so (`Closeable` extends `AutoCloseable`).

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
anthony-swirldslabs: @artemananiev , while the PBJ interfaces may try to be at the same level as `InputStream` and `ByteBuffer`, they also have their own specification defined in their JavaDocs. And if I read the JavaDoc for `long ReadableSequentialData.readBytes(@NonNull final byte[] dst, final int offset, final int maxLength)` correctly, then it does indeed align with that of the `InputStream.readNBytes`. So if we can take advantage of this method directly and also simplify our code here, then I agree with @jsync-swirlds . Please let me know if I'm missing any differences in the respective specifications that would invalidate the point.

artemananiev: OK, I think I know where most confusion comes from. In JDK terms, "remaining" means the bytes that can be are immediately available, they can be read without blocking. In PBJ, "remaining" semantics is slightly different, it means everything up to the limit.

With that in mind, if `readBytes()` is guaranteed to read `Math.min(remaining, maxLength)` bytes, regardless of how many bytes are immediately available in the stream, then I agree using `InputStream.readNBytes()` is the way to go. Let me make this change. It would simplify the code quite a bit and would eliminate need in this tricky "read one next byte" workaround.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialTestBase.java
artemananiev: Do I read it correctly, is your proposal to not close anything in the tests (aside from Closeable vs AutoCloseable, which should be taken care of in a separate fix)?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
artemananiev: Changed to use `readNBytes()`

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
artemananiev: Switched to `readNBytes()` here

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
artemananiev: This odd method is dropped in favor of using `readNBytes()`

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
artemananiev: Switched to use `readNBytes()` here

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialTestBase.java
jsync-swirlds: I propose to _add an issue_ to replace `AutoCloseable` with `Closeable` as the interface implemented in PBJ I/O classes.  As part of that issue we can replace `Exception` with `IOException` on the tests.
We should not change the behavior of the tests.

jsync-swirlds (approved):

anthony-swirldslabs (approved):
Looks good.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialTestBase.java
artemananiev: I filed https://github.com/hashgraph/pbj/issues/134 about it. Note that even after that issue is fixed, we won't be able to use try-with-resources in all tests, only some of them. The problem is `Closeable` (or `AutoCloseable` today) is at the class level, `ReadableStreamingData` and `WriteableStreamingData`, while tests work with interfaces `ReadableSequentialData` and `WriteableSequentialData`, which are not closeable. That's why we will still have extra instanceof checks.

Issue #110, opened 2023-11-07
ReadableStreamingData limit is not set if it's wrapped over a file stream


artemananiev:
When a `ReadableStreamingData` object is created on top of a file input stream, its limit is not set. Code snippet:

```java
try (final InputStream fileIn = Files.newInputStream(path, StandardOpenOption.READ)) {
    final ReadableSequentialData in = new ReadableStreamingData(fileIn);
    System.out.println(in.limit());
}
```

To work it around, I have to explicitly set a limit like this:

```java
    in.limit(Files.size(path));
```

It would be great to have it done automatically. For example, provide a new constructor that accepts a file path, or maybe something like that.

PR #136, opened 2023-12-04
fix: 110: ReadableStreamingData limit is not set if it's wrapped over a file stream

artemananiev:
Fix summary:

* A new constructor is added to `ReadableStreamingData` to create a streaming object for a file path. It opens a new file input stream and sets the limit to the size of the file
* A new constructor is added to the same class to create a streaming object from a byte array. It creates a ByteArrayInputStream, the limit is set to byte array length

Testing:

* New unit tests are added to cover both new constructors and check the limits

Fixes: https://github.com/hashgraph/pbj/issues/110

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
@@ -47,6 +47,9 @@ public class ReadableStreamingData implements ReadableSequentialData, AutoClosea
      * @throws IOException if an I/O error occurs
      */
     public ReadableStreamingData(@NonNull final Path file) throws IOException {
+        if (!Files.isRegularFile(file) || !Files.isReadable(file)) {
+            throw new IOException("Cannot read file: " + file);
+        }
         this.in = Files.newInputStream(file, StandardOpenOption.READ);
         this.limit = Files.size(file);
     }
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
@@ -4,10 +4,14 @@ import com.hedera.pbj.runtime.io.DataAccessException;
 import com.hedera.pbj.runtime.io.ReadableSequentialData;
 import com.hedera.pbj.runtime.io.buffer.BufferedData;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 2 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
```diff
@@ -27,14 +31,36 @@ public class ReadableStreamingData implements ReadableSequentialData, AutoClosea
...
+    /**
+     * Opens a new input stream to read a given file and creates a new streaming data object on top
+     * of this stream.
+     *
+     * @param file the file, can not be null
+     * @throws IOException if an I/O error occurs
+     */
+    public ReadableStreamingData(@NonNull final Path file) throws IOException {
+        this.in = Files.newInputStream(file, StandardOpenOption.READ);
+        this.limit = Files.size(file);
```
jsync-swirlds, line 51: Would it be worth checking here if `file` is a "regular" file?
We should never be trying to read from any other type, but given `Files.size` is undefined for any non-regular file (e.g. pipes, sockets, devices, etc...), it might be worth checking.

jsync-swirlds (dismissed):

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingDataTest.java
```diff
@@ -180,4 +183,32 @@ void limitNotChanged() {
...
+    void zeroLimitReadEmptyFile() throws IOException {
+        final Path file = Files.createTempFile(getClass().getSimpleName(), "zeroLimitReadEmptyFile");
+        try (final var stream = new ReadableStreamingData(file)) {
+            assertThat(stream.limit()).isEqualTo(0);
+        }
+    }
+
+    @Test
+    void nonZeroLimitReadFile() throws IOException {
+        final Path file = Files.createTempFile(getClass().getSimpleName(), "nonZeroLimitReadFile");
```
anthony-swirldslabs, line 196: Should the test delete the file? The spec for Files.createTempFile() says that one needs to use `DELETE_ON_CLOSE`, or call `File.deleteOnExit()` for this to actually happen.

anthony-swirldslabs (dismissed):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
artemananiev: OK, I added a check that the provided file is a regular file (sym links followed), and the file is readable.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingDataTest.java
artemananiev: Good catch. Added code to delete the temp files on exit.

anthony-swirldslabs (approved):

jsync-swirlds (approved):

PR #137, opened 2023-12-07
Fix: exposed multiple methods of ProtoParserTools and added unit test…

### LINKED ISSUE #74: Some sizeOf methods aren't exposed publicly from ProtoWriterTools
Several `sizeOfXYZ()` methods are not public in `ProtoWriterTools`. For example:

* sizeOfTag()
* sizeOfVarInt32()
* sizeOfUnsignedVarInt32()
* etc.

I don't know why other methods are public, but these aren't. My guess is these methods are too low level and aren't expected to be used by 3rd-party code. For example, to get size of an integer, they should always use `sizeOfInteger()`, which internally calls these private methods.

However, in MerkleDb PBJ code I do need at least `sizeOfTag()`, since MerkleDb writes and reads field tags manually, not using codecs. Also, `sizeOfInteger()` has valid optimization of skipping default int values (i.e. zero), but in case of MerkleDb it's sometimes helpful to write default values, too, to make file structure a little bit more regular (still protobuf compatible).

imalygin:
Exposed multiple methods of ProtoParserTools and added unit tests for them. Also, added `readNextFieldNumber` method

**Related issue(s)**:

Fixes #74 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
@@ -1,5 +1,10 @@
 package com.hedera.pbj.runtime;
 
+import static com.hedera.pbj.runtime.ProtoConstants.WIRE_TYPE_DELIMITED;
+import static com.hedera.pbj.runtime.ProtoConstants.WIRE_TYPE_FIXED_32_BIT;
+import static com.hedera.pbj.runtime.ProtoConstants.WIRE_TYPE_FIXED_64_BIT;
+import static com.hedera.pbj.runtime.ProtoConstants.WIRE_TYPE_VARINT_OR_ZIGZAG;
+
 import com.hedera.pbj.runtime.io.WritableSequentialData;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import com.hedera.pbj.runtime.io.buffer.RandomAccessData;
@@ -18,7 +23,7 @@ import java.util.function.ToIntFunction;
 public final class ProtoWriterTools {
 
     /** The number of leading bits of the tag that are used to store field type, the rest is field number */
-    private static final int TAG_TYPE_BITS = 3;
+    static final int TAG_TYPE_BITS = 3;
 
     /** Instance should never be created */
     private ProtoWriterTools() {}
@@ -34,14 +39,14 @@ public final class ProtoWriterTools {
      */
     public static ProtoConstants wireType(final FieldDefinition field) {
         return switch (field.type()) {
-            case FLOAT -> ProtoConstants.WIRE_TYPE_FIXED_32_BIT;
-            case DOUBLE -> ProtoConstants.WIRE_TYPE_FIXED_64_BIT;
-            case INT32, INT64, SINT32, SINT64, UINT32, UINT64 -> ProtoConstants.WIRE_TYPE_VARINT_OR_ZIGZAG;
-            case FIXED32, SFIXED32 -> ProtoConstants.WIRE_TYPE_FIXED_32_BIT;
-            case FIXED64, SFIXED64 -> ProtoConstants.WIRE_TYPE_FIXED_64_BIT;
-            case BOOL -> ProtoConstants.WIRE_TYPE_VARINT_OR_ZIGZAG;
-            case BYTES, MESSAGE, STRING -> ProtoConstants.WIRE_TYPE_DELIMITED;
-            case ENUM -> ProtoConstants.WIRE_TYPE_VARINT_OR_ZIGZAG;
+            case FLOAT -> WIRE_TYPE_FIXED_32_BIT;
+            case DOUBLE -> WIRE_TYPE_FIXED_64_BIT;
+            case INT32, INT64, SINT32, SINT64, UINT32, UINT64 -> WIRE_TYPE_VARINT_OR_ZIGZAG;
+            case FIXED32, SFIXED32 -> WIRE_TYPE_FIXED_32_BIT;
+            case FIXED64, SFIXED64 -> WIRE_TYPE_FIXED_64_BIT;
+            case BOOL -> WIRE_TYPE_VARINT_OR_ZIGZAG;
+            case BYTES, MESSAGE, STRING -> WIRE_TYPE_DELIMITED;
+            case ENUM -> WIRE_TYPE_VARINT_OR_ZIGZAG;
         };
     }
 
@@ -94,21 +99,21 @@ public final class ProtoWriterTools {
         }
         switch (field.type()) {
             case INT32 -> {
-                writeTag(out, field, ProtoConstants.WIRE_TYPE_VARINT_OR_ZIGZAG);
+                writeTag(out, field, WIRE_TYPE_VARINT_OR_ZIGZAG);
                 out.writeVarInt(value, false);
             }
             case UINT32 -> {
-                writeTag(out, field, ProtoConstants.WIRE_TYPE_VARINT_OR_ZIGZAG);
+                writeTag(out, field, WIRE_TYPE_VARINT_OR_ZIGZAG);
                 out.writeVarLong(Integer.toUnsignedLong(value), false);
             }
             case SINT32 -> {
-                writeTag(out, field, ProtoConstants.WIRE_TYPE_VARINT_OR_ZIGZAG);
+                writeTag(out, field, WIRE_TYPE_VARINT_OR_ZIGZAG);
                 out.writeVarInt(value, true);
             }
             case SFIXED32, FIXED32 -> {
                 // The bytes in protobuf are in little-endian order -- backwards for Java.
                 // Smallest byte first.
-                writeTag(out, field, ProtoConstants.WIRE_TYPE_FIXED_32_BIT);
+                writeTag(out, field, WIRE_TYPE_FIXED_32_BIT);
                 out.writeInt(value, ByteOrder.LITTLE_ENDIAN);
             }
             default -> throw unsupported();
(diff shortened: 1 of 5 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -805,7 +825,7 @@ private static int sizeOfUnsignedVarInt32(final int value) {
      * @param value The int value to get encoded size for
      * @return the number of bytes for encoded value
      */
-    private static int sizeOfUnsignedVarInt64(long value) {
+    public static int sizeOfUnsignedVarInt64(long value) {
```
artemananiev, line 828: Do these two methods need to be public, given that `sizeOfVarIn32/64()` is now available?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -1250,14 +1270,9 @@ public static int sizeOfEnumList(FieldDefinition field, List<? extends EnumWithP
...
-        if (!field.oneOf() && list.isEmpty()) {
-            return 0;
-        }
         int size = 0;
-
         for (final String value : list) {
-            size += sizeOfTag(field, ProtoConstants.WIRE_TYPE_DELIMITED);
+            size += sizeOfTag(field, WIRE_TYPE_DELIMITED);
             final int strSize = sizeOfStringNoTag(field, value);
             size += sizeOfVarInt32(strSize) + strSize;
```
artemananiev, line 1277: Can be rewritten as
```java
size += sizeOfDelimited(field, sizeOfStringNoTag(field, value));
```

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -1052,7 +1072,7 @@ public static int sizeOfString(FieldDefinition field, String value) {
             return 0;
         }
         final int size = sizeOfStringNoTag(field, value);
-        return sizeOfTag(field, ProtoConstants.WIRE_TYPE_DELIMITED) + sizeOfVarInt32(size) + size;
+        return sizeOfTag(field, WIRE_TYPE_DELIMITED) + sizeOfVarInt32(size) + size;
```
artemananiev, line 1075: Use `sizeOfDelimited()` here?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -1052,7 +1072,7 @@ public static int sizeOfString(FieldDefinition field, String value) {
             return 0;
         }
         final int size = sizeOfStringNoTag(field, value);
```
artemananiev, line 1074: I wonder if `field` is actually needed in `sizeOfStringNoTag()` It's only used to check for oneOf() in case of empty strings, but for empty strings the size is zero anyway...

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -1293,14 +1304,14 @@ public static <T> int sizeOfMessageList(FieldDefinition field, List<T> list, ToI
...
-        }
         int size = 0;
         for (final RandomAccessData value : list) {
-            size += Math.toIntExact(sizeOfTag(field, ProtoConstants.WIRE_TYPE_DELIMITED) + sizeOfVarInt32(Math.toIntExact(value.length())) + value.length());
+            size += Math.toIntExact(sizeOfTag(field, WIRE_TYPE_DELIMITED) + sizeOfVarInt32(Math.toIntExact(value.length())) + value.length());
         }
         return size;
     }
+
+    public static int sizeOfDelimited(final FieldDefinition field, final int length) {
```
artemananiev, line 1314: 👍

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
```diff
@@ -0,0 +1,759 @@
...
+    }
+
+    @Test
+    void testWriteOptionalFloat() {
+        FieldDefinition definition = createOptionalFieldDefinition(FLOAT);
+        final float valToWrite = randomLargeInt();
+        writeOptionalFloat(bufferedData, definition, valToWrite);
+        bufferedData.flip();
+        assertTypeDelimitedTag(definition);
+        assertEquals(MAX_VAR_FLOAT_SIZE, bufferedData.readVarInt(false));
```
artemananiev, line 382: `MAX_VAR_FLOAT_SIZE` name is confusing. Floats are not variable, they are fixed 4 bytes. Optional floats are either 0 bytes (null value), or sizeOfDelimited(field, 4) bytes. `OPTIONAL_FLOAT_SIZE` would be a better name, in my opinion. Also, check my comment to `MAX_VAR_FLOAT_SIZE`.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
```diff
@@ -0,0 +1,759 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.random.RandomGenerator;
+
+class ProtoWriterToolsTest {
+
+    public static final int SIZE_OF_TAG = 1;
+    public static final int SIZE_OF_LENGTH_VAR = 1;
+    public static final int MAX_VAR_INT_SIZE = 6;
+    public static final int MAX_VAR_FLOAT_SIZE = 5;
```
artemananiev, line 81: What does this mean, exactly?

My math is: max value here is when field ID is max (like 2^29 - 1), in this case total size of non-null optional float will be 5 (delimited tag) + 5 (delimited length) + 4 (fixed float)

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
imalygin: You're right, they don't need to be. Fixed

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
imalygin: Done.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Utf8Tools.java
```diff
@@ -20,6 +20,9 @@ public class Utf8Tools {
      *     surrogates)
      */
     static int encodedLength(CharSequence sequence) throws IOException {
```
artemananiev, line 22: final

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Utf8Tools.java
```diff
@@ -20,6 +20,9 @@ public class Utf8Tools {
...
+        if(sequence == null) {
```
artemananiev, line 23: Add a whitespace between `if` and `(`

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
imalygin: Yes, it can be removed. I covered the case of null oneOf() case (that required a check for null in `Utf8Tools#encodedLength`, after that the field could be safely removed.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoParserToolsTest.java
```diff
@@ -46,7 +67,7 @@ void testReadInt64() {
     @Test
     void testReadUint32() {
         testRead(() ->
-                rng.nextInt(0, Integer.MAX_VALUE),
+                rng.nextInt(0, MAX_VALUE),
```
artemananiev, line 70: I like `Integer.MAX_VALUE` more than just `MAX_VALUE`. What if `Integer.MAX_VALUE ` and `Long.MAX_VALUE` are used in the same class? Leaving it up to you, though.

imalygin: Yes, qualified name makes sense in this case

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
```diff
@@ -0,0 +1,759 @@
...
+        writeOptionalFloat(bufferedData, definition, valToWrite);
+        bufferedData.flip();
+        assertTypeDelimitedTag(definition);
+        assertEquals(MAX_VAR_FLOAT_SIZE, bufferedData.readVarInt(false));
+        assertFixed32Tag(definition.type().optionalFieldDefinition);
+        assertEquals(valToWrite, bufferedData.readFloat(ByteOrder.LITTLE_ENDIAN));
+    }
+
+    @Test
+    void testWriteOptionalFloat_null() {
```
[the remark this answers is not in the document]
imalygin: Yes, it does. I added tests.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
imalygin: I renamed the constants

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
imalygin: Yes, it's confusing. Renamed the constant and used a sum of `FLOAT_SIZE` and `TAG_SIZE`

artemananiev (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Utf8Tools.java
imalygin: Fixed

artemananiev (approved):

PR #138, opened 2023-12-08
fix 68: introduce SingleFuzzTest

### LINKED ISSUE #68: Add Tests for Malformed Data
### Problem

### Solution

- Very deep nested messages
- Large Lists
- Probably Others

anthony-swirldslabs:
**Description**:
Introducing the `SingleFuzzTest` class to support fuzz testing.

This is the first in a series of pull requests that will ultimately build a framework to run fuzz testing for pbj.

Here I'm introducing a new utility class that can perform a single randomized run of a fuzz test and return a result value that indicates whether the test was successful or more testing may be needed. See javadocs for more details.

Please note that building the entire fuzz testing framework is outside of scope of this particular PR and will be done in future, separate PRs.

**Related issue(s)**:

Fixes #68 

**Notes for reviewer**:
```
pbj-integration-tests]$ ./gradlew test --info
...
STANDARD_OUT
    SingleFuzzTest 0: Object: AccountID[shardNum=3, realmNum=2, account=OneOf[kind=ACCOUNT_NUM, value=1]]
    SingleFuzzTest 0: Bytes: ByteArrayBufferedData[8,3,16,2,24,1]
    SingleFuzzTest 0: Fuzz bytes: ByteArrayBufferedData[8,3,16,2,223,1]
    SingleFuzzTest 0: Fuzz exception: Cannot understand wire_type of 7
    A fuzz test for AccountID[shardNum=3, realmNum=2, account=OneOf[kind=ACCOUNT_NUM, value=1]] resulted in DESERIALIZATION_FAILED
...
```

**Checklist**

### DIFF
--- a/pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/SingleFuzzTest.java
+++ b/pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/SingleFuzzTest.java
@@ -56,7 +56,7 @@ public final class SingleFuzzTest {
             dataBuffer = write(object, codec, size);
         } catch (Exception ex) {
             // The test expects a valid input object, so we don't expect this to happen here
-            throw new RuntimeException("Unable to write the object", ex);
+            throw new FuzzTestException("Unable to write the object", ex);
         }
 
         System.out.println(prefix + "Bytes: " + dataBuffer);
--- a/pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/SingleFuzzTest.java
+++ b/pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/SingleFuzzTest.java
@@ -1,0 +1,109 @@
+package com.hedera.pbj.integration.fuzz;
+
+import com.hedera.pbj.runtime.Codec;
+import com.hedera.pbj.runtime.io.buffer.BufferedData;
+
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A single fuzz test.
+ * <p>
+ * This class exposes a single public static method that allows a caller to pass
+ * a valid object, a Protobuf codec for this object, as well as a random numbers
+ * generator. The method will then perform a single fuzz test run with the given
+ * data and return a SingleFuzzTestResult describing the outcome of the test run.
+ * <p>
+ * A fuzz testing framework is expected to use this class to execute multiple runs
+ * of the same test (each being random per the given random numbers generator)
+ * until the set of outcomes satisfies the testing framework threshold,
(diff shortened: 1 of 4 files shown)


pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/SingleFuzzTest.java
```diff
@@ -0,0 +1,109 @@
...
+ * or the framework runs out of the maximum number of attempts, or a timeout
+ * occurs.
+ * <p>
+ * This class is thread-safe and can be called concurrently from multiple threads
+ * as long as the input parameters are immutable or thread-safe.
+ */
+public final class SingleFuzzTest {
+    private final static AtomicInteger TEST_ID_GENERATOR = new AtomicInteger(0);
+
+    private static <T> BufferedData write(final T object, final Codec<T> codec, final int size) throws Exception {
```
artemananiev, line 29: As far as I remember, `Codec` has a method to measure size of object, in bytes. So `size` here is probably redundant.

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/SampleFuzzTest.java
```diff
@@ -0,0 +1,35 @@
...
+        AccountID accountID = AccountID.newBuilder().accountNum(1).realmNum(2).shardNum(3).build();
+        SingleFuzzTestResult singleFuzzTestResult = SingleFuzzTest.fuzzTest(
+                accountID,
+                AccountID.PROTOBUF,
+                new Random()
+        );
+        System.out.println("A fuzz test for " + accountID + " resulted in " + singleFuzzTestResult);
+
+        // It's a no-op by design currently. See javadoc for details.
+        assertNotNull(singleFuzzTestResult);
```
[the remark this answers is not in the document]
anthony-swirldslabs: That's correct. However, Codacy insists that a test must have a call to assert*() and/or fail(), so I had to add this no-op assertion in the second commit (hence the comment right above it.)

pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/SingleFuzzTest.java
anthony-swirldslabs: You're correct. However, the code calling this method has already called codec.measureRecord() at lines 53 and 84 below. The implementation of the measureRecord() is such that it iterates over all the fields of an object, which may be rather expensive for large objects. Since I'll need to run this test many times in order to gather statistics (because a single fuzz test run may not reveal issues), I figured I'd optimize it a bit here and not call measureRecord() the second time since I already have the value. Please let me know what you think.

artemananiev (approved):

PR #153, opened 2023-12-16
fix 68: extend fuzz test to more models

### LINKED ISSUE #68: Add Tests for Malformed Data

anthony-swirldslabs:
**Description**:
1. Adding about 8 more test models to run the fuzz test against. The total number of test objects from their `ARGUMENTS` is a little over 300.
2. Splitting the `integTest` and `fuzzTest` Gradle targets to run the fuzz tests with assertions off, and making the regular `test` target depend on both.
3. Minor adjustments to heuristics used in the fuzz test.

Please note that the `THRESHOLD` mechanism will be overhauled in a future PR where I'll replace the individual tests failures with gathering of statistics across all runs, and then deciding if the test as a whole should fail or not. But this is OUTSIDE OF SCOPE of this particular PR.

**Related issue(s)**:

Fixes #68 

**Notes for reviewer**:
1. The output of the test is similar as before but it now tests against more than 300 test objects.
2. The test takes a couple of minutes to run currently. I'll be looking at optimizing it a bit in a future PR.

**Checklist**

### DIFF
--- a/pbj-integration-tests/build.gradle.kts
+++ b/pbj-integration-tests/build.gradle.kts
@@ -59,7 +59,6 @@ testing {
         useJUnitJupiter()
         targets.all {
             testTask {
-                useJUnitPlatform()
                 // We are running a lot of tests 10s of thousands, so they need to run in parallel
                 systemProperties["junit.jupiter.execution.parallel.enabled"] = true
                 systemProperties["junit.jupiter.execution.parallel.mode.default"] = "concurrent"
@@ -95,13 +94,9 @@ testing {
     }
 }
 
-tasks.test.configure() {
+tasks.test {
     actions.clear()
     dependsOn("integTest", "fuzzTest")
-}
-
-tasks.test {
-    useJUnitPlatform()
 }
 
 jmh {
--- a/pbj-integration-tests/build.gradle.kts
+++ b/pbj-integration-tests/build.gradle.kts
@@ -79,8 +79,6 @@ tasks.withType<Test>().configureEach {
     systemProperties["junit.jupiter.execution.parallel.mode.default"] = "concurrent"
     // us parallel GC to keep up with high temporary garbage creation, and allow GC to use 40% of CPU if needed
     jvmArgs("-XX:+UseParallelGC", "-XX:GCTimeRatio=90")
-//                jvmArgs("-XX:+UseZGC","-XX:ZAllocationSpikeTolerance=2")
-//                jvmArgs("-XX:+UseG1GC", "-XX:GCTimeRatio=90", "-XX:MaxGCPauseMillis=100")
     // Some also need more memory
     minHeapSize = "512m"
     maxHeapSize = "4096m"
--- a/pbj-integration-tests/build.gradle.kts
+++ b/pbj-integration-tests/build.gradle.kts
@@ -57,8 +57,9 @@ testing {
     @Suppress("UnstableApiUsage")
     suites.getByName<JvmTestSuite>("test") {
         useJUnitJupiter()
-        targets.configureEach {
+        targets.all {
             testTask {
+                useJUnitPlatform()
                 // We are running a lot of tests 10s of thousands, so they need to run in parallel
                 systemProperties["junit.jupiter.execution.parallel.enabled"] = true
                 systemProperties["junit.jupiter.execution.parallel.mode.default"] = "concurrent"
@@ -71,7 +72,32 @@ testing {
                 maxHeapSize = "4096m"
             }
         }
+
+        tasks.register<Test>("integTest") {
+            testClassesDirs = sources.output.classesDirs
+            classpath = sources.runtimeClasspath
+
+            shouldRunAfter(tasks.test)
+
+            useJUnitPlatform { excludeTags("FUZZ_TEST") }
+            enableAssertions = true
+        }
+
+        tasks.register<Test>("fuzzTest") {
+            testClassesDirs = sources.output.classesDirs
+            classpath = sources.runtimeClasspath
+
+            shouldRunAfter(tasks.test)
+
+            useJUnitPlatform { includeTags("FUZZ_TEST") }
+            enableAssertions = false
+        }
     }
+}
+
+tasks.test.configure() {
+    actions.clear()
+    dependsOn("integTest", "fuzzTest")
 }
 
 tasks.test {
--- a/pbj-integration-tests/build.gradle.kts
+++ b/pbj-integration-tests/build.gradle.kts
@@ -57,36 +57,10 @@ testing {
     @Suppress("UnstableApiUsage")
     suites.getByName<JvmTestSuite>("test") {
         useJUnitJupiter()
-        targets.all {
-            testTask {
-                // We are running a lot of tests 10s of thousands, so they need to run in parallel
-                systemProperties["junit.jupiter.execution.parallel.enabled"] = true
-                systemProperties["junit.jupiter.execution.parallel.mode.default"] = "concurrent"
-                // us parallel GC to keep up with high temporary garbage creation, and allow GC to use 40% of CPU if needed
-                jvmArgs("-XX:+UseParallelGC", "-XX:GCTimeRatio=90")
-//                jvmArgs("-XX:+UseZGC","-XX:ZAllocationSpikeTolerance=2")
-//                jvmArgs("-XX:+UseG1GC", "-XX:GCTimeRatio=90", "-XX:MaxGCPauseMillis=100")
-                // Some also need more memory
-                minHeapSize = "512m"
-                maxHeapSize = "4096m"
-            }
-        }
-
-        tasks.register<Test>("integTest") {
-            testClassesDirs = sources.output.classesDirs
-            classpath = sources.runtimeClasspath
-
-            shouldRunAfter(tasks.test)
-
-            useJUnitPlatform { excludeTags("FUZZ_TEST") }
-            enableAssertions = true
-        }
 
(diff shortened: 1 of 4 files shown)


pbj-integration-tests/build.gradle.kts
```diff
@@ -57,8 +57,9 @@ testing {
     @Suppress("UnstableApiUsage")
     suites.getByName<JvmTestSuite>("test") {
         useJUnitJupiter()
-        targets.configureEach {
+        targets.all {
             testTask {
+                useJUnitPlatform()
```
jjohannes, line 62:
```suggestion
```

This is already set by `useJUnitJupiter()` above. See: https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html#differences_between_similar_methods_on_jvmtestsuite_and_test_task_types

pbj-integration-tests/build.gradle.kts
```diff
@@ -71,9 +72,34 @@ testing {
...
+
+            shouldRunAfter(tasks.test)
+
+            useJUnitPlatform { includeTags("FUZZ_TEST") }
+            enableAssertions = false
+        }
     }
 }
 
+tasks.test.configure() {
```
jjohannes, line 98:
```suggestion
tasks.test {
```

Does the same, but is more concise.

pbj-integration-tests/build.gradle.kts
```diff
@@ -71,9 +72,34 @@ testing {
...
+    actions.clear()
```
jjohannes, line 99: This is a bit of a nasty hack. 🤔 Have you considered using this task – `test` – for the "normal" integration tests and not add an additional `integTest` task? Because the tests are in `test` and usually it is expected that the task `test` executes the tests in that directory. I think this also leads to unexpected behavior when executing a single test from IDEA (if you care about that).

You could also consider to:
- Move all tests into a folder called `integTest` and then register a new test suite for that folder.
- (Maybe) Move the FuzzTests into a separate folder called `fuzzTest` and register a test suite for that. Then you do not need the filtering by tag.

Registering the test suites automatically sets up the tasks with the corresponding names. The Gradle configuration would look like this:

```
testing {
    suites { 
        register<JvmTestSuite>("integTest") { 
            dependencies {
                implementation(project()) 
            }
            targets.all {
                tasks.test { dependsOn(testTask) }
            }
        }
        register<JvmTestSuite>("fuzzTest") { 
            dependencies {
                implementation(project()) 
            }
            targets.all {
                testTask { enableAssertions = false }
                tasks.test { dependsOn(testTask) }
            }
        }
    }
}
```

pbj-integration-tests/build.gradle.kts
```diff
@@ -71,9 +72,34 @@ testing {
...
+        }
+
+        tasks.register<Test>("fuzzTest") {
+            testClassesDirs = sources.output.classesDirs
+            classpath = sources.runtimeClasspath
+
+            shouldRunAfter(tasks.test)
+
+            useJUnitPlatform { includeTags("FUZZ_TEST") }
+            enableAssertions = false
```
jjohannes, line 93: Why is this needed?

jjohannes (changes requested):

pbj-integration-tests/build.gradle.kts
anthony-swirldslabs: Will remove.

pbj-integration-tests/build.gradle.kts
anthony-swirldslabs: Will change.

pbj-integration-tests/build.gradle.kts
anthony-swirldslabs: Thanks for the suggestion! That indeed looks cleaner. I'll try to implement this and see how it goes.

pbj-integration-tests/build.gradle.kts
anthony-swirldslabs: Addressed in latest commit.

pbj-integration-tests/build.gradle.kts
anthony-swirldslabs: Addressed in latest commit.

pbj-integration-tests/build.gradle.kts
anthony-swirldslabs: Unfortunately, I didn't manage to get this working. For one, all the jvm args had to be repeated in each suite definition separately, even though the `enableAssertions` is the only difference between the two. I also had to extend the configurations for the new targets from ordinary targets similar to 

https://docs.gradle.org/current/userguide/java_testing.html#sec:configuring_java_integration_tests

otherwise Gradle wouldn't run them.

But apart from these minor things, ultimately I was unable to make the Gradle script compile the generated source at `pbj-integration-tests/build/generated/source/pbj-proto/test/` and then include the compiled classes into the classpath of the `fuzzTest` target. I'm sure I missed something trivial, but due to my poor expertise in Gradle unfortunately I was unable to find a solution for this.

While I agree that the solution in the current fix where we redefine the `test` target doesn't look conventional, it's pretty straightforward from the Gradle script perspective and gets the job done as far as running both the regular integ and the new fuzz tests is concerned. However, if you know how to solve the issues I described above, I'd be happy to learn.

jjohannes: Thanks for giving it a try. If the `generated/source/pbj-proto/test/` makes this complicated (I didn't think much about that) then I agree that we should keep things in `src/test`.

My remark on keeping the setup this way is that I would then keep using `test` to run the normal integration tests. Which means **not** do `actions.clear()` and not do an additional `tasks.register<Test>("integTest")`

If you keep things as they are now, you could remove the whole

```
    targets.all {
            testTask {
                // We are running a lot of tests 10s of thousands, so they need to run in parallel
                systemProperties["junit.jupiter.execution.parallel.enabled"] = true
                systemProperties["junit.jupiter.execution.parallel.mode.default"] = "concurrent"
                // us parallel GC to keep up with high temporary garbage creation, and allow GC to use 40% of CPU if needed
                jvmArgs("-XX:+UseParallelGC", "-XX:GCTimeRatio=90")
//                jvmArgs("-XX:+UseZGC","-XX:ZAllocationSpikeTolerance=2")
//                jvmArgs("-XX:+UseG1GC", "-XX:GCTimeRatio=90", "-XX:MaxGCPauseMillis=100")
                // Some also need more memory
                minHeapSize = "512m"
                maxHeapSize = "4096m"
            }
        }

```

block. Because `targets.all { testTask {} }` is the same as `test.tasks {}`. The `all` might suggest that you configure more than one task, but it is not the case. It only configures the task associated with the test suite, and there is always only one – https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html#test_suite_target. The ability to add a new task to a test suite is still a missing feature. I know this part is quite confusing. ☹️ 

What you probably want to do is (if the `jvmArgs ` and `memory` settings are needed? Seems to work without as well 😄):

```
tasks.withType<Test>().configureEach {
   // ...
}
```

anthony-swirldslabs: Thanks for the tip! In the latest commit, I'm only adding the `fuzzTest` target and making the `test` depend on it, and it all seems to work fine, and it looks a lot nicer in the Gradle script itself. Please have a look.

jjohannes (approved):

pbj-integration-tests/build.gradle.kts
jjohannes: Perfect. Gradle part looks good to me now!

pbj-integration-tests/build.gradle.kts
```diff
@@ -57,25 +57,33 @@ testing {
...
+    dependsOn("fuzzTest")
+}
+
+tasks.withType<Test>().configureEach {
+    // We are running a lot of tests 10s of thousands, so they need to run in parallel
+    systemProperties["junit.jupiter.execution.parallel.enabled"] = true
+    systemProperties["junit.jupiter.execution.parallel.mode.default"] = "concurrent"
+    // us parallel GC to keep up with high temporary garbage creation, and allow GC to use 40% of CPU if needed
+    jvmArgs("-XX:+UseParallelGC", "-XX:GCTimeRatio=90")
+//                jvmArgs("-XX:+UseZGC","-XX:ZAllocationSpikeTolerance=2")
```
imalygin, line 82: please remove these commented lines if they are not intended to be used later or provide some context in which uncommenting these lines may be useful.

pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/FuzzTest.java
```diff
@@ -37,6 +37,18 @@ public static <T> FuzzTestResult<T> fuzzTest(final T object, final double thresh
         final Random random = new Random();
```
imalygin, line 37: Please, print out the seed of this `Random` object.

anthony-swirldslabs: Random doesn't have a getter for its seed.
I'm working on creating a Random with a fixed seed right now, but that will come in a separate PR.

pbj-integration-tests/build.gradle.kts
anthony-swirldslabs: Please note that I merely moved these lines from another place. However, I agree that having a commented code isn't ideal, so I removed it in the latest commit. Please have a look.

imalygin (approved):

PR #155, opened 2023-12-19
fix 68: fix the Random seed and overhaul thresholds

### LINKED ISSUE #68: Add Tests for Malformed Data

anthony-swirldslabs:
**Description**:
1. By default, the `fuzzTest` will use a constant seed for a single Random object used for all the fuzz tests. And these tests now run sequentially. This will ensure that the tests don't fail randomly. And this also allows us to reproduce the failures if they occur.
2. A new `randomFuzzTest` Gradle target is added to enable running the tests with a random seed for discovering new issues. This new target must be run manually if/when desired.
3. The threshold values and their application to the results of the tests has been totally overhauled. The test now gathers all the results and computes statistics over the individual pass/failure statuses as well as the individual `DESERIALIZATION_FAILED` rates. We pass or fail the fuzz test as a whole if the pass rate across all the models drops below 90%, or the mean of the `DESERIALIZATION_FAILED` rate across all the models drops below 90%. Now that an individual test failure won't fail the whole test, the individual `THRESHOLD` value has also been bumped up to 80%.

We can tweak the thresholds values further in the future, but I think that the current values look pretty good already.

Please see javadocs/code comments for more details.

**Related issue(s)**:

Fixes #68 

**Notes for reviewer**:
```
$ gr fuzzTest --info
...
Fuzz tests are configured to use a FIXED seed for `new Random(seed)`, and the seed value for this run is: 837582698436792
...
Fuzz tests PASSED with passRate = 94% and deserializationFailedMean = 94%
...
```

**Checklist**

### DIFF
--- a/pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/FuzzTest.java
+++ b/pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/FuzzTest.java
@@ -30,11 +30,10 @@ public class FuzzTest {
      * for the most desirable DESERIALIZATION_FAILED outcome to determine
      * if the test passed or not.
      */
-    public static <T> FuzzTestResult<T> fuzzTest(final T object, final double threshold) {
+    public static <T> FuzzTestResult<T> fuzzTest(final T object, final double threshold, final Random random) {
         final long startNanoTime = System.nanoTime();
 
         final Codec<T> codec = getCodec(object);
-        final Random random = new Random();
         final int repeatCount = estimateRepeatCount(object, codec);
 
         if (repeatCount == 0) {
@@ -50,7 +49,8 @@ public class FuzzTest {
         }
 
         final Map<SingleFuzzTestResult, Long> resultCounts = IntStream.range(0, repeatCount)
-                .parallel()
+                // Note that we must run this stream sequentially to enable
+                // reproducing the tests for a given random seed.
                 .mapToObj(n -> SingleFuzzTest.fuzzTest(object, codec, random))
                 .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
 
--- a/pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/SampleFuzzTest.java
+++ b/pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/SampleFuzzTest.java
@@ -36,6 +36,10 @@ import static org.junit.jupiter.api.Assumptions.assumeFalse;
  * is considered passed or failed.
  */
 public class SampleFuzzTest {
+    // Flip to true to print out results stats for every tested model object.
+    // When false, only the fuzz test summary is printed to stdout.
+    private static final boolean debug = false;
+
     /**
      * A percentage threshold for the share of DESERIALIZATION_FAILED outcomes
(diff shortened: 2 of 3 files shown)


pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/SampleFuzzTest.java
```diff
@@ -63,18 +103,76 @@ private static Stream<?> objectTestCases() {
...
-        String resultDescription = fuzzTestResult.format();
-        System.out.println(resultDescription);
-        assertTrue(fuzzTestResult.passed(), resultDescription);
+        final Random random = buildRandom();
+
+        final List<? extends FuzzTestResult<?>> results = objectTestCases()
+                // Note that we must run this stream sequentially to enable
+                // reproducing the tests for a given random seed.
+                .map(object -> FuzzTest.fuzzTest(object, THRESHOLD, random))
+                .peek(result -> System.out.println(result.format()))
```
artemananiev, line 138: Should it be enabled in debug mode only?

pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/FuzzTest.java
```diff
@@ -50,7 +49,8 @@ public static <T> FuzzTestResult<T> fuzzTest(final T object, final double thresh
         }
 
         final Map<SingleFuzzTestResult, Long> resultCounts = IntStream.range(0, repeatCount)
-                .parallel()
+                // Note that we must run this stream sequentially to enable
+                // reproducing the tests for a given random seed.
```
artemananiev, line 53: Running tests in parallel is very beneficial to keep test time low. In case of "fixed random", it is possible for every test to use its own `Random` instance, all built from the same seed? Would it be fully deterministic?

pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/SampleFuzzTest.java
anthony-swirldslabs: Good idea. Just did that in the last comment.

pbj-integration-tests/src/main/java/com/hedera/pbj/integration/fuzz/FuzzTest.java
anthony-swirldslabs: I think this isn't a major concern currently:

```
$ gr fuzzTest --info
...
    Fuzz tests PASSED with passRate = 94% and deserializationFailedMean = 94%
...
> Task :fuzzTest
...
BUILD SUCCESSFUL in 47s
```

Let's solve that when that becomes a problem.

anthony-swirldslabs: And yeah, gr is defined as:

```
alias gr='./gradlew --rerun-tasks'
```

So it rebuilt everything related to this target, and sill managed in less than a minute.

imalygin (approved):

jjohannes (approved):
👍 for the Gradle part

PR #158, opened 2023-12-22
feat: Added support of compareTo

### LINKED ISSUE #103: Certain generated classes need to be Comparable
### Problem

As noted in [a recent PR](https://github.com/hashgraph/hedera-services/pull/8749) in hedera-services, we have a substantial number of protobuf-defined Java objects that need to be sorted or used in sorted data structures.
It would be exceptionally helpful if those objects could implement `Comparable`.
To that end we would like a mechanism to enable selected protobuf-defined objects to implement Comparable.

### Solution

- Add a directive that marks a particular `message` as needing to implement `Comparable`
    - this might be similar to the package directive, but placed within a `message` block. Perhaps something like this (for AccountID):
        - `// <<<pbj.comparable = "shardNum,realmNum,accountNum,alias">>>`
    - The fields listed would be compared in the order listed in the generated `compareTo` method.
- Adjust the PBJ compiler to add the `Comparable` interface to object generated from `message` blocks thus annotated, and generate the `compareTo` method as requested.
- It would be a compilation error to apply this directive to any `message` body that has fields that are not either comparable protobuf types (e.g. numeric or string) or messages that _also_ contain a `pbj.comparable` directive.
- If necessary for implementation efficiency, limiting this directive to `message` blocks that include *only* numeric and/or string fields is reasonable in the short-term.

### Alternatives

Continue to create ad-hoc `Comparator` implementations that often fail to take into account all fields, leaving open a possibility of significant errors and/or rework should the network need to enable multiple shards, change certain undocumented assumptions, or enable use of multiple realms.

imalygin:
PR highlights: 
- by default all fields of a class are used in `compareTo` method. 
- `pbj.comparable` option in a comment can be used to specify limited set of fields to be used in `compareTo`. 
```
// <<<pbj.comparable = "int32Number, text, subObject" >>>
```
- there was a small modification in Protobuf grammar definition to support it
- `Bytes` and `BytesValue` fields are compared using byte-by-byte comparison 
- repeatable fields are not supported

Misc:
- refactored ModelGenerator: extracted code-generating code into several independent methods
- upgraded Gradle version to 8.5 to support Java 21
- added dependency to Mockito framework in pbj-compiler

**Related issue(s)**:

Fixes #103

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
@@ -644,7 +644,7 @@ public final class Common {
 						return;
 					}
 				}
-				throw new IllegalArgumentException(("Field %s.%s specified in `pbj.comparable` option has is supposed to implement " +
+				throw new IllegalArgumentException(("Field %s.%s specified in `pbj.comparable` option must implement " +
 						"`Comparable` interface but it doesn't.").formatted(className, field.nameCamelFirstLower()));
 			} catch (IOException e) {
 				throw new RuntimeException(e);
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
@@ -2,8 +2,12 @@ package com.hedera.pbj.compiler.impl;
 
 import com.hedera.pbj.compiler.impl.grammar.Protobuf3Parser;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import org.jetbrains.annotations.NotNull;
 
+import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
@@ -183,7 +187,7 @@ public final class Common {
 	 * @return The generated code for getting the hashCode value.
 	 */
 	public static String getFieldsHashCode(final List<Field> fields, String generatedCodeSoFar) {
-		for(Field f : fields) {
+		for (Field f : fields) {
 			if (f.parent() != null) {
 				final OneOfField oneOfField = f.parent();
 				generatedCodeSoFar += getFieldsHashCode(oneOfField.fields(), generatedCodeSoFar);
@@ -347,7 +351,7 @@ public final class Common {
 	 * @return The generated code for getting the object equality
 	 */
 	public static String getFieldsEqualsStatements(final List<Field> fields, String generatedCodeSoFar) {
-		for(Field f : fields) {
+		for (Field f : fields) {
 			if (f.parent() != null) {
 				final OneOfField oneOfField = f.parent();
 				generatedCodeSoFar += getFieldsEqualsStatements(oneOfField.fields(), generatedCodeSoFar);
@@ -504,111 +508,153 @@ public final class Common {
 
 	/**
 	 * Generate the compareTo method content for the provided fields
-	 * @param fields The fields of this object.
-	 * @param generatedCodeSoFar the generated code so far (non-empty in case of nested objects)
+	 *
+	 * @param fields                The fields of this object.
+	 * @param generatedCodeSoFar    the generated code so far (non-empty in case of nested objects)
+	 * @param destinationSrcDir a directory where the previously generated code is saved
 	 * @return The generated code for compareTo method body
 	 */
-	public static String getFieldsCompareToStatements(final List<Field> fields, String generatedCodeSoFar) {
-		for(Field f : fields) {
-			if (f.parent() != null) {
-				final OneOfField oneOfField = f.parent();
-				generatedCodeSoFar += getFieldsCompareToStatements(oneOfField.fields(), generatedCodeSoFar);
-			}
-
+	public static String getFieldsCompareToStatements(final List<Field> fields, String generatedCodeSoFar, File destinationSrcDir) {
+		for (Field f : fields) {
 			if (f.optionalValueType()) {
 				generatedCodeSoFar += getPrimitiveWrapperCompareToGeneration(f);
-			}
-			else if (f.repeated()) {
-				// intentionally skipping repeated fields
+			} else if (f.repeated()) {
+				throw new UnsupportedOperationException("Repeated fields are not supported in compareTo method");
 			} else {
-				f.nameCamelFirstLower();
 				if (f.type() == Field.FieldType.FIXED32 ||
 						f.type() == Field.FieldType.INT32 ||
 						f.type() == Field.FieldType.SFIXED32 ||
-						f.type() == Field.FieldType.SINT32 ||
-						f.type() == Field.FieldType.UINT32) {
+						f.type() == Field.FieldType.SINT32) {
 					generatedCodeSoFar +=
 							"""
-                            result = Integer.compare($fieldName, thatObj.$fieldName);
-                            if (result != 0) {
-                                return result;
-                            }
-                            """.replace("$fieldName", f.nameCamelFirstLower());
+							result = Integer.compare($fieldName, thatObj.$fieldName);
+							if (result != 0) {
+							    return result;
+							}
+							""".replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.UINT32) {
+						generatedCodeSoFar +=
+       						"""
+							result = Integer.compareUnsigned($fieldName, thatObj.$fieldName);
+							if (result != 0) {
+							    return result;
+							}
+							""".replace("$fieldName", f.nameCamelFirstLower());
+
 				} else if (f.type() == Field.FieldType.FIXED64 ||
 						f.type() == Field.FieldType.INT64 ||
 						f.type() == Field.FieldType.SFIXED64 ||
-						f.type() == Field.FieldType.SINT64 ||
-						f.type() == Field.FieldType.UINT64) {
+						f.type() == Field.FieldType.SINT64) {
 					generatedCodeSoFar +=
 							"""
-                            result = Long.compare($fieldName, thatObj.$fieldName);
-                            if (result != 0) {
-                                return result;
-                            }
-                            """.replace("$fieldName", f.nameCamelFirstLower());
+							result = Long.compare($fieldName, thatObj.$fieldName);
+							if (result != 0) {
+							    return result;
+							}
+							""".replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.UINT64) {
+					generatedCodeSoFar +=
+							"""
+							result = Long.compareUnsigned($fieldName, thatObj.$fieldName);
+							if (result != 0) {
+							    return result;
+							}
+							""".replace("$fieldName", f.nameCamelFirstLower());
 				} else if (f.type() == Field.FieldType.BOOL) {
 					generatedCodeSoFar +=
 							"""
-                            result = Boolean.compare($fieldName, thatObj.$fieldName);
-                            if (result != 0) {
-                                return result;
-                            }
-                             """.replace("$fieldName", f.nameCamelFirstLower());
+							result = Boolean.compare($fieldName, thatObj.$fieldName);
+							if (result != 0) {
+							    return result;
+							}
+							""".replace("$fieldName", f.nameCamelFirstLower());
 				} else if (f.type() == Field.FieldType.FLOAT) {
 					generatedCodeSoFar +=
 							"""
-                            result = Float.compare($fieldName, thatObj.$fieldName);
-                            if (result != 0) {
-                                return result;
-                            }
-                             """.replace("$fieldName", f.nameCamelFirstLower());
+							result = Float.compare($fieldName, thatObj.$fieldName);
+							if (result != 0) {
+							    return result;
+							}
+							""".replace("$fieldName", f.nameCamelFirstLower());
 				} else if (f.type() == Field.FieldType.DOUBLE) {
 					generatedCodeSoFar +=
 							"""
-                            result = Double.compare($fieldName, thatObj.$fieldName);
-                            if (result != 0) {
-                                return result;
-                            }
-                             """.replace("$fieldName", f.nameCamelFirstLower());
-				} else if(f.type() == Field.FieldType.BYTES) {
-					generatedCodeSoFar +=
-							"""
-							// byte-to-byte comparison is most likely impractical, so we just compare the length
-							if ($fieldName == null && thatObj.$fieldName != null) {
-								return -1;
-							} else if ($fieldName != null && thatObj.$fieldName == null) {
-								return 1;
-							} else if ($fieldName != null) {
-								result = (int) ($fieldName.length() - thatObj.$fieldName.length());
-							}
-							if(result != 0) {
-								return result;
+							result = Double.compare($fieldName, thatObj.$fieldName);
+							if (result != 0) {
+							     return result;
 							}
 							""".replace("$fieldName", f.nameCamelFirstLower());
 				} else if (f.type() == Field.FieldType.STRING ||
-						f.type() == Field.FieldType.ENUM ||
-						f.parent() == null /* Process a sub-message */) {
-					generatedCodeSoFar += (
+						f.type() == Field.FieldType.ENUM) {
+					generatedCodeSoFar += generateCompareToForObject(f);
+				} else if (f.type() == Field.FieldType.BYTES) {
+					generatedCodeSoFar +=
 							"""
-                            if ($fieldName == null && thatObj.$fieldName != null) {
-                                return -1;
-                            }
-                            if ($fieldName != null &&  thatObj.$fieldName == null) {
-                                return 1;
-                            }
-                            if ($fieldName != null) {
-                                result = $fieldName.compareTo(thatObj.$fieldName);
-                            }
-                            if(result != 0) {
-                                return result;
-                            }
-                            """).replace("$fieldName", f.nameCamelFirstLower());
-				}  else {
-					throw new IllegalArgumentException("Unexpected field type for getting HashCode - " + f.type().toString());
+							result = $fieldName.compareTo(thatObj.$fieldName);
+							if (result != 0) {
+								return result;
+							}
+							 """.replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.MESSAGE || f.type() == Field.FieldType.ONE_OF) {
+					verifyComparable(f, destinationSrcDir);
+					generatedCodeSoFar += generateCompareToForObject(f);
+				}else {
+					throw new IllegalArgumentException("Unexpected field type for getting CompareTo - " + f.type().toString());
 				}
 			}
 		}
 		return generatedCodeSoFar.indent(DEFAULT_INDENT * 2);
+	}
+
+	@NotNull
+	private static String generateCompareToForObject(Field f) {
+		return """
+				if ($fieldName == null && thatObj.$fieldName != null) {
+				    return -1;
+				}
+				if ($fieldName != null && thatObj.$fieldName == null) {
+				    return 1;
+				}
+				if ($fieldName != null) {
+				    result = $fieldName.compareTo(thatObj.$fieldName);
+				}
+				if (result != 0) {
+				    return result;
+				}
+				""".replace("$fieldName", f.nameCamelFirstLower());
+	}
+
+	/**
+	 * Verify that the field is comparable.
+	 * @param field The field to verify.
+	 * @param destinationSrcDir The directory where the previously generated code is saved.
+	 */
+	private static void verifyComparable(final Field field, File destinationSrcDir) {
+		if (field instanceof final SingleField singleField) {
+			if (singleField.type() != Field.FieldType.MESSAGE) {
+				// everything else except message and bytes is comparable for sure
+				return;
+			}
+			// let's check if the message implements Comparable
+			final String className = singleField.javaFieldType();
+			final File javaFile = getJavaFile(destinationSrcDir, singleField.messageTypeModelPackage(), className);
+			try (BufferedReader reader = new BufferedReader(new FileReader(javaFile))) {
+				String line;
+				while ((line = reader.readLine()) != null) {
+					if (line.contains("implements Comparable")) {
+						return;
+					}
+				}
+				throw new IllegalArgumentException(("Field %s.%s specified in `pbj.comparable` option has is supposed to implement " +
+						"`Comparable` interface but it doesn't.").formatted(className, field.nameCamelFirstLower()));
+			} catch (IOException e) {
+				throw new RuntimeException(e);
+			}
+        } if (field instanceof final OneOfField oneOfField) {
+			oneOfField.fields().forEach(v -> verifyComparable(v, destinationSrcDir));
+		} else {
+			throw new UnsupportedOperationException("Unexpected field type - " + field.getClass());
+		}
 	}
 
 	/**
@@ -626,7 +672,7 @@ public final class Common {
                     } else if ($fieldName != null) {
                         result = $compareStatement;
                     }
-                    if(result != 0) {
+                    if (result != 0) {
                         return result;
                     }
                     """;
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
@@ -340,11 +340,11 @@ public final class Common {
 	}
 
 	/**
-	 * Recursively calculates the hashcode for a message fields.
+	 * Recursively calculates `equals` statement for a message fields.
 	 *
 	 * @param fields The fields of this object.
 	 * @param generatedCodeSoFar The accumulated hash code so far.
-	 * @return The generated code for getting the hashCode value.
+	 * @return The generated code for getting the object equality
 	 */
 	public static String getFieldsEqualsStatements(final List<Field> fields, String generatedCodeSoFar) {
 		for(Field f : fields) {
@@ -417,7 +417,7 @@ public final class Common {
                             }
                             """).replace("$fieldName", f.nameCamelFirstLower());
                 } else {
-                    throw new IllegalArgumentException("Unexpected field type for getting HashCode - " + f.type().toString());
+                    throw new IllegalArgumentException("Unexpected field type for getting Equals - " + f.type().toString());
                 }
             }
 		}
@@ -499,6 +499,168 @@ public final class Common {
 		}
 		""").replace("$fieldName", f.nameCamelFirstLower());
 		return generatedCodeSoFar;
+	}
+
+
+	/**
+	 * Generate the compareTo method content for the provided fields
+	 * @param fields The fields of this object.
+	 * @param generatedCodeSoFar the generated code so far (non-empty in case of nested objects)
+	 * @return The generated code for compareTo method body
+	 */
+	public static String getFieldsCompareToStatements(final List<Field> fields, String generatedCodeSoFar) {
+		for(Field f : fields) {
+			if (f.parent() != null) {
+				final OneOfField oneOfField = f.parent();
+				generatedCodeSoFar += getFieldsCompareToStatements(oneOfField.fields(), generatedCodeSoFar);
+			}
+
+			if (f.optionalValueType()) {
+				generatedCodeSoFar += getPrimitiveWrapperCompareToGeneration(f);
+			}
+			else if (f.repeated()) {
+				// intentionally skipping repeated fields
+			} else {
+				f.nameCamelFirstLower();
+				if (f.type() == Field.FieldType.FIXED32 ||
+						f.type() == Field.FieldType.INT32 ||
+						f.type() == Field.FieldType.SFIXED32 ||
+						f.type() == Field.FieldType.SINT32 ||
+						f.type() == Field.FieldType.UINT32) {
+					generatedCodeSoFar +=
+							"""
+                            result = Integer.compare($fieldName, thatObj.$fieldName);
+                            if (result != 0) {
+                                return result;
+                            }
+                            """.replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.FIXED64 ||
+						f.type() == Field.FieldType.INT64 ||
+						f.type() == Field.FieldType.SFIXED64 ||
+						f.type() == Field.FieldType.SINT64 ||
+						f.type() == Field.FieldType.UINT64) {
+					generatedCodeSoFar +=
+							"""
+                            result = Long.compare($fieldName, thatObj.$fieldName);
+                            if (result != 0) {
+                                return result;
+                            }
+                            """.replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.BOOL) {
+					generatedCodeSoFar +=
+							"""
+                            result = Boolean.compare($fieldName, thatObj.$fieldName);
+                            if (result != 0) {
+                                return result;
+                            }
+                             """.replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.FLOAT) {
+					generatedCodeSoFar +=
+							"""
+                            result = Float.compare($fieldName, thatObj.$fieldName);
+                            if (result != 0) {
+                                return result;
+                            }
+                             """.replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.DOUBLE) {
+					generatedCodeSoFar +=
+							"""
+                            result = Double.compare($fieldName, thatObj.$fieldName);
+                            if (result != 0) {
+                                return result;
+                            }
+                             """.replace("$fieldName", f.nameCamelFirstLower());
+				} else if(f.type() == Field.FieldType.BYTES) {
+					generatedCodeSoFar +=
+							"""
+							// byte-to-byte comparison is most likely impractical, so we just compare the length
+							if ($fieldName == null && thatObj.$fieldName != null) {
+								return -1;
+							} else if ($fieldName != null && thatObj.$fieldName == null) {
+								return 1;
+							} else if ($fieldName != null) {
+								result = (int) ($fieldName.length() - thatObj.$fieldName.length());
+							}
+							if(result != 0) {
+								return result;
+							}
+							""".replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.STRING ||
+						f.type() == Field.FieldType.ENUM ||
+						f.parent() == null /* Process a sub-message */) {
+					generatedCodeSoFar += (
+							"""
+                            if ($fieldName == null && thatObj.$fieldName != null) {
+                                return -1;
+                            }
+                            if ($fieldName != null &&  thatObj.$fieldName == null) {
+                                return 1;
+                            }
+                            if ($fieldName != null) {
+                                result = $fieldName.compareTo(thatObj.$fieldName);
+                            }
+                            if(result != 0) {
+                                return result;
+                            }
+                            """).replace("$fieldName", f.nameCamelFirstLower());
+				}  else {
+					throw new IllegalArgumentException("Unexpected field type for getting HashCode - " + f.type().toString());
+				}
+			}
+		}
+		return generatedCodeSoFar.indent(DEFAULT_INDENT * 2);
+	}
+
+	/**
+	 * Generates the compareTo code for a primitive wrapper field.
+	 * @param f The field for which to generate the compareTo code.
+	 * @return The generated code for compareTo method body
+	 */
+	private static String getPrimitiveWrapperCompareToGeneration(Field f) {
+		final String template =
+				"""
+                    if ($fieldName == null && thatObj.$fieldName != null) {
+                        return -1;
+                    } else if ($fieldName != null && thatObj.$fieldName == null) {
+                        return 1;
+                    } else if ($fieldName != null) {
+                        result = $compareStatement;
+                    }
+                    if(result != 0) {
+                        return result;
+                    }
+                    """;
+
+
+		final String compareStatement;
+		switch (f.messageType()) {
+			case "StringValue" ->
+					compareStatement =
+							"$fieldName.compareTo(thatObj.$fieldName)";
+			case "BoolValue" ->
+					compareStatement =
+							"java.lang.Boolean.compare($fieldName, thatObj.$fieldName)";
+			case "Int32Value", "UInt32Value" ->
+					compareStatement =
+							"java.lang.Integer.compare($fieldName, thatObj.$fieldName)";
+			case "Int64Value", "UInt64Value" ->
+					compareStatement =
+							"java.lang.Long.compare($fieldName, thatObj.$fieldName)";
+			case "FloatValue" ->
+					compareStatement =
+							"java.lang.Float.compare($fieldName, thatObj.$fieldName)";
+			case "DoubleValue" ->
+					compareStatement =
+							"java.lang.Double.compare($fieldName, thatObj.$fieldName)";
+			case "BytesValue" ->
+					compareStatement =
+							"(int)($fieldName.length() - thatObj.$fieldName.length())";
+			default -> throw new UnsupportedOperationException("Unhandled optional message type:" + f.messageType());
+		}
+
+		return template
+				.replace("$compareStatement", compareStatement)
+				.replace("$fieldName", f.nameCamelFirstLower());
 	}
 
 	/**
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
@@ -2,7 +2,6 @@ package com.hedera.pbj.compiler.impl;
 
 import com.hedera.pbj.compiler.impl.grammar.Protobuf3Parser;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import org.jetbrains.annotations.NotNull;
 
 import java.io.BufferedReader;
 import java.io.File;
@@ -606,7 +605,7 @@ public final class Common {
 		return generatedCodeSoFar.indent(DEFAULT_INDENT * 2);
 	}
 
-	@NotNull
+	@NonNull
 	private static String generateCompareToForObject(Field f) {
 		return """
 				if ($fieldName == null && thatObj.$fieldName != null) {
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
@@ -10,6 +10,7 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 /**
@@ -32,9 +33,10 @@ public final class Common {
 	public static final int TYPE_LENGTH_DELIMITED = 2;
 	/** Wire format code for fixed 32bit number */
 	public static final int TYPE_FIXED32 = 5;
+    private static final Pattern COMPARABLE_PATTERN = Pattern.compile("class\\s+\\w+\\s+implements\\s+Comparable");
 
 
-	/**
+    /**
 	 * Makes a tag value given a field number and wire type.
 	 *
 	 * @param wireType the wire type part of tag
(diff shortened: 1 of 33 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -253,77 +131,155 @@ public void generate(final Protobuf3Parser.MessageDefContext msgDef,
...
+	 * @param fields The fields to filter
+	 * @return the filtered fields
+	 */
+	@NonNull
+	private static List<Field> filterComparableFields(final MessageDefContext msgDef,
+													  final ContextualLookupHelper lookupHelper,
+													  final List<Field> fields) {
+		final Set<String> comparableFields = lookupHelper.getComparableFields(msgDef);
+		if(comparableFields.isEmpty()) {
+			// consider all fields in compareTo by defalut
```
anthony-swirldslabs, line 232: This seems a bit extreme, especially if the current implementation doesn't support comparisons of certain field types as discussed above. I'd suggest not to implement Comparable for models that don't have the optional annotation comment added explicitly. This would simplify the implementation as well as reduce the footprint of the generated Java classes (both source, and more importantly, the binaries) for the model codecs.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -253,77 +131,155 @@ public void generate(final Protobuf3Parser.MessageDefContext msgDef,
...
+			return fields;
+		}
+		return fields.stream().filter(f -> comparableFields.contains(f.nameCamelFirstLower())).collect(Collectors.toList());
```
anthony-swirldslabs, line 235: Now that I got here, I see why you used Sets above. However, for the reasons from previous comments, I'd suggest to use Lists there, and then this method wouldn't be needed at all. This is because the model owner may want to specify a particular order of fields for the purpose of comparison, which may not always be identical to the order in which the fields are defined in the model (based on the business use-case etc.)

anthony-swirldslabs (changes requested):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +501,168 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+							"""
+                            result = Double.compare($fieldName, thatObj.$fieldName);
+                            if (result != 0) {
+                                return result;
+                            }
+                             """.replace("$fieldName", f.nameCamelFirstLower());
+				} else if(f.type() == Field.FieldType.BYTES) {
+					generatedCodeSoFar +=
+							"""
+							// byte-to-byte comparison is most likely impractical, so we just compare the length
```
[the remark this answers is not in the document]
jsync-swirlds: Even safer, if we cannot support comparison of fields, then do not implement `Comparable` for that message.  Generally most messages should not be `Comparable`, just those that explicitly request to be `Comparable`; and then, ideally, the request specifies the fields and order in which to compare them.
Failing to compile when a message requests to be `Comparable` but has non-comparable fields specified, helps the author make reasonable choices about what should or should not be `Comparable`.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +501,168 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+			if (f.parent() != null) {
+				final OneOfField oneOfField = f.parent();
+				generatedCodeSoFar += getFieldsCompareToStatements(oneOfField.fields(), generatedCodeSoFar);
+			}
+
+			if (f.optionalValueType()) {
+				generatedCodeSoFar += getPrimitiveWrapperCompareToGeneration(f);
+			}
+			else if (f.repeated()) {
+				// intentionally skipping repeated fields
```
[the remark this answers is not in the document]
jsync-swirlds: If we do not support repeated fields for comparison (which is entirely reasonable), then we should refuse to compile any message with repeated fields that requests to implement `Comparable`.  That makes it clear both for the message author (via a compile error) and for the user of the Java class that those objects should not be used in sorted structures.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +501,168 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
 
+
+	/**
+	 * Generate the compareTo method content for the provided fields
+	 * @param fields The fields of this object.
+	 * @param generatedCodeSoFar the generated code so far (non-empty in case of nested objects)
+	 * @return The generated code for compareTo method body
+	 */
+	public static String getFieldsCompareToStatements(final List<Field> fields, String generatedCodeSoFar) {
+		for(Field f : fields) {
```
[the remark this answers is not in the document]
jsync-swirlds: This codebase would probably benefit from running the same spotless rules as the hedera-services codebase for consistency.
Perhaps Nathan can help get that setup for this repo?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +501,168 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+			case "Int64Value", "UInt64Value" ->
+					compareStatement =
+							"java.lang.Long.compare($fieldName, thatObj.$fieldName)";
+			case "FloatValue" ->
+					compareStatement =
+							"java.lang.Float.compare($fieldName, thatObj.$fieldName)";
+			case "DoubleValue" ->
+					compareStatement =
+							"java.lang.Double.compare($fieldName, thatObj.$fieldName)";
+			case "BytesValue" ->
```
jsync-swirlds, line 655: I believe this will fail to work correctly for many of the current `bytes` values in hedera-services messages (e.g. [ContractID](https://github.com/hashgraph/hedera-protobufs/blob/add-pbj-types-for-state/services/basic_types.proto#L155C13-L155C13)), which will need byte-wise comparison of the array content in order to be deterministic.
It's entirely reasonable to refuse to implement `Comparable` for any message with a `bytes` field (and refuse to compile that if `Comparable` is requested), at least initially, but comparing on length only seems very likely to create hard-to-debug ISS events.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +501,168 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+			} else {
+				f.nameCamelFirstLower();
+				if (f.type() == Field.FieldType.FIXED32 ||
+						f.type() == Field.FieldType.INT32 ||
+						f.type() == Field.FieldType.SFIXED32 ||
+						f.type() == Field.FieldType.SINT32 ||
+						f.type() == Field.FieldType.UINT32) {
+					generatedCodeSoFar +=
+							"""
+                            result = Integer.compare($fieldName, thatObj.$fieldName);
```
artemananiev, line 532: I wonder how `Integer.compare()` handles unsigned 32-bit ints...

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
```diff
@@ -49,5 +49,13 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(kind, value);
     }
+
+    @Override
+    public int compareTo(OneOf<E> thatObj) {
+        if(thatObj == null) {
+            return 1;
+        }
+        return kind.compareTo(thatObj.kind);
```
artemananiev, line 58: If kinds are equal, shouldn't values be compared next?

pbj-integration-tests/src/main/proto/comparable.proto
```diff
@@ -0,0 +1,46 @@
...
+}
+
+message ComparableSubObj {
+  int32 num = 1;
+}
+
+/**
+ * Example of a message with a limited set of fields that are comparable.
+ */
+// <<<pbj.comparable = "int32Number, text, subObject" >>>
```
artemananiev, line 37: This should lead to a failure in build time, since `ComparableSubObj` is not comparable (as noted in many comments above, messages should be comparable only if explicitly annotated with pbj.comparable)

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
jsync-swirlds: It doesn't (at least not correctly).
Here's the full code of Integer.compare (it's quite simple):
`return (x < y) ? -1 : ((x == y) ? 0 : 1);`
It compares integers as signed.
If the values are unsigned, then this would be:
`result = Integer.compareUnsigned($fieldName, thatObj.$fieldName);`

The long.compare code below should be similar with unsigned fields separated to use `Long.compareUnsigned()`.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
imalygin: @jsync-swirlds I tried applying spotless rules here, and that resulted in a disaster with indentations in the generated code. So, I don't think it's going to be that straightforward.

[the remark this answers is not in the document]
imalygin: @anthony-swirldslabs 

> we do expect the fields list to be in the exact order in which the fields are declared in the model definition, right?

It's not necessarily an expectation. It just happened to be this way, it's a reasonable default. As long as this standard applies consistently (and it always should the case), there should be no issues
However, you raised a good point. What if we have predefined set of fields we want to use in `compareTo`? Right now the order is not preserved, and it's not good. It should be.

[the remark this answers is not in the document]
jsync-swirlds: Regarding spotless, it's not setup on this repo currently.  I would not expect it to work until that is done (which probably requires assistance from Nathan or another member of that team).

I am just suggesting that (aside from this PR) it might be worthwhile to get that effort on the priority list so it can be done at some point.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
imalygin: @anthony-swirldslabs 

>one can always sort all the repeated values in both the objects, and then compare the sorted arrays lexicographically.

Even though it's certainly possible, it seems impractical for performance reasons. 

> we could throw an exception

Yes, I will modify `LookupHelper#extractComparableFields` accordingly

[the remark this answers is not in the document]
imalygin: @jsync-swirlds 
> we should refuse to compile any message with repeated fields that requests to implement `Comparable`
If the request is explicit, I agree - it should throw an exception. By default, I think it's reasonable to omit it from `compareTo`.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +501,168 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+				generatedCodeSoFar += getFieldsCompareToStatements(oneOfField.fields(), generatedCodeSoFar);
+			}
+
+			if (f.optionalValueType()) {
+				generatedCodeSoFar += getPrimitiveWrapperCompareToGeneration(f);
+			}
+			else if (f.repeated()) {
+				// intentionally skipping repeated fields
+			} else {
+				f.nameCamelFirstLower();
```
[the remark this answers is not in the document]
imalygin: Probably, some refactoring artifact. Not even sure how it got there. I'll remove it

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
imalygin: @anthony-swirldslabs 
I had doubts about this one, but it looks like it would be more consistent just not support byte array comparison at all. 

@jsync-swirlds 
>Generally most messages should not be Comparable,

I thought the general agreement was to make them all comparable by default and customize the list of fields where needed.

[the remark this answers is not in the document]
jsync-swirlds: That is not a general agreement I am aware of; perhaps the discussion was a bit under-communicated.  Also, it will break a lot of necessary assumptions to implement universally.
I am not confident it is possible to make everything Comparable and be _correct_ in that implementation.  Given that, implementing an incorrect or incomplete comparison method is worse than not implementing anything.  Only implementing Comparable when requested seems, to me, to be a good balance.

The Issue that this PR addresses, and the understanding I have, is that we will implement Comparable *when requested* and then only when a correct and complete implementation is reasonable; throwing a compile error if Comparable is requested but the message includes currently unsupported field types.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
jsync-swirlds: The problem with a "default" compareTo that short-cuts repeated fields like this is that there will soon be code that checks if Comparable is implemented; if it is implemented then that code must assume the implementation is _correct_.  This will not be correct by default as it will not impose a consistent total order (as required by the [Comparable interface](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Comparable.html)) on records with repeated fields.

Simply not supporting repeated fields (and only writing a compareTo if Comparable is explicitly requested) helps ensure we don't provide a comparison method to consuming code that is likely to cause unexpected failures.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
```diff
@@ -576,6 +592,35 @@ private void buildMessage(
         }
     }
 
+    /**
+     * Extract the set of fields that are comparable for a given message.
+     * @param msgDef The message defenition to get comparable fields for
+     * @return set of field names that are comparable
+     */
+    static Set<String> extractComparableFields(final MessageDefContext msgDef) {
```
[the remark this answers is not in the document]
imalygin: Agreed.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: Got it, thank you. I will remove support for Bytes

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
rbair23: > @jsync-swirlds I tried applying spotless rules here, and that resulted in a disaster with indentations in the generated code. So, I don't think it's going to be that straightforward.

Ya, we should do it as part of a separate PR. But it would be nice to get spotless going.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
rbair23: I think we should only implement Comparable on messages that have been annotated with the right comment and in that case we throw an error if the object cannot be made comparable.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
rbair23: > The Issue that this PR addresses, and the understanding I have, is that we will implement Comparable when requested and then only when a correct and complete implementation is reasonable; throwing a compile error if Comparable is requested but the message includes currently unsupported field type

This is also my expectation. I think this is safer, and if we find that we can in fact safely generate Comparable for all types in the future, then we can relax this. We really only need a few simple types to be Comparable in services code today, and those we can easily annotate and we can be sure they will be proper implementations.

BTW, the assumption that we can skip byte arrays is probably not true. The Account ID has a one-of third field that can be a byte array, and sorting by Account ID is definitely something we will require.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
jsync-swirlds: We may still need to support Bytes (ContractID comes to mind) eventually; but it can be a subsequent PR, and when we do we *must* implement byte-wise comparison in order to ensure a correct compareTo for messages with bytes fields.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +501,168 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+                        return result;
+                    }
+                    """;
+
+
+		final String compareStatement;
+		switch (f.messageType()) {
+			case "StringValue" ->
+					compareStatement =
+							"$fieldName.compareTo(thatObj.$fieldName)";
```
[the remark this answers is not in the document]
rbair23: I love switch expressions 😄

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
jsync-swirlds: The values would need to be compared in order to be correct.
If that's too much for this first-pass, we should reject any message with OneOf structures until that's supported (in a future PR, most likely).

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
rbair23: I think we will need the bytes for Account ID, otherwise we won't be able to use the feature in services.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
netopyr: I agree with @jsync-swirlds and @rbair23. In Services, we need only a few keys to be comparable. But if a record is comparable, we must be sure it is correct.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
netopyr: Right `AccountID` and `ContractID` both may contain `bytes`.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
netopyr: `AccountID` needs to be comparable and it contains a `OneOf` structure.

jsync-swirlds: I agree.  I also know this is quite a complex PR. If we can get a correct, but limited, result quickly, then add the OneOf and bytes handling in a follow-up; I believe that may be a good path forward.

Not my call, but perhaps worth considering.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: okay, I will add support for bytes then.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
imalygin: Okay, I'll try make it work in a separate PR

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
imalygin: @anthony-swirldslabs @jsync-swirlds @rbair23 

Fair enough. I made changes to the implementation.
The current state of things:
- `Bytes` field _is_ supported. These fields are compared byte-by-byte
- OneOf is now properly supported. All the kinds of `OneOf` must implement `Comparable`. There is a validation for that. If it's not the case, the build fails.
- The order of fields is `pbj.comparable` option is preserved. Now they are used in `compareTo` in the order they are listed.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
imalygin: Sounds good, I  made the change to the implementation

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +501,168 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+		final String template =
+				"""
+                    if ($fieldName == null && thatObj.$fieldName != null) {
+                        return -1;
+                    } else if ($fieldName != null && thatObj.$fieldName == null) {
+                        return 1;
+                    } else if ($fieldName != null) {
+                        result = $compareStatement;
+                    }
+                    if(result != 0) {
```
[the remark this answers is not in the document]
imalygin: Done

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
```diff
@@ -66,6 +76,9 @@ public final class LookupHelper {
     /** Map from proto file path to list of other proto files it imports */
     private final Map<String, Set<String>> protoFileImports = new HashMap<>();
 
+    /** Map from fully qualified msgDef name to set of field names that are comparable */
+    private final Map<String, Set<String>> comparableFieldsByMsg = new HashMap<>();
```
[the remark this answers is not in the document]
imalygin: Yes, fair. I fixed it

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
```diff
@@ -576,6 +592,35 @@ private void buildMessage(
         }
     }
 
+    /**
+     * Extract the set of fields that are comparable for a given message.
+     * @param msgDef The message defenition to get comparable fields for
+     * @return set of field names that are comparable
+     */
+    static Set<String> extractComparableFields(final MessageDefContext msgDef) {
+        if(msgDef.optionComment() == null || msgDef.optionComment().getText() == null) {
```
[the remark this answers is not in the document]
imalygin: Fixed

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: Fixed

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
imalygin: Fixed

pbj-integration-tests/src/main/proto/comparable.proto
imalygin: `ComparableSubObj` is comparable. However, I added a negative test case for what you're talking about - see `non_compilable_comparable_sub_obj.proto`

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -2,8 +2,12 @@
 
 import com.hedera.pbj.compiler.impl.grammar.Protobuf3Parser;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import org.jetbrains.annotations.NotNull;
```
jsync-swirlds, line 5: Incorrect annotation (we do not use the jetbrains annotations).  Might be worth adjusting IDE settings to prefer the correct annotation (one line above).

imalygin: Oops. Removed.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +504,210 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+							     return result;
+							}
+							""".replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.STRING ||
+						f.type() == Field.FieldType.ENUM) {
+					generatedCodeSoFar += generateCompareToForObject(f);
+				} else if (f.type() == Field.FieldType.BYTES) {
+					generatedCodeSoFar +=
+							"""
+							result = $fieldName.compareTo(thatObj.$fieldName);
```
anthony-swirldslabs, line 592: Minor: in theory this can throw an NPE if one of the objects didn't initialize its bytes field for whatever reason. This can happen via `Model.newBuilder().bytesField(null).build()`.

I'd suggest to simply use the generateCompareToForObject() that you've introduced here.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -347,6 +347,25 @@ public int read() throws IOException {
...
+        // Compare the lengths first
+        final int minLength = Math.min(length, otherData.length);
+        for (int i = 0; i < minLength; i++) {
+            final int compare = Byte.compare(getByte(i), otherData.getByte(i));
+            if (compare != 0) {
+                return compare;
+            }
+        }
+        // If all compared elements are equal, the shorter array is considered less
+        return length - otherData.length;
```
jsync-swirlds, line 366: Nit: Might be easier to test/verify if we return `Integer.compare(length, otherData.length)` here.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +504,210 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+							if (result != 0) {
+								return result;
+							}
+							 """.replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.MESSAGE || f.type() == Field.FieldType.ONE_OF) {
+					verifyComparable(f, destinationSrcDir);
+					generatedCodeSoFar += generateCompareToForObject(f);
+				}else {
```
anthony-swirldslabs, line 600: Wouldn't checking `data instanceof Comparable` directly in the generateCompareToForObject() for both the values suffice and be more generic (and allow to simplify this long if/elif chain, and avoid the destinationSrcDir argument, too, and remove the verifyComparable() altogether as well)?

Or all this complexity is only for performance? Do you have any numbers as to how much this saves (in percentages)? I'd guess instanceof checks should be pretty efficient in JVM.

Minor: a space before `else` please.

jsync-swirlds (commented):
Impressive effort so far, but there are a few minor issues, and some items that need to be fixed.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +504,210 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+				// everything else except message and bytes is comparable for sure
+				return;
+			}
+			// let's check if the message implements Comparable
+			final String className = singleField.javaFieldType();
+			final File javaFile = getJavaFile(destinationSrcDir, singleField.messageTypeModelPackage(), className);
+			try (BufferedReader reader = new BufferedReader(new FileReader(javaFile))) {
+				String line;
+				while ((line = reader.readLine()) != null) {
+					if (line.contains("implements Comparable")) {
```
anthony-swirldslabs, line 643: Just make sure nobody puts a comment to their .proto file with this exact substring! :)

Like mentioned above, I'd prefer to avoid this incomplete parsing method here if possible at all.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
anthony-swirldslabs: I still suggest to use the idiomatic switch expression here.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
anthony-swirldslabs: I believe this method still only compares the lengths for bytes at line 701 below.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
jsync-swirlds: Perhaps just open an issue for now?  It will probably take some effort, and isn't strictly related to the issue driving this PR.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
```diff
@@ -576,6 +593,52 @@ private void buildMessage(
...
+        final var matcher = OPTION_COMMENT.matcher(msgDef.optionComment().getText());
+        if (matcher.find()) {
+            final String optionName = matcher.group(1);
+            final String optionValue = matcher.group(2);
+            if (optionName.equals(PBJ_COMPARABLE_OPTION_NAME)) {
+                final Set<String> regularFieldNames = msgDef.messageBody().messageElement().stream()
+                        .filter(v -> v.field() != null)
+                        .filter(v -> {
+                            if (v.field().REPEATED() != null) {
+                                throw new IllegalArgumentException(("Field `%s` specified in `%s` option is repeated. " +
```
anthony-swirldslabs, line 614: I'm not sure if this stream goes over fields listed in the annotation. I believe this stream iterates over all the fields declared in the model. So this `throw` will activate even when the annotation doesn't request a repeated field to be included in the compareTo.

I suggest to filter by the values from the `optionValue` first, and only then perform other validations.

Note that you'd still need to additionally check that optionValue doesn't list any non-existing fields.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +505,210 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+				} else if (f.type() == Field.FieldType.UINT32) {
+						generatedCodeSoFar +=
+       						"""
+							result = Integer.compareUnsigned($fieldName, thatObj.$fieldName);
+							if (result != 0) {
+							    return result;
+							}
+							""".replace("$fieldName", f.nameCamelFirstLower());
+
+				} else if (f.type() == Field.FieldType.FIXED64 ||
```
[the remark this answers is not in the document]
imalygin: The following guides suggest that FIXED64 and SFIXED64 are mapped to 64bit integer

https://protobuf.dev/programming-guides/proto3/
https://microsoft.github.io/typespec/next/libraries/protobuf/reference/data-types

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +505,210 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+	 * @return The generated code for compareTo method body
+	 */
+	public static String getFieldsCompareToStatements(final List<Field> fields, String generatedCodeSoFar, File destinationSrcDir) {
+		for (Field f : fields) {
+			if (f.optionalValueType()) {
+				generatedCodeSoFar += getPrimitiveWrapperCompareToGeneration(f);
+			} else if (f.repeated()) {
+				throw new UnsupportedOperationException("Repeated fields are not supported in compareTo method");
+			} else {
+				if (f.type() == Field.FieldType.FIXED32 ||
```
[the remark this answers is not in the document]
imalygin: I looked into a couple of guides about Protobuf types, they suggest that FIXED32 and SFIXED32 are mapped to 32bit integer.

https://protobuf.dev/programming-guides/proto3/
https://microsoft.github.io/typespec/next/libraries/protobuf/reference/data-types

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -501,6 +505,210 @@ private static String getRepeatedEqualsGeneration(String generatedCodeSoFar, Fie
...
+	@NotNull
+	private static String generateCompareToForObject(Field f) {
+		return """
+				if ($fieldName == null && thatObj.$fieldName != null) {
+				    return -1;
+				}
+				if ($fieldName != null && thatObj.$fieldName == null) {
+				    return 1;
+				}
+				if ($fieldName != null) {
```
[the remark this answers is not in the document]
imalygin: If I understand your suggestion correctly, this condition would ensure that the field is indeed `Comparable`. That is not necessary because the field is one of the following:
1. a primitive which maps to a comparable class (including String and Enum)
2. a wrapper for a primitive, we unwrap it and then use `compareTo` of the primitive
3. An object of a type that was verified by `Common#verifyComparable`. If the object doesn't implement `Comparable` and yet it's listed as a field that has to be included into `compareTo` of the class, there will be a build failure
See `CompareToNegativeTest#testNonComparableSubObj`

Unless I'm missing something, this explicit cast would be unnecessary

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
jsync-swirlds: Hmm...  Protobuf seems to mix metaphors a lot (and I read the encoding information slightly differently).  Thanks for clarifying.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
jsync-swirlds: Hmm...  Protobuf seems to mix metaphors a lot (and I read the encoding information slightly differently).  Thanks for clarifying.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
jsync-swirlds: That's why I asked the question.  It's not immediately clear if the _apparent_ type at the point this code is generated is known, by the compiler, to be `Comparable`.
Thanks for clarifying.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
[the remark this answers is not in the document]
imalygin: Sounds good. Sorry, I missed the comment. It was hidden by GitHub

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -88,140 +92,15 @@ public void generate(final Protobuf3Parser.MessageDefContext msgDef,
...
-				final String enumComment = """
-									/**
-									 * Enum for the type of "%s" oneof value
-									 */""".formatted(oneOfField.name());
-				final String enumString = createEnum(enumComment ,"",enumName,maxIndex,enumValues, true)
-						.indent(DEFAULT_INDENT * 2);
-				oneofEnums.add(enumString);
-				fields.add(oneOfField);
-				imports.add("com.hedera.pbj.runtime");
+				oneofGetters.addAll(generateCodeForOneOf(lookupHelper, item, javaRecordName, imports, oneofEnums, fields));
```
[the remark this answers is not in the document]
imalygin: In this case the refactoring would fall into a category of tasks with low importance and likely be neglected.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
```diff
@@ -29,7 +30,10 @@
 import java.nio.file.Path;
 import java.util.*;
```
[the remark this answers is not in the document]
imalygin: Agreed

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
```diff
@@ -14,7 +14,7 @@
  * @param value    The actual value in the "oneof". May be null.
  * @param <E>      The enum type
  */
-public record OneOf<E>(E kind, Object value) {
+public record OneOf<E extends Enum<E>>(E kind, Object value) implements Comparable<OneOf<E>> {
```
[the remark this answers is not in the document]
imalygin: Fair enough. I added another representation of `oneOf` field - `ComparableOneOf` which is, as the name suggests, is specifically for comparable values.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
```diff
@@ -49,5 +49,17 @@ public boolean equals(Object o) {
...
+    @Override
+    public int compareTo(OneOf<E> thatObj) {
+        if (thatObj == null) {
+            return 1;
+        }
+        final int kindCompare = kind.compareTo(thatObj.kind);
+        if (kindCompare != 0) {
+            return kindCompare;
+        }
+        return ((Comparable) value).compareTo(thatObj.value);
```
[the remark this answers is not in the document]
imalygin: Yes, I refactored it.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: Yes, it's a good idea, thank you

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: If `verifyComparable` fails, it will result in build-time failure, and that's exactly what I want to have. In the generated code there will be a guarantee that the objects implement `Comparable` .

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
imalygin: Done

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: Fair enough. I used `class\s+\w+\s+implements\s+Comparable` regex instead

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -677,31 +672,16 @@ private static String getPrimitiveWrapperCompareToGeneration(Field f) {
...
-					compareStatement =
-							"(int)($fieldName.length() - thatObj.$fieldName.length())";
+		final String compareStatement = switch (f.messageType()) {
+			case "StringValue" -> "$fieldName.compareTo(thatObj.$fieldName)";
+			case "BoolValue" -> "java.lang.Boolean.compare($fieldName, thatObj.$fieldName)";
+			case "Int32Value", "UInt32Value" -> "java.lang.Integer.compare($fieldName, thatObj.$fieldName)";
+			case "Int64Value", "UInt64Value" -> "java.lang.Long.compare($fieldName, thatObj.$fieldName)";
+			case "FloatValue" -> "java.lang.Float.compare($fieldName, thatObj.$fieldName)";
+			case "DoubleValue" -> "java.lang.Double.compare($fieldName, thatObj.$fieldName)";
+			case "BytesValue" -> "(int)($fieldName.length() - thatObj.$fieldName.length())";
```
jsync-swirlds, line 682: nit: This might be clearer as `java.lang.Byte.compareUnsigned($fieldName, thatObj.$fieldName)`

imalygin: It turned out that `BytesValue` translates to `com.hedera.pbj.runtime.io.buffer.Bytes` which is comparable now, so I refactored it to this:
```
case "StringValue", "BytesValue" -> "$fieldName.compareTo(thatObj.$fieldName)";
```

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
imalygin: @anthony-swirldslabs should be fixed now

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
imalygin: > So this throw will activate even when the annotation doesn't request a repeated field to be included in the compareTo.

Good catch. I will add a test for that.

imalygin: Done

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -347,6 +347,25 @@ public int read() throws IOException {
...
+     * Compare this {@link Bytes} object to another {@link Bytes} object. The comparison is done on a byte-by-byte
+     * @param otherData the object to be compared.
+     * @return a negative integer, zero, or a positive integer as this object is less than, equal to, or greater than
+     */
+    @Override
+    public int compareTo(Bytes otherData) {
+        // Compare the lengths first
+        final int minLength = Math.min(length, otherData.length);
+        for (int i = 0; i < minLength; i++) {
+            final int compare = Byte.compare(getByte(i), otherData.getByte(i));
```
jsync-swirlds, line 360: Should this be compareUnsigned?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
```diff
@@ -628,11 +638,15 @@ static List<String> extractComparableFields(final MessageDefContext msgDef) {
                 return Arrays.stream(optionValue.split(","))
                         .map(String::trim)
                         .peek(v -> {
+                            if(repeatedFields.contains(v)){
```
anthony-swirldslabs, line 641: Minor: whitespace

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
imalygin: This probably wouldn't matter much as long as it's consistently applied, however `ByteString` class from Protobuf contains `ByteString#UNSIGNED_LEXICOGRAPHICAL_COMPARATOR` which suggests that unsigned comparison is the way to go. So, I'll do as you propose.

jsync-swirlds (approved):
Excellent work, Ivan.

anthony-swirldslabs (approved):

PR #163, opened 2024-01-04
fix 68: improve parsing

### LINKED ISSUE #68: Add Tests for Malformed Data

anthony-swirldslabs:
**Description**:
Addressing a few parsing issues that led to discrepancies between Google Protobuf behavior and PBJ's. Specifically:
1. Re-implementing https://github.com/hashgraph/pbj/pull/47 to reduce the scope of the try/catch block that swallows EOFException to only do that at the beginning of reading a field. When this exception is thrown anywhere else during the parsing of the field, then the parser errors out too.
2. Adding length/position checks to the repeated fields parser.
3. Modifying the string parser to error out on invalid UTF-8 data.
4. Modifying the delimited field skipper to error out on negative length.
5. Modifying the skip() implementation to error out if there's not enough bytes remaining in the buffer to skip.

Also introducing a couple of helper classes - `Sneaky` and `UncheckedThrowingFunction`, as well as adding a new static field to PBJ models Test files to reference their corresponding Google Protoc model class. This latter one-liner part isn't technically needed in this fix, but I'll be using it in a future PR shortly, and I don't want to touch anything from pbj-core in that PR.

**Related issue(s)**:

Fixes #68 

**Notes for reviewer**:
Build passes and existing tests pass. In a future PR, the results of this fix will help improve the fuzz testing framework, but that's a part of that separate future PR.

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
@@ -114,54 +114,58 @@ class CodecParseMethodGenerator {
 
                     // -- PARSE LOOP ---------------------------------------------
                     // Continue to parse bytes out of the input stream until we get to the end.
-                    try {
-                        while (input.hasRemaining()) {
+                    while (input.hasRemaining()) {
+                        // Note: ReadableStreamingData.hasRemaining() won't flip to false
+                        // until the end of stream is actually hit with a read operation.
+                        // So we catch this exception here and **only** here, because an EOFException
+                        // anywhere else suggests that we're processing malformed data and so
+                        // we must re-throw the exception then.
+                        final int tag;
+                        try {
                             // Read the "tag" byte which gives us the field number for the next field to read
                             // and the wire type (way it is encoded on the wire).
-                            final int tag = input.readVarInt(false);
+                            tag = input.readVarInt(false);
+                        } catch (EOFException e) {
+                            // There's no more fields. Stop the parsing loop.
+                            break;
+                        }
 
-                            // The field is the top 5 bits of the byte. Read this off
-                            final int field = tag >>> TAG_FIELD_OFFSET;
+                        // The field is the top 5 bits of the byte. Read this off
+                        final int field = tag >>> TAG_FIELD_OFFSET;
 
-                            // Ask the Schema to inform us what field this represents.
-                            final var f = getField(field);
+                        // Ask the Schema to inform us what field this represents.
+                        final var f = getField(field);
 
-                            // Given the wire type and the field type, parse the field
-                            switch (tag) {
-                $caseStatements
-                                default -> {
-                                    // The wire type is the bottom 3 bits of the byte. Read that off
-                                    final int wireType = tag & TAG_WRITE_TYPE_MASK;
-                                    // handle error cases here, so we do not do if statements in normal loop
-                                    // Validate the field number is valid (must be > 0)
-                                    if (field == 0) {
-                                        throw new IOException("Bad protobuf encoding. We read a field value of " + field);
-                                    }
-                                    // Validate the wire type is valid (must be >=0 && <= 5). Otherwise we cannot parse this.
-                                    // Note: it is always >= 0 at this point (see code above where it is defined).
-                                    if (wireType > 5) {
-                                        throw new IOException("Cannot understand wire_type of " + wireType);
-                                    }
-                                    // It may be that the parser subclass doesn't know about this field
-                                    if (f == null) {
-                                        if (strictMode) {
-                                            // Since we are parsing is strict mode, this is an exceptional condition.
-                                            throw new UnknownFieldException(field);
-                                        } else {
-                                            // We just need to read off the bytes for this field to skip it and move on to the next one.
-                                            skipField(input, ProtoConstants.get(wireType));
-                                        }
+                        // Given the wire type and the field type, parse the field
+                        switch (tag) {
+            $caseStatements
+                            default -> {
+                                // The wire type is the bottom 3 bits of the byte. Read that off
+                                final int wireType = tag & TAG_WRITE_TYPE_MASK;
+                                // handle error cases here, so we do not do if statements in normal loop
+                                // Validate the field number is valid (must be > 0)
+                                if (field == 0) {
+                                    throw new IOException("Bad protobuf encoding. We read a field value of " + field);
+                                }
+                                // Validate the wire type is valid (must be >=0 && <= 5). Otherwise we cannot parse this.
+                                // Note: it is always >= 0 at this point (see code above where it is defined).
+                                if (wireType > 5) {
+                                    throw new IOException("Cannot understand wire_type of " + wireType);
+                                }
+                                // It may be that the parser subclass doesn't know about this field
+                                if (f == null) {
+                                    if (strictMode) {
+                                        // Since we are parsing is strict mode, this is an exceptional condition.
+                                        throw new UnknownFieldException(field);
                                     } else {
-                                        throw new IOException("Bad tag [" + tag + "], field [" + field + "] wireType [" + wireType + "]");
+                                        // We just need to read off the bytes for this field to skip it and move on to the next one.
+                                        skipField(input, ProtoConstants.get(wireType));
                                     }
+                                } else {
+                                    throw new IOException("Bad tag [" + tag + "], field [" + field + "] wireType [" + wireType + "]");
                                 }
                             }
                         }
-                    }
-                    catch (EOFException e) {
-                        // Do nothing.
-                        // This is a workaround of hasRemaining() returning true, even there are no more elements.
-                        // The subsequent readByte will throw EOFException if there ar no more elements.
                     }
                     return new $modelClassName($fieldsList);
                 }
@@ -215,13 +219,20 @@ class CodecParseMethodGenerator {
                 "field=" + fieldNum + " [" + field.name() + "] */ -> {\n");
         sb.append("""
 				// Read the length of packed repeated field data
-				final var length = input.readVarInt(false);
+				final long length = input.readVarInt(false);
+				if (input.remaining() < length) {
+				    throw new BufferUnderflowException();
+				}
 				final var beforeLimit = input.limit();
+				final long beforePosition = input.position();
 				input.limit(input.position() + length);
 				while (input.hasRemaining()) {
 				    $tempFieldName = addToList($tempFieldName,$readMethod);
 				}
-				input.limit(beforeLimit);"""
+				input.limit(beforeLimit);
+				if (input.position() != beforePosition + length) {
+				    throw new BufferUnderflowException();
+				}"""
                 .replace("$tempFieldName", "temp_" + field.name())
                 .replace("$readMethod", readMethod(field))
                 .indent(DEFAULT_INDENT)
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
@@ -208,18 +208,19 @@ public final class ProtoParserTools {
         if (input.remaining() < length) {
             throw new BufferUnderflowException();
         }
-        final byte[] bytes = new byte[length];
-        final long bytesRead = input.readBytes(bytes);
+        final ByteBuffer bb = ByteBuffer.allocate(length);
+        final long bytesRead = input.readBytes(bb);
         if (bytesRead != length) {
             throw new BufferUnderflowException();
         }
+        bb.rewind();
 
         try {
             // Shouldn't use `new String()` because we want to error out on malformed UTF-8 bytes.
             return StandardCharsets.UTF_8.newDecoder()
                     .onMalformedInput(CodingErrorAction.REPORT)
                     .onUnmappableCharacter(CodingErrorAction.REPORT)
-                    .decode(ByteBuffer.wrap(bytes))
+                    .decode(bb)
                     .toString();
         } catch (CharacterCodingException e) {
             throw new MalformedProtobufException("Malformed UTF-8 string encountered", e);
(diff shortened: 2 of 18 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -114,55 +114,59 @@ static String generateParseInternalMethod(final String modelClassName, final Lis
...
+                                if (field == 0) {
+                                    throw new IOException("Bad protobuf encoding. We read a field value of " + field);
+                                }
+                                // Validate the wire type is valid (must be >=0 && <= 5). Otherwise we cannot parse this.
+                                // Note: it is always >= 0 at this point (see code above where it is defined).
+                                if (wireType > 5) {
+                                    throw new IOException("Cannot understand wire_type of " + wireType);
+                                }
+                                // It may be that the parser subclass doesn't know about this field
+                                if (f == null) {
```
artemananiev, line 156: Not strictly related to this PR, but somewhat related... Current code is, basically:
```java
FieldDefinition f = getField(field);
switch (tag) {
    // number of cases for known fields
    default: {
        if (f == null) {
            // throw an exception, if strict mode, skip otherwise
        }
        throw new IOException()
    }
}
```

I wonder if `f == null` check should be right after `getField()`. Does it really make sense to parse a value, if the field is unknown?

anthony-swirldslabs: I guess this is because the current implementation skips unknown fields in the default: case of the switch when running in non-strict mode. That case can technically be moved before the switch right after we fetch the field definition. However, I don't see any benefits of doing so because the logic of the code will remain exactly the same, and the clarity (or absence thereof) won't change. Also, this will remove the default: branch from the switch which isn't really ideal because this may lead to hidden bugs. So I'd prefer to keep this part as it is now. Let me know what you think.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -211,7 +214,16 @@ public static String readString(final ReadableSequentialData input) {
             throw new BufferUnderflowException();
         }
 
-        return new String(bytes,StandardCharsets.UTF_8);
+        try {
+            // Shouldn't use `new String()` because we want to error out on malformed UTF-8 bytes.
+            return StandardCharsets.UTF_8.newDecoder()
+                    .onMalformedInput(CodingErrorAction.REPORT)
+                    .onUnmappableCharacter(CodingErrorAction.REPORT)
+                    .decode(ByteBuffer.wrap(bytes))
```
artemananiev, line 222: This, unfortunately, creates an extra ByteBuffer object. I can't think of how to avoid it, though

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
```diff
@@ -151,14 +151,16 @@ public byte readByte() {
     /** {@inheritDoc} */
     @Override
     public long skip(final long n) {
-        final long clamped = Math.min(n, limit);
+        if (n > limit) {
```
artemananiev, line 154: Should it be `position + n > limit`?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
```diff
@@ -151,14 +151,16 @@ public byte readByte() {
     /** {@inheritDoc} */
     @Override
     public long skip(final long n) {
-        final long clamped = Math.min(n, limit);
```
artemananiev, line 154: This behavior is stated in `ReadableSequentialData.skip()`, so its JavaDoc should be updated.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
```diff
@@ -151,14 +151,16 @@ public byte readByte() {
...
+        }
 
-        if (clamped <= 0) {
+        if (n <= 0) {
             return 0;
         }
 
         try {
-            long numSkipped = in.skip(clamped);
+            long numSkipped = in.skip(n);
```
artemananiev, line 163: Does it make sense to check that `numSkipped` is equal to `n`?

artemananiev: Interesting... The method is allowed to skip less bytes than requested and return the number of actually skipped bytes. However, it might be only because the method never tried to skip past input stream limit. Now method contract is going to be changed (I assume), so if n is greater than the limit, then an exception is thrown. In this case, the method will be guaranteed to always skip the requested number of bytes or throw an exception, right? In this case, if `in.skip(n)` skipped less than `n`, it must result in an exception, too. What do you think?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/WritableStreamingData.java
```diff
@@ -100,11 +101,13 @@ public boolean hasRemaining() {
      * @return the actual number of bytes skipped.
      */
     @Override
-    public long skip(long count) {
+    public long skip(final long count) {
         try {
-            // We can only skip UP TO count, but if there are fewer bytes remaining, then we can only skip that many.
+            // We can only skip UP TO count.
             // And if the maximum bytes we can end up skipping is not positive, then we can't skip any bytes.
-            count = Math.min(count, remaining());
```
artemananiev, line 107: As with `ReadableSequentialData`, the current behavior is specified in `SequentialData.skip()` JavaDoc. It states that if the requested number of bytes is past the limit, bytes are skipped only up to the limit, but no exceptions are thrown. If we're going to change this and throw an exception, `SequentialData.skip()` JavaDoc should be updated.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/WritableStreamingData.java
```diff
@@ -100,11 +101,13 @@ public boolean hasRemaining() {
...
+            if (count > remaining()) {
+                throw new BufferUnderflowException();
```
artemananiev, line 109: s/BufferUnderflowException/BufferOverflowException/ , since this happens on writing rather than on reading.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialTestBase.java
```diff
@@ -154,9 +157,9 @@ void limitNotBeforePosition() {
     @DisplayName("Skipping more bytes than are available")
     void skipMoreThanAvailable() {
         final var stream = sequence("0123456789".getBytes(StandardCharsets.UTF_8));
-        assertThat(stream.skip(20)).isEqualTo(10);
-        assertThat(stream.hasRemaining()).isFalse();
-        assertThat(stream.remaining()).isZero();
+        assertThrows(BufferUnderflowException.class, () -> stream.skip(20));
+        assertThat(stream.hasRemaining()).isTrue();
+        assertThat(stream.remaining()).isEqualTo(10);
```
artemananiev, line 162: Is this really specified anywhere? I mean, if `skip()` throws an exception, is the current stream position guaranteed to be unchanged? It should be either added to JavaDoc, or this assertion is to be removed from the test.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
anthony-swirldslabs: True. I haven't found a public API that would allow me to decode bytes and throw exceptions w/o me having to create a ByteBuffer. Let's see if Oleg discovers any performance degradation because of this. If so, then the only other option would be to cut'n'paste the code from the OpenJDK String implementation, which I'd prefer not to do unless we have to.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
anthony-swirldslabs: Yep, it should.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
anthony-swirldslabs: It's actually `SequentialData.skip()`. Updated.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/ReadableStreamingData.java
anthony-swirldslabs: This seems to make sense in theory. However, the InputStream.skip() doc says:

```
The skip method may, for a variety of reasons, end up skipping over some smaller number of bytes, possibly 0. This may result from any of a number of conditions; reaching end of file before n bytes have been skipped is only one possibility.
```

So it may skip less than `n` bytes for a random, not necessarily bad reason.

So I'm unsure if we should error out on this, given that we honestly return the number of bytes actually skipped, allowing the caller to make its own decision.

Please let me know what you think.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/WritableStreamingData.java
anthony-swirldslabs: Yep, they both inherit from the SequentialData, so I've just updated that javadoc already.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/stream/WritableStreamingData.java
anthony-swirldslabs: Good catch. Fixed.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialTestBase.java
anthony-swirldslabs: That makes sense. I'll remove this assertion from the test because it's impossible to guarantee this behavior for all implementations.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
jsync-swirlds: The `new String` alternative used previously _copies_ the array, so I would argue that this new approach is significantly less pressure on memory.  ByteBuffer.wrap allocates a HeapByteBuffer that uses the existing array as its backing store, so it's a pretty small allocation by comparison.
If you want to avoid the call to `wrap`, I believe the following will work:
Instead of creating the byte[] on line 211, you could create a `ByteBuffer` there and use `long readBytes(@NonNull final ByteBuffer dst)` on the `ReadableSequentialData`.  Then you don't allocate the array separately, just the ByteBuffer that's needed here.  It doesn't avoid the ByteBuffer or reduce total allocations, but does make the code a little tidier.

anthony-swirldslabs: That's an awesome suggestion, @jsync-swirlds . Thanks!
I just implemented it and pushed it into this branch.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
artemananiev: My whole idea was to fail faster. If the field is unknown, it doesn't make sense to check the tag. The default handler will be preserved to throw an exception. For example, the field may be known, but its wire type is different for some reason. In this case, none of the switch cases will match, and the default handler will be used.

I agree this is not critical, but I'm still leaning towards moving the field check right before the switch statement.

anthony-swirldslabs: Fail fast is a good thing in general, I agree. Let me file a task under the PBJ Hardening epic to address this.

anthony-swirldslabs: I've just added it to https://github.com/hashgraph/pbj/issues/164

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/BufferedData.java
```diff
@@ -297,8 +297,10 @@ public long remaining() {
      * {@inheritDoc}
      */
     @Override
-    public long skip(long count) {
-        count = Math.min(count, buffer.remaining());
+    public long skip(final long count) {
+        if (count > Integer.MAX_VALUE || (int) count > buffer.remaining()) {
```
artemananiev, line 301: I wonder if the cast to (int) is really needed here

anthony-swirldslabs: I've added it for clarity. The compiler would perform the cast anyway because buffer.remaining() is int, but it could instead expand that int to a long. So I'm just explicitly telling the compiler to down-cast the count to int for the purpose of the comparison now that I've ensured that the value fits in an integer type in the previous condition.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
anthony-swirldslabs: @artemananiev : so, I closed https://github.com/hashgraph/pbj/issues/181 as WillNotFix. Please see the issue for more details.

PR #166, opened 2024-01-09
fix 125: add unit tests

### LINKED ISSUE #125: Identify unit testing gaps and fill them
Identify any gaps in unit test coverage and areas that need some more and fill them. Out of the three areas (code generator tool, runtime library, and generated code), the runtime library does need more coverage. The code generator tool probably needs some cleanup and maybe more tests for what remains.

```[tasklist]
### Tasks
- [x] Check that the new `hashcode()` and `equals()` methods are tested in the generated code, including corner cases
- [ ] #28
- [ ] To do - add more specific tasks
```

anthony-swirldslabs:
**Description**:
Adding some unit tests which should improve the coverage for PBJ.

**Related issue(s)**:

Fixes #125

**Notes for reviewer**:
```
pbj-core]$ gr test
...
BUILD SUCCESSFUL in 20s
```

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
+++ b/pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
@@ -59,24 +59,34 @@ import static java.util.Collections.emptyList;
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
+import com.hedera.pbj.runtime.io.WritableSequentialData;
 import com.hedera.pbj.runtime.io.buffer.BufferedData;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.hedera.pbj.runtime.io.buffer.RandomAccessData;
+import com.hedera.pbj.runtime.test.Sneaky;
 import net.bytebuddy.utility.RandomString;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.MethodSource;
 import test.proto.Apple;
 
 import java.io.IOException;
+import java.nio.BufferOverflowException;
 import java.nio.ByteOrder;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.random.RandomGenerator;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 class ProtoWriterToolsTest {
 
@@ -98,7 +108,7 @@ class ProtoWriterToolsTest {
 
     @BeforeEach
     void setUp() {
-        bufferedData = BufferedData.allocate(100);
+        bufferedData = BufferedData.allocate(200);
     }
 
     @Test
@@ -140,6 +150,16 @@ class ProtoWriterToolsTest {
         writeTag(bufferedData, definition, WIRE_TYPE_FIXED_64_BIT);
         bufferedData.flip();
         assertEquals((definition.number() << TAG_TYPE_BITS) | WIRE_TYPE_FIXED_64_BIT.ordinal(), bufferedData.readVarInt(false));
+    }
+
+    @Test
+    void testWriteInteger_zero() {
+        FieldDefinition definition = createFieldDefinition(INT32);
+        final int valToWrite = 0;
+        final long positionBefore = bufferedData.position();
+        writeInteger(bufferedData, definition, valToWrite);
+        final long positionAfter = bufferedData.position();
+        assertEquals(positionBefore, positionAfter);
     }
 
     @Test
@@ -190,6 +210,16 @@ class ProtoWriterToolsTest {
     void testWriteInteger_unsupported(FieldType type) {
         FieldDefinition definition = createFieldDefinition(type);
         assertThrows(RuntimeException.class, () -> writeInteger(bufferedData, definition, RNG.nextInt()));
+    }
+
+    @Test
+    void testWriteLong_Zero() {
+        FieldDefinition definition = createFieldDefinition(INT64);
+        final long valToWrite = 0L;
+        final long positionBefore = bufferedData.position();
+        writeLong(bufferedData, definition, valToWrite);
+        final long positionAfter = bufferedData.position();
+        assertEquals(positionBefore, positionAfter);
     }
 
     @Test
@@ -280,17 +310,42 @@ class ProtoWriterToolsTest {
         assertEquals(0, bufferedData.length());
     }
 
+    private static EnumWithProtoMetadata mockEnum(int ordinal) {
+        EnumWithProtoMetadata enumWithProtoMetadata = mock(EnumWithProtoMetadata.class);
+        doReturn(ordinal).when(enumWithProtoMetadata).protoOrdinal();
+        doReturn(RANDOM_STRING.nextString()).when(enumWithProtoMetadata).protoName();
+        return enumWithProtoMetadata;
+    }
+
     @Test
     void testWriteEnum() {
         FieldDefinition definition = createFieldDefinition(ENUM);
-        EnumWithProtoMetadata enumWithProtoMetadata = mock(EnumWithProtoMetadata.class);
-        int expectedOrdinal = RNG.nextInt();
-        when(enumWithProtoMetadata.protoOrdinal()).thenReturn(expectedOrdinal);
-        when(enumWithProtoMetadata.protoName()).thenReturn(RANDOM_STRING.nextString());
+        final int expectedOrdinal;
+        EnumWithProtoMetadata enumWithProtoMetadata = mockEnum(expectedOrdinal = RNG.nextInt());
         ProtoWriterTools.writeEnum(bufferedData, definition, enumWithProtoMetadata);
         bufferedData.flip();
         assertVarIntTag(definition);
         assertEquals(expectedOrdinal, bufferedData.readVarInt(false));
+    }
+
+    @Test
+    void testWriteEnumZeroOrdinal() {
+        FieldDefinition definition = createFieldDefinition(ENUM);
+        EnumWithProtoMetadata enumWithProtoMetadata = mockEnum(0);
+        final long positionBefore = bufferedData.position();
+        ProtoWriterTools.writeEnum(bufferedData, definition, enumWithProtoMetadata);
+        final long positionAfter = bufferedData.position();
+        assertEquals(positionAfter, positionBefore);
+    }
+
+    @Test
+    void testWriteString_empty() throws IOException {
+        FieldDefinition definition = createFieldDefinition(STRING);
+        String valToWrite = "";
+        final long positionBefore = bufferedData.position();
+        writeString(bufferedData, definition, valToWrite);
+        final long positionAfter = bufferedData.position();
+        assertEquals(positionAfter, positionBefore);
     }
 
     @Test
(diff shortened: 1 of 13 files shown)


pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
```diff
@@ -142,6 +152,16 @@ void testWriteTagSpecialWireType() {
...
+        final int valToWrite = 0;
+        final long positionBefore = bufferedData.position();
+        writeInteger(bufferedData, definition, valToWrite);
+        final long positionAfter = bufferedData.position();
+        assertEquals(positionBefore, positionAfter);
+    }
+
     @Test
     void testWriteInteger_int32() {
         FieldDefinition definition = createFieldDefinition(INT32);
```
artemananiev, line 167: This test case writes a random int. If the int happens to be 0 (the default int value), the test would fail. Does it make sense to protect from this as a part of current PR?

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
```diff
@@ -192,6 +212,16 @@ void testWriteInteger_unsupported(FieldType type) {
         assertThrows(RuntimeException.class, () -> writeInteger(bufferedData, definition, RNG.nextInt()));
     }
 
+    @Test
+    void testWriteLong_Zero() {
```
artemananiev, line 216: s/Zero/zero/ , or better use something like "default", cause the test case actually verifies how default values are (not) written.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
```diff
@@ -280,19 +310,44 @@ void testWriteBoolean_false() {
...
+
     @Test
     void testWriteEnum() {
         FieldDefinition definition = createFieldDefinition(ENUM);
-        EnumWithProtoMetadata enumWithProtoMetadata = mock(EnumWithProtoMetadata.class);
-        int expectedOrdinal = RNG.nextInt();
-        when(enumWithProtoMetadata.protoOrdinal()).thenReturn(expectedOrdinal);
-        when(enumWithProtoMetadata.protoName()).thenReturn(RANDOM_STRING.nextString());
+        final int expectedOrdinal;
+        EnumWithProtoMetadata enumWithProtoMetadata = mockEnum(expectedOrdinal = RNG.nextInt());
```
artemananiev, line 324: I wonder why `expectedOrdinal` is initialized here rather than on the previous line?

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
```diff
@@ -986,6 +1059,268 @@ void testSizeOfDelimited() {
...
+                0xffff,
+                0xfffff,
+                0xffffff,
+                0xfffffff,
+                0xffffffff
+        ));
+        final long finish = bufferedData.position();
+
+        int tag = bufferedData.getVarInt(start, false);
+        assertEquals(WIRE_TYPE_DELIMITED.ordinal(), tag & 0x3);
```
artemananiev, line 1091: `ProtoParserTools` has a constant for this 0x03 value, it can be used here. Also, the constant seems to be useful not only while parsing, but also on writing, so it can be moved to `ProtoConstants`, maybe?

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
```diff
@@ -986,6 +1059,268 @@ void testSizeOfDelimited() {
...
+        int sizeOfTag = ProtoWriterTools.sizeOfVarInt32(tag);
```
artemananiev, line 1092: Can `ProtoWriterTools.sizeOfTag()` be used here?

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
```diff
@@ -986,6 +1059,268 @@ void testSizeOfDelimited() {
...
+        assertEquals(expectedSize, size);
+        int sizeOfSize = ProtoWriterTools.sizeOfVarInt32(size);
+
+        assertEquals(finish - start - sizeOfTag - sizeOfSize, size);
+
+        long firstLong = bufferedData.getVarLong(start + sizeOfTag + sizeOfSize, zigZag);
+        assertEquals(0x0f, firstLong);
+    }
+
+    private static interface WriterMethod<T> {
```
artemananiev, line 1154: I'm not sure, but can `ProtoWriter` interface be reused here?

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
```diff
@@ -986,6 +1059,268 @@ void testSizeOfDelimited() {
...
+            assertEquals((definition.number() << 3) | WIRE_TYPE_DELIMITED.ordinal(), tag);
+            int sizeOfTag = ProtoWriterTools.sizeOfVarInt32(tag);
+
+            UnpackedField<T> value = readerMethod.read(bufferedData, start + offset + sizeOfTag);
+            assertEquals(list.get(i), value.value());
+
+            offset += sizeOfTag + value.size();
+        }
+
+        assertEquals(finish,start + offset);
```
artemananiev, line 1322: Whitespace

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialDataTest.java
```diff
@@ -11,6 +20,38 @@ protected ReadableSequentialData emptySequence() {
         return new StubbedSequence(new byte[0]);
     }
 
+    @NonNull
+    private ReadableSequentialData throwingSequence() {
```
artemananiev, line 24: This would test `StubbedSequence` only. Instead, it would be great to have this method as abstract in `ReadableSequentialTestBase` and implement it wherever needed. It would test all ReadableSequentialData implementations.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialDataTest.java
```diff
@@ -11,6 +20,38 @@ protected ReadableSequentialData emptySequence() {
...
+        return new StubbedSequence(new byte[] { 1 }, () -> new EOFException());
+    }
+
+    @Test
```
artemananiev, line 28: These new test cases can be moved to `ReadableSequentialTestBase`, too

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableTestBase.java
```diff
@@ -1293,12 +1334,30 @@ void readZigZag() {
...
+                    (byte) 0b10000010,
+                    (byte) 0b10000010,
+                    (byte) 0b10000010,
+                    (byte) 0b10000010,
+                    (byte) 0b10000010,
+                    (byte) 0b10000010
+            });
+            // When we try to decode an int, the lack of a "terminator" bit causes a DataEncodingException
+            assertThatThrownBy(() -> seq.readVarInt(false)).isInstanceOf(DataEncodingException.class);
+        }
```
artemananiev, line 1355: Is there a test that calls `readVarInt()` for a sequence of bytes, where more than 5 bytes have continuation bit set, and the last one does not? `readVarLong()` should handle such a sequence just fine, but (to my surprise), `readVarInt()` should not fail either. It should read it as a long and then downcast to int, if I remember correctly.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/buffer/BufferedDataTestBase.java
```diff
@@ -106,6 +106,22 @@ void sliceThenReadVarInt(final int num) {
         assertEquals(afterSecondIntPos - afterFirstIntPos, slicedBuf.position());
     }
 
+    @ParameterizedTest
+    @ValueSource(ints = {0, 1, 2, 4, 7, 8, 15, 16, 31, 32, 33, 127, 128, 512, 1000, 1024, 4000,
+            16384, 65535, 65536, 65537, 0xFFFFFF, 0x1000000, 0x1000001, 0x7FFFFFFF,
+            -1, -7, -8, -9, -127, -128, -129, -65535, -65536, -0xFFFFFF, -0x1000000, -0x1000001, -0x80000000})
+    @DisplayName("readVar() won't read beyond 10 bytes")
+    void readVarFromLargeBuffer(final int num) {
+        final var buf = allocate(100);
```
artemananiev, line 115: Var ints (including longs) may not take more than 10 bytes

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
anthony-swirldslabs: Good catch. Addressed it for this and a few more subsequent tests.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
anthony-swirldslabs: Changed to `zero`.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
anthony-swirldslabs: No particular reason. Moved the assignment to the declaration.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
anthony-swirldslabs: That's a very good catch. I think it should actually be 0x7 because there's 3 bits in the wire type. I'll use a common constant for that.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
anthony-swirldslabs: Yes, in theory. However, this particular test verifies the bytes literally written to the buffer by a particular method `ProtoWriterTools.writeIntegerList`. So I want to check the raw written bytes only and not rely on other high(er)-level logic implemented in ProtoWriterTools.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
anthony-swirldslabs: The ProtoWriter seems to accept an output sink and a single data variable, while this interface would be unable to infer a particular field definition from the given data object alone. So I need this custom interface here that also accepts the field definition.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoWriterToolsTest.java
anthony-swirldslabs: Fixed.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialDataTest.java
anthony-swirldslabs: Yes, that was my original idea, too. But I found it unfeasible to support this test in one of the descendants of the TestBase - I recall I just couldn't reach the particular branch that would throw the exception from the test that uses this sequence. That's the reason why it's sitting here.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableSequentialDataTest.java
anthony-swirldslabs: Same reason as above: I didn't find an easy way to reach a branch that would trigger the exception in one of the descendent classes.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/ReadableTestBase.java
anthony-swirldslabs: Yes, readVarInt() basically delegates to readVarLong(). So there's no way to meaningfully test this until/unless we actually separate them and start failing the readVarInt on longer sequences. But I guess this would introduce some incompatibilities with Google's implementation (IIRC, right?), that's why we simply delegate for now.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/buffer/BufferedDataTestBase.java
anthony-swirldslabs: Yes, yes, but as the name of the test suggests it specifically wants to use a longer buffer to check an edge case somewhere in the implementation. 100 is long enough for me.

That edge case may have existed in the old readVarInt implementation and perhaps you removed it when you overhauled its implementation a few weeks ago. However, the test seems good still - it still covers the edge case and/or will cover it if/when we re-introduce it in the future.

imalygin (approved):

PR #167, opened 2024-01-11
fix 104: Parsing should never wrap

### LINKED ISSUE #104: Parsing should never wrap
### Description

When parsing data the input buffer can not be assumed to be owned by parser so wrapped bytes can never pass though parsing.

### Steps to reproduce

Parse a protobuf object with field of type bytes from a byte[] backed BufferedData. Then change the input byte[] after and the byte value in parsed object will change.

### Additional context

_No response_

### Hedera network

other

### Version

0

### Operating system

None

anthony-swirldslabs:
**Description**:
Problem statement: in PBJ, parsing a model from a data buffer should create copies of data, especially for byte arrays, rather than use a direct view to the original buffer. Note that strings are covered for us by JDK itself because it always creates a copy. So of all the primitive types that we support in PBJ, only byte arrays need to be taken care of.

In this fix, the Codec.parse(Bytes) replaces usages of `Bytes.toReadableSequentialData` with a new `Bytes.toCopyingReadableSequentialData`. The latter initializes a returned `RandomAccessSequenceAdapter` object with a special flag that makes its `Bytes readBytes(int length)` method create a copy of the returned bytes.

Also, a test is added that checks all the three non-test implementations of the ReadableSequentialData on whether the wrapping occurs when parsing objects from these Data objects. Note that it is made sure that the test covering Bytes calls the particular Codec method that calls the `toCopyingReadableSequentialData` method.

The proposed solution is the least intrusive from the performance perspective. Ideally, the Bytes implementation itself should ensure to only return a view of the underlying data when explicitly requested. But it turns out that pretty much the entire implementation is built on the assumption of returning the view of the underlying data for performance reasons. So fixing this in Bytes directly would probably introduce performance regressions in other places.

Note that if anyone calls `Codec.parse(bytes.toReadableSequentialData)` explicitly, then they would still run into the wrapping issue. We can make this method package-private or otherwise restrict access to it if we don't want this to happen. As of today, this method was only used in Codec, which is being replaced in this fix, and also there's a single usage in `ReadableSequentialData.view()` default implementation. However, this very implementation actually creates a new bytes array. So the default implementation doesn't really provide a view. In any case, it doesn't seem to need fixing.

**Update on the above paragraph:** I just discovered that hedera-services uses toReadableSequentialData() left and right when passing it to a codec, even though the codec now provides a convenience method for parsing Bytes directly. So I'll be removing those calls from hedera-services in a separate PR.

Also, a bug in Bytes.replicate() is fixed. Previously it returned an incorrect, shortened replica when it itself was a view on a portion of another array with a non-zero offset. A new test is written for that as well.

**Related issue(s)**:

Fixes #104

**Notes for reviewer**:
The new test at pbj-integration-tests/src/test/java/com/hedera/pbj/intergration/test/ParserNeverWrapsTest.java passes.

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
@@ -209,10 +209,9 @@ public final class Bytes implements RandomAccessData {
      */
     @NonNull
     public Bytes replicate() {
-        final var newLength = length - start;
-        final var bytes = new byte[newLength];
-        System.arraycopy(buffer, start, bytes, 0, newLength);
-        return new Bytes(bytes, 0, newLength);
+        final var bytes = new byte[length];
+        System.arraycopy(buffer, start, bytes, 0, length);
+        return new Bytes(bytes, 0, length);
     }
 
     /**
@@ -320,6 +319,17 @@ public final class Bytes implements RandomAccessData {
     @NonNull
     public ReadableSequentialData toReadableSequentialData() {
         return new RandomAccessSequenceAdapter(this);
+    }
+
+    /**
+     * Create and return a new {@link ReadableSequentialData} that is backed by this {@link Bytes}
+     * and that returns a replicated Bytes object upon a call to readBytes(int length).
+     *
+     * @return A {@link ReadableSequentialData} backed by this {@link Bytes}.
+     */
+    @NonNull
+    public ReadableSequentialData toCopyingReadableSequentialData() {
+        return new RandomAccessSequenceAdapter(this, true);
     }
 
     /**
--- a/pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/buffer/BytesTest.java
+++ b/pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/buffer/BytesTest.java
@@ -67,6 +67,20 @@ final class BytesTest {
         assertEquals(2, replicatedBytes.length());
         assertEquals(arr[1], replicatedBytes.getByte(0));
         assertEquals(arr[2], replicatedBytes.getByte(1));
+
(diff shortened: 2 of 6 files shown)


pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/buffer/BytesTest.java
```diff
@@ -56,6 +56,19 @@ void toStringWorks1() {
...
+        byte[] arr = new byte[] { 0, 1, 2 };
+
+        // Only wrap the last two elements:
+        Bytes bytes = Bytes.wrap(arr, 1, 2);
+        Bytes replicatedBytes = bytes.replicate();
+
+        assertEquals(2, replicatedBytes.length());
+        assertEquals(arr[1], replicatedBytes.getByte(0));
+        assertEquals(arr[2], replicatedBytes.getByte(1));
+    }
```
artemananiev, line 70: Should it also test that changing bytes in the original array doesn't affect the replicated object?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -209,10 +209,9 @@ public long getLong(final long offset, @NonNull final ByteOrder byteOrder) {
      */
     @NonNull
     public Bytes replicate() {
```
artemananiev, line 211: I wonder how different this method is from `Bytes getBytes(offset, length)`?

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/buffer/BytesTest.java
anthony-swirldslabs: Good idea. Just pushed a commit with this improvement.

imalygin (approved):

jsync-swirlds (approved):

PR #168, opened 2024-01-12
fix 106: bytes should never be null

### LINKED ISSUE #106: Bytes should never be null on PBJ generated classes
### Problem

Given a generated class that supports protobuf bytes, we represent this as a `Bytes` object. A lot of code is doing null checks when a default Bytes.EMPTY return from the pbj generated object would have made the code simpler.

### Solution

According to the [specification](https://protobuf.dev/programming-guides/proto3/), the default value for `bytes` is empty bytes. And in our parser we honor this, making sure the bytes are always Bytes.EMPTY or equivalent, and never null. But we do not mark the Bytes argument in the record as NonNull, so code that uses a PBJ object tries to be defensive when it shouldn't need to. If we annotated this properly, then static analysis could prove it is safe. And we should absolutely forbid null in the builder for Bytes arguments.

### Alternatives

_No response_

anthony-swirldslabs:
**Description**:
* Annotate `bytes` fields with `@NonNull`.
* Modify the builder to also use the annotation for `bytes` fields, as well as substitute `null` values passed to the builder with their default values.
* Added a test to verify this behavior.

**Related issue(s)**:

Fixes #106 

**Notes for reviewer**:
All tests (including the new one) in all of pbj-runtime/compiler/integ tests pass.

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
@@ -40,6 +40,8 @@ import java.util.stream.Collectors;
 @SuppressWarnings({"StringConcatenationInLoop", "EscapedSpace"})
 public final class ModelGenerator implements Generator {
 
+	private static final String NON_NULL_ANNOTATION = "@NonNull";
+
 	private static final String HASH_CODE_MANIPULATION =
 		"""
 		// Shifts: 30, 27, 16, 20, 5, 18, 10, 24, 30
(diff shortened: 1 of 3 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -223,12 +226,24 @@ public record $javaRecordName(
...
 
+	/**
+	 * Returns a set of annotations for a given field.
+	 * @param field a field
+	 * @return an empty string, or a string with Java annotations ending with a space
+	 */
+	private static String getFieldAnnotations(final Field field) {
+		return switch (field.type()) {
+			case MESSAGE -> "@Nullable ";
+			case BYTES -> NON_NULL_ANNOTATION + " ";
```
artemananiev, line 243: Why is there a constant for NonNull, but there is no for Nullable?

artemananiev (approved):

jsync-swirlds (approved):

Issue #76, opened 2023-07-28
Need API to write repeated bytes fields one message by one


artemananiev:
As far as I can see, current assumptions in `ProtoWriterTools` regarding writing bytes are:

* If a field is repeated, use `writeBytesList()`, which accepts a list of bytes objects
* If a field is not repeated, use `writeBytes()`, which accepts just a single bytes object

Both methods have corresponding assertions.

This doesn't work very well for MerkleDb. In MerkleDb PBJ, data files contain a lot of similar data items (e.g. virtual hash records), so I decided to have a repeated bytes field for this purpose. However, these data items are written one by one, not all at a time, as there may be a lot of them. My first attempt was to use `writeBytes()` above, but it failed to work because of its assertion:

```java
        assert !field.repeated() : "Use writeBytesList with repeated types";
```

Both `writeBytes()` and `writeBytesList()` use `writeBytesNoChecks()` internally, which does exactly what's needed in MerkleDb. However, this method is not public. I had to re-implement it in MerkleDb code.

PR #169, opened 2024-01-13
fix: 76: Need API to write repeated bytes fields one message by one

artemananiev:
Fix summary:
* New methods are added to `ProtoWriterTools` to write repeated fields one by one (in contrast to `writeXYZList()`, which expects all repeated values are passed as a list): `writeOneRepeatedBytes()`, `writeOneRepeatedMessage()`, and `writeOneRepeatedString()`
* These methods are identical to the corresponding `writeXYZ()`, except the assertion about field repeated type is opposite
* New tests are provided for the new methods. They check that the new methods, if called multiple times, are identical to the corresponding `writeXYZList()`

Fixes: https://github.com/hashgraph/pbj/issues/76

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
@@ -288,10 +288,7 @@ public final class ProtoWriterTools {
      */
     public static void writeBytes(final WritableSequentialData out, final FieldDefinition field,
             final RandomAccessData value) throws IOException {
-        assert field.type() == FieldType.BYTES ||
-                field.type() == FieldType.STRING ||
-                field.type() == FieldType.MESSAGE
-                : "Not a byte[] type " + field;
+        assert field.type() == FieldType.BYTES : "Not a byte[] type " + field;
         assert !field.repeated() : "Use writeBytesList with repeated types";
         writeBytesNoChecks(out, field, value, true);
     }
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
@@ -227,28 +227,47 @@ public final class ProtoWriterTools {
     }
 
     /**
-     * Write a string to data output
+     * Write a string to data output, assuming the field is non-repeated.
      *
      * @param out The data output to write to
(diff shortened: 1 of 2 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -259,29 +278,53 @@ private static void writeStringNoChecks(WritableSequentialData out, FieldDefinit
...
      * @throws IOException If a I/O error occurs
      */
-    public static void writeBytes(WritableSequentialData out, FieldDefinition field, RandomAccessData value) throws IOException {
-        assert field.type() == FieldType.BYTES : "Not a byte[] type " + field;
+    public static void writeBytes(final WritableSequentialData out, final FieldDefinition field,
+            final RandomAccessData value) throws IOException {
+        assert field.type() == FieldType.BYTES ||
+                field.type() == FieldType.STRING ||
+                field.type() == FieldType.MESSAGE
+                : "Not a byte[] type " + field;
```
artemananiev, line 294: I found one of `ProtoWriterTools` test cases calls this method (indirectly) for a string field, providing string bytes as a wrapped Bytes object. If assertions were enabled, this test case would fail.

Rather than removing or rewriting the test case, I decided to loose the assertion here. Why not allowing this method to be called for any delimited (bytes, string, message) field?

anthony-swirldslabs: An arbitrary bytes arrays may not necessarily represent a valid UTF-8 string. This method can successfully write the bytes array, but the deserialization operation may fail for the written bytes then. Same deal with messages.

I'd suggest to keep the assertion as it is and try and modify the test (and the code the test was written for) to avoid this risky code-path. If the code-path is still useful, then a new method should be introduced (similar to "NoChecks") that would bypass the validation. This would make the code that uses this code-path aware of what it's doing rather then implicitly assuming that everything should work, which it may not as explained above.

anthony-swirldslabs (dismissed):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
artemananiev: OK, I reverted the assertion and fixed the test instead

imalygin (approved):

anthony-swirldslabs (approved):

Issue #175, opened 2024-01-19
CompareTo Generation fails if fields include underscores


jsync-swirlds:
### Description

If a field in a proto message contains underscores in the name, the new CompareTo generation fails with two messages:
 * `Exception while processing file: /Users/user/Projects/Hashgraph/Pbj/pbj-integration-tests/src/main/proto/comparable.proto`
 * `Cannot invoke "com.hedera.pbj.compiler.impl.Field.optionalValueType()" because "f" is null`.  

Note that no further information is given, so it is a particularly difficult error to track down.

### Steps to reproduce

1. Modify pbj-integration-tests/src/main/proto/comparable.proto
   * change `bytes bytes` to `bytes bytes_field` on line 26
   * change `, bytes"`, to `, bytes_field"` on line 18
1. run `./gradlew clean assemble -Dsonar.gradle.skipCompile=true --warning-mode=summary --no-daemon --no-build-cache --no-configuration-cache`

The compilation will fail with the referenced error

jsync-swirlds:
Here is a patch that resolves this issue (and includes an abbreviated stack trace in compiler output for compiler exceptions).
[fix-175-CompareTo-Generation-fails-if-fields-include.patch](https://github.com/hashgraph/pbj/files/13992362/fix-175-CompareTo-Generation-fails-if-fields-include.patch)

poulok:
@jsync-swirlds is your branch ready for review? Do you need someone to take it over?

jsync-swirlds:
I sent a draft PR.  I have bandwidth to complete the PR if that's easiest.

PR #177, opened 2024-01-19
fix: 175 CompareTo Generation fails if fields include underscores

jsync-swirlds:
* Adjusted ModelGenerator to use field `name` instead of `nameCamelFirstLower` when filtering comparable fields.
* Added an abbreviated stack trace for exceptions when compiling proto files, to help debug future issues like this.
* Modified comparable.proto in integration tests, renaming `bytes bytes` to `bytes test_bytes` to prevent regression.
* Modified the spotless conventions to allow Java 21 to be used

Fixes: #175

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/PbjCompilerTask.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/PbjCompilerTask.java
@@ -35,6 +35,8 @@ import org.gradle.api.tasks.TaskAction;
 
 /** Gradle Task that generates java src code from protobuf proto schema files. */
 public abstract class PbjCompilerTask extends SourceTask {
+    private static final int MAX_TRACE_FRAMES = 8;
+    private static final String STACK_ELEMENT_INDENT = "    ";
 
     /**
      * Set the java main directory that we write generated code into
@@ -120,6 +122,13 @@ public abstract class PbjCompilerTask extends SourceTask {
                     }
                 } catch (Exception e) {
                     System.err.println("Exception while processing file: " + protoFile);
+                    // Print an abbreviated stack trace for help in debugging.
+                    System.err.println(e);
+                    var trace = e.getStackTrace();
+                    int count = 0;
+                    for (var element : trace) {
+                        if (count++ < MAX_TRACE_FRAMES) System.err.println(STACK_ELEMENT_INDENT + element);
+                    }
                     throw e;
                 }
             }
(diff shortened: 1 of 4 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/PbjCompilerTask.java
```diff
@@ -35,6 +35,8 @@
 
 /** Gradle Task that generates java src code from protobuf proto schema files. */
 public abstract class PbjCompilerTask extends SourceTask {
+    private final static int MAX_TRACE_FRAMES = 8;
+    private static final String STACK_ELEMENT_INDENT = "    ";
```
anthony-swirldslabs, line 39: Nit: please list type modifiers in a consistent order, e.g. always `static final`.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/PbjCompilerTask.java
```diff
@@ -120,6 +122,13 @@ public static void compileFilesIn(Iterable<File> sourceFiles,
                     }
                 } catch (Exception e) {
                     System.err.println("Exception while processing file: " + protoFile);
+                    // Print an abbreviated stack trace for help in debugging.
+                    System.err.println(e);
+                    var trace = e.getStackTrace();
+                    int count = 0;
+                    for (var element : trace) {
+                        if(count++ < MAX_TRACE_FRAMES) System.err.println(STACK_ELEMENT_INDENT + element);
```
anthony-swirldslabs, line 130: Minor: space after `if` please.

anthony-swirldslabs (approved):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/PbjCompilerTask.java
jsync-swirlds: We definitely need to get spotless on this repo...
done.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/PbjCompilerTask.java
jsync-swirlds: Missed that, I think the IDE mixed one of them.
Thanks, and done.

anthony-swirldslabs (approved):

PR #188, opened 2024-01-29
fix: Fix double-wrap of `ParseExecption`

jsync-swirlds:
* Added a test in the catch block to rethrow ParseException rather than wrapping it in another ParseException

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
@@ -173,6 +173,9 @@ class CodecParseMethodGenerator {
                         }
                         return new $modelClassName($fieldsList);
                     } catch (final Exception anyException) {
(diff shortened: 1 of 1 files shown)


anthony-swirldslabs (approved):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -173,6 +173,9 @@ static String generateParseInternalMethod(final String modelClassName, final Lis
                         }
                         return new $modelClassName($fieldsList);
                     } catch (final Exception anyException) {
+                        if (anyException instanceof ParseException parseException) {
```
netopyr, line 176: Interesting solution. Why didn't you add a second catch-block?

jsync-swirlds: ParseException isn't always thrown (depends on the "case statements" produced below), only when there is a child message, so the compiler would fail in many cases with the catch block approach.

PR #190, opened 2024-01-30
fix: 118: revisit exceptions

### LINKED ISSUE #118: Revisit data access exceptions in RandomAccessData, Bytes, BufferedData, and other classes
Here is one of the methods in `RandomAccessData` interface:

```
    /**
     * Gets the signed byte at the given {@code offset}.
     *
     * @param offset The offset into data to get a byte from.
     * @return The signed byte at given {@code offset}
     * @throws BufferUnderflowException if the buffer is empty
     * @throws IndexOutOfBoundsException If the given {@code offset} is negative or not less than {@link #length()}
     */
    byte getByte(final long offset);
```

Method semantics is "get a byte at the given offset", which is identical to byte arrays. IOOBE is consistent with what exceptions are used by arrays, but BufferUnderflowException looks alien here. These BufferXYZExceptions are used by buffers on sequential access, when there is no more data to read / write. On random access, java.nio buffer classes throw IOOBE, too.

This ticket is to iterate over PBJ classes and interfaces and make sure exceptions used to indicate illegal access are consistent across code.

anthony-swirldslabs:
**Description**:
1. Revisiting exceptions thrown by core classes in PBJ.
2. Removing DataAccessException and replacing it with direct usages of the UncheckedIOException.

**Related issue(s)**:

Fixes #118 

**Notes for reviewer**:
`gr test` passes in all of pbj-runtime, pbj-compiler, and pbj-integration-tests.

**Checklist**


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ReadableSequentialData.java
```diff
@@ -28,7 +29,7 @@ public interface ReadableSequentialData extends SequentialData {
      *
      * @return The signed byte at the current {@link #position()}
      * @throws BufferUnderflowException If there are no bytes remaining in this sequence
-     * @throws DataAccessException If an I/O error occurs
+     * @throws UncheckedIOException If an I/O error occurs
```
jsync-swirlds, line 32: It doesn't change a lot, but if you're wanting to document this, you can add the exeception in an actual `throws` clause on the method as well.  Doing that can help callers be aware of the exception without reading Javadoc (which many people don't look at very often).

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
```diff
@@ -546,4 +551,61 @@ default boolean contains(final long offset, @NonNull final RandomAccessData data
         }
         return true;
     }
+
+    /**
+     * Throws {@code IndexOutOfBoundsException} if the given {@code offset} is negative
+     * or greater than/equal to the given {@code length}.
+     *
+     * @param offset offset
```
jsync-swirlds, line 559: Would be helpful to fill in the parameter documentation on these methods.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
```diff
@@ -546,4 +551,61 @@ default boolean contains(final long offset, @NonNull final RandomAccessData data
...
+        if (offset < 0 || offset >= length) {
+            throw new IndexOutOfBoundsException("offset " + offset + " is out of bounds for length " + length);
+        }
+    }
+
+    /**
+     * Throws {@code IndexOutOfBoundsException} if the given {@code offset} is negative
+     * or greater than/equal to the given {@code length}, and throws {@code BufferUnderflowException}
+     * if the {@code offset + dataLength - 1} exceeds the given {@code length}.
+     *
```
jsync-swirlds, line 572: Might want a `<p>` here so the blank line remains in the rendered output.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessSequenceAdapter.java
```diff
@@ -138,6 +138,9 @@ public long readBytes(@NonNull final ByteBuffer dst) {
         final var length = Math.min(dst.remaining(), remaining());
         final var finalLimit = dstPos + Math.min(length, dst.remaining());
         dst.limit(Math.toIntExact(finalLimit));
+        if (length == 0) {
```
jsync-swirlds, line 141: nit: possibly move this test before the costly math operations.

anthony-swirldslabs: Nope, that's the catch. I originally did just that and discovered a lot of tests fail. The tests rely on a side-effect of all these operations that actually call dst.limit() at line 140, so we cannot skip the call and return earlier.

That's a terrible side-effect and an undocumented one. But that's what today's code relies on. So I decided not to break the existing code. We'll need a separate issue if we want to remove this side-effect. Sadly, it's present in a few places, not just this one.

jsync-swirlds: Ahh.  That is, indeed, a very unfortunate (and undesirable) condition.
Perhaps file an issue to eventually fix that side effect reliance(so it isn't forgotten)?

anthony-swirldslabs: I filed https://github.com/hashgraph/pbj/issues/194

jsync-swirlds (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ReadableSequentialData.java
anthony-swirldslabs: I wish this wasn't true. I wish people actually read the documentation.

I may not fully agree with this suggestion because at first glance one could think that these checked exceptions when they see a method signature with the `throws` clause. Also, we have to mention the exceptions at least twice now, which isn't nice. But I implemented this in the latest commit. Please have a look.

imalygin (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SequentialData.java
```diff
@@ -78,7 +80,7 @@ default long remaining() {
      *
      * @param count number of bytes to skip. If 0 or negative, then no bytes are skipped.
      * @return the actual number of bytes skipped.
-     * @throws DataAccessException if an I/O error occurs
+     * @throws UncheckedIOException if an I/O error occurs
      */
-    long skip(long count);
+    long skip(long count) throws UncheckedIOException;
```
artemananiev, line 85: Should this method also throw `BufferUnderflowException`? Skipping bytes falls into the same category as reading bytes, and methods to read bytes usually throw underflow exceptions in other classes.

BTW, JavaDoc for this method explicitly mentions buffer overflow/underflow exceptions. Hmm... This is interesting. This is a base interface, which is extended by `ReadableSequentialData` and `WritableSequentialData`, both have "skip" semantics, but overflows are interpreted differently (underflow vs overflow)

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/WritableSequentialData.java
```diff
@@ -384,9 +391,11 @@ default void writeDouble(final double value, @NonNull final ByteOrder byteOrder)
      *
      * @param value integer to write in var int format
      * @param zigZag use protobuf zigZag varint encoding, optimized for negative numbers
-     * @throws DataAccessException if an I/O error occurs
+     * @throws BufferOverflowException If there are fewer than eight bytes remaining before {@link #limit()}
```
artemananiev, line 394: Var ints (and longs) may take up to 10 bytes, not 8

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/ByteArrayBufferedData.java
```diff
@@ -84,10 +82,10 @@ public byte getByte(final long offset) {
     public long getBytes(final long offset, @NonNull final byte[] dst, final int dstOffset, final int maxLength) {
         validateLen(maxLength);
         final long len = Math.min(maxLength, length() - offset);
-        validateCanRead(offset, len);
         if (len == 0) {
             return 0;
         }
+        checkOffsetToRead(offset, length(), len);
```
artemananiev, line 88: Since the offset is now checked after length, calls like `getBytes(-1, dst, 0, 0)` will now return 0, but threw an exception previously. I wonder if there are any tests for this behavior.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/ByteArrayBufferedData.java
```diff
@@ -84,10 +82,10 @@ public byte getByte(final long offset) {
     public long getBytes(final long offset, @NonNull final byte[] dst, final int dstOffset, final int maxLength) {
         validateLen(maxLength);
         final long len = Math.min(maxLength, length() - offset);
```
artemananiev, line 84: `offset` is not validated at this point yet, it can be (for example) negative. Now assume the following call: `getBytes(-1, dst, 0, length + 1)`. It would initialize `len` to `length() + 1`, which doesn't look good

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
```diff
@@ -81,8 +78,9 @@ default long getBytes(final long offset, @NonNull final byte[] dst) {
      * @param maxLength The maximum number of bytes to be written to the given {@code dst} array; must be non-negative
      *                and no larger than {@code dst.length - offset}
      * @throws NullPointerException if {@code dst} is null
-     * @throws IndexOutOfBoundsException If {@code offset} is out of bounds of {@code dst} or if
-     *                                  {@code offset + maxLength} is not less than {@code dst.length}
+     * @throws IndexOutOfBoundsException If {@code dstOffset} is out of bounds of {@code dst},
+     *             or if {@code dstOffset + maxLength} is not less than {@code dst.length},
```
artemananiev, line 82: "less" or "less or equal"?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
```diff
@@ -567,4 +570,67 @@ default boolean contains(final long offset, @NonNull final RandomAccessData data
...
+     * @param length the maximum offset plus one (1) that is allowed to be accessed.
+     *               It may be equal to the total length of the underlying buffer,
+     *               or be less than the total length when a limit value is used.
+     * @param dataLength the length of the data to read
+     */
+    default void checkOffsetToRead(final long offset, final long length, final long dataLength) {
+        if (offset < 0 || offset > length || (offset == length && dataLength != 0)) {
+            throw new IndexOutOfBoundsException("offset " + offset + " is out of bounds for length " + length);
+        }
+        if (offset + dataLength - 1 >= length) {
```
artemananiev, line 608: `offset + dataLength - 1 >= length`
==
`offset + dataLength > length`
==
`offset > length - dataLength`

The latter form is preferable to avoid overflows.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SequentialData.java
anthony-swirldslabs: That's correct. However, depending on whether the sequence is readable or writable, either under- or over-flow exception would be thrown. It would be a bit confusing to officially list both of them in the throws clause at once. I think that the existing remark in the javadoc is good enough in this tricky case.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/ByteArrayBufferedData.java
anthony-swirldslabs: Added a separate check.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/ByteArrayBufferedData.java
anthony-swirldslabs: I've moved this above the `if`, so now it will throw.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
anthony-swirldslabs: Good catch. I rephrased it.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
anthony-swirldslabs: Thanks for the tip. I've updated it.

jsync-swirlds (approved):

PR #195, opened 2024-02-02
fix: 156: optimize OneOf.hashCode()

### LINKED ISSUE #156: OneOf.hashCode() implicitly allocates arrays
### Problem

The implementation of `OneOf.hashCode()` is straightforward:
```
        return Objects.hash(kind, value);
```
One side effect, however, is an implicit allocation of an array for the parameters. That makes this method the biggest heap consumer during transaction handling.

### Solution

Combine component `hashCode()` values explicitly, f.e.:
```
        return kind.hashCode * 31 + value.hashCode();
```

### Alternatives

_No response_

anthony-swirldslabs:
**Description**:
Objects.hash(Object...) implicitly creates an array to hold all the arguments passed in, which has a high memory foot-print.

I'm replacing this call with a manual computation of the hash code. ~~Note that the value computed by the Objects.hash() is slightly different from the explicit computation (probably because it uses a seed value, while this fix implicitly uses a seed of 0), so I had to update the unit test as well. However,~~ I think it's a good idea to keep the unit test in sync with the actual code, to account for any changes that JDK may implement in Objects.hash() in the future.

**Related issue(s)**:

Fixes #156 

**Notes for reviewer**:
`gr test` passes in all the three PBJ repos.

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
@@ -47,7 +47,7 @@ public record OneOf<E extends Enum<E>>(E kind, Object value) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(kind, value);
+        return kind.hashCode() * 31 + Objects.hashCode(value);
     }
 
 }
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
@@ -47,7 +47,7 @@ public record OneOf<E extends Enum<E>>(E kind, Object value) {
 
     @Override
     public int hashCode() {
-        return kind.hashCode() * 31 + Objects.hashCode(value);
+        return (31 + kind.hashCode()) * 31 + (value == null ? 0 : value.hashCode());
     }
 
 }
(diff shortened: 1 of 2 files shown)


imalygin (dismissed):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
```diff
@@ -47,7 +47,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(kind, value);
+        return kind.hashCode() * 31 + Objects.hashCode(value);
```
OlegMazurov, line 50: Why not essentially copy `Arrays.hashCode()` and preserve the value?
```
return (31 + kind.hashCode()) * 31 + (value == null ? 0 : value.hashCode());
```

anthony-swirldslabs: Not sure why this might be important, but okay, let's preserve the value.

imalygin (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/OneOf.java
anthony-swirldslabs: BTW, I confirm that I've verified that the new value matches the old implementation. However, I'm still rewriting the unit test so that it tests our implementation rather than assumes that it matches Objects.hash().

OlegMazurov: `MerkleDb` uses `key.hashCode()` internally. It it ever changes we have to do data migration.

OlegMazurov (approved):

Issue #209, opened 2024-02-29
Extended Validation in Schemas


jasperpotts:
### Problem

We would like the ability to provide extended validation rules for fields in the schema and then validate them at parse time in PBJ. The primary reason for this is security to limit the effects of badly written binary protobuf to cause the reader to spend more resources than absolutely necessary.  The secondary reason is improved rigger in schemas so they are more precise and there is less scope for mistakes. The rules covered should include:

- Min/Max sizes for Strings, Bytes and Sub-Message fields
- Min/Max for repeated fields
- Number ranges for numeric types, accepted options.
- Max recessive message depth
- Message max size in bytes

The rules we choose should be very fast to check at parse time and not have a noticeable performance impact on parsing. 

There is also the aim that this can be used for tightening the specification documentation for protobuf APIs. So ideally rules documentation will be supported for explaining these rules and what they exist etc.

### Solution

We could define our own or choose to be compatible with an existing option. For example "protovalidate" library from Buf.
https://github.com/bufbuild/protovalidate
The full options are way too much with full Google Common Expression Language support but maybe the Standard Constraints subset is applicable.
https://github.com/bufbuild/protovalidate/blob/main/docs/standard-constraints.md

There is also a older version called "protoc-gen-validate (PGV)" https://github.com/bufbuild/protoc-gen-validate that seems more widely used and might be a better fit.

History blog on V1 and V2 https://buf.build/blog/protoc-gen-validate-v1-and-v2

david-bakin-sl:
We have, in the HAPI protobufs, a `TokenTransferList` message which has two repeated fields - one for fungible transfers and one for non-fungible transfers. And there's a rule: "Each `TokenTransferList` can specify up to 10 adjustments."  But that would be a sum over the size of the two repeated fields.  Maybe that could be a supported constraint.

PR #210, opened 2024-02-29
fix: 208: limit max size

### LINKED ISSUE #208: Limit max Bytes size
Introduce a max size for a Bytes object to be parsed from binary form. Ideally, we want to configure this value in the model definition, but that can be implemented as a feature at a later time.

For now, we want to hard-code a maximum supported number in the parser and fail the parsing if we encounter a larger Bytes object in an input stream.

@jasperpotts to provide the number.

anthony-swirldslabs:
**Description**:
This fix introduces a default limit of 21K on repeated/length-encoded field sizes. When the parser encounters a field with a greater size, it errors out with ParseException.

This is a quick solution for release 48.

In a future release, we'll extend the functionality as described at https://github.com/hashgraph/pbj/issues/209.

**Related issue(s)**:

Fixes #208 

**Notes for reviewer**:
All unit tests and all integration tests, including the fuzzTest, pass in all three PBJ sub-projects.

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
@@ -211,6 +211,7 @@ public final class ProtoParserTools {
      * Read a String field from data input
      *
      * @param input the input to read from
+     * @param maxSize the maximum allowed size
      * @return Read string
      * @throws ParseException if the length is greater than maxSize
      */
@@ -284,7 +285,7 @@ public final class ProtoParserTools {
 
     /**
      * Skip over the bytes in a stream for a given wire type. Assumes you have already read tag.
-     * 
+     *
      * @param input The input to move ahead
      * @param wireType The wire type of field to skip
      * @throws IOException For unsupported wire types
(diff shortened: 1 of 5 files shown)


imalygin (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -200,7 +200,27 @@ public static double readDouble(final ReadableSequentialData input) {
      * @return Read string
      */
     public static String readString(final ReadableSequentialData input) throws IOException {
+        try {
+            return readString(input, Long.MAX_VALUE);
```
jasperpotts, line 204: Don't love the calling here with Long.MAX_VALUE to do the equivalent of disabling check as we still pay performance cost.

jasperpotts (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
anthony-swirldslabs: Yes, I was thinking the same when making this change. I decided to go with reusing the code instead of making a copy for a simple reason: in an ideal world, nobody should be calling this particular method outside of PBJ, and PBJ Codecs now only call the maxSize-version of this method now.

So as far as parsing with PBJ codecs is concerned, the performance should be top-notch. And those unlucky who decided to call ProtoParserTools methods directly from elsewhere should reconsider their choices. :)

Issue #211, opened 2024-03-01
(EXISTENTIAL) Reference fields generated with `@NonNull` are assumed non-null but not required to be


tinker-michaelj:
### Description

Consider the following test:
```
    @Test
    void gameOver() {
        final var protoBytes = new ProtoBytes(null);
        ProtoBytes.PROTOBUF.toBytes(protoBytes);
    }
```
The _second_ line throws an NPE:
```
java.lang.NullPointerException: Cannot invoke "com.hedera.pbj.runtime.io.buffer.RandomAccessData.length()" because "value" is null
	at com.hedera.pbj.runtime.ProtoWriterTools.writeBytesNoChecks(ProtoWriterTools.java:327)
	at com.hedera.pbj.runtime.ProtoWriterTools.writeBytes(ProtoWriterTools.java:294)
	at com.hedera.hapi.node.state.primitives.codec.ProtoBytesProtoCodec.write(ProtoBytesProtoCodec.java:71)
```

As PBJ is used in Services, this means:
 - A transaction handled at consensus could silently put a `ProtoBytes` with null value into a mutable state; which would,
 - Asynchronously kill the state-hashing threads after the state is fast-copied; and,
 - Shut down the entire network.

### Steps to reproduce

As above.

### Hedera network

other

### Version

`develop`

### Operating system

None

jasperpotts:
We have two options here:

anthony-swirldslabs:
Converting null to a default value would require adding an explicit constructor to the generated record, which seems to add a lot of boilerplate code. While our Builder implementation substitute null Bytes with Bytes.EMPTY, the Protobuf Java spec says "Note that no Java protocol buffer methods accept or return nulls unless otherwise specified": https://protobuf.dev/reference/java/java-generated/ . And null is only mentioned in the context of enums, and only as a return value.

So it seems reasonable to update the record validator constructor to throw an NPE when the supplied Bytes is null.

Protobuf also defaults strings to empty strings, so it seems reasonable to treat them in a way similar to how we treat Bytes.

PR #224, opened 2024-03-07
fix: 220: Re-implement null checks for Bytes and Strings

### LINKED ISSUE #220: Re-implement null checks for Bytes and Strings
`null` was disallowed for Bytes and String at https://github.com/hashgraph/pbj/pull/214

However, services code seems to assign `null` Bytes left and right, e.g.:

https://github.com/hashgraph/hedera-services/blob/c92de0026ec8079d45e480b0eb3b450fbd06f198/hedera-node/hedera-token-service/src/main/java/com/hedera/node/app/service/token/AliasUtils.java#L75-L85

And then the `null` return value fails in a PBJ model validator.

We need to re-implement that fix and instead inject a full-fledged constructor into models that would substitute `null` with default values for Bytes and Strings.

anthony-swirldslabs:
**Description**:
Re-implementing the original fix for https://github.com/hashgraph/pbj/issues/211 by allowing null values for Bytes and Strings as inputs to model constructor/Builder, and automatically substituting the nulls with default values (Bytes.EMPTY or "").

This is necessary because code in hedera-services constructs objects and passes null values for such fields in a few places, so we have to support the null values as inputs. However, the fix makes sure that internally, we don't allow null values for private fields of these types in the model records.

Note that the Builder constructor does allow null OneOf fields as inputs. This is by design. This prevents any potential failures if anyone passes such nulls today because they were always allowed. This also implies that the code would use convenience methods exposed by the Builder for various OneOf variations. And finally, during a call to `build()`, an uninitialized OneOf field will still be caught by the model constructor. So it's fine to allow null as input for OneOf fields in the Builder constructor.

**Related issue(s)**:

Fixes #220 

**Notes for reviewer**:
* All tests pass in PBJ.
* Also, new unit tests are added.
* Also, hedera-services unit tests pass.
* Finally, while 4 HAPI tests from `gr hapiTestToken -Dfile.encoding=UTF-8 --scan` still fail, they fail with errors unrelated to this fix, and they fail all the same when running with PBJ w/o this fix. So as far as the original regression related to the 'null' values handling is concerned, it is now fixed with this PR, and the current failures are completely unrelated to this PBJ behavior.

**Checklist**


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -363,6 +361,58 @@ public int hashCode() {
...
+							);
+							break;
+						}
+						default:
+							sb.append("this.$name = $name;".replace("$name", field.nameCamelFirstLower()));
+							break;
+					}
+					return sb.toString();
+				}).collect(Collectors.joining("\n")).indent(DEFAULT_INDENT * 2))
+				;//.indent(DEFAULT_INDENT);
```
imalygin, line 413: please, clean it up.

imalygin (approved):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
anthony-swirldslabs: Good catch. I'll do this a follow-up PR when I'm re-applying the `maxDepth` fix later today.

PR #229, opened 2024-03-12
fix: stabilize hashCode() for Enums using proto ordinals

### LINKED ISSUE #227: Generated hashCode() for OneOf classes must not depend on Enum.hashCode()
All PBJ objects have auto-generated `hashCode()` implementation. For `OneOf` fields, hash codes are calculated like this:

```java
        return (31 + kind.hashCode()) * 31 + (value == null ? 0 : value.hashCode());
```

`kind` above is a enum, so its hash code is actually `Enum.hashCode()`. Here is the problem: this method is not deterministic, it returns different values across process runs:

```java
        int hc = hash;
        if (hc == 0) {
            hc = hash = System.identityHashCode(this);
        }
        return hc;
```

Since we use hash codes in long-term storage, they must be the same across node runs. It means `Enum.hashCode()` must not be used anywhere in our code. There may be other usages than in `OneOf`.

One quick option is to use `kind.ordinal()` (or some derivative) instead of `kind.hashCode()`, but there may be better solutions.

anthony-swirldslabs:
**Description**:
Modifying the original implementation at https://github.com/hashgraph/pbj/pull/228 by switching from using `Enum.name().hashCode()` to `Integer.hashCode(EnumWithProtoMetadata.protoOrdinal())`.

**Related issue(s)**:

Fixes #227 

**Notes for reviewer**:
* All tests in all three PBJ sub-projects pass.

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ComparableOneOf.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ComparableOneOf.java
@@ -44,8 +44,8 @@ public record ComparableOneOf<E extends Enum<E>>(E kind, Comparable value) imple
 
     @Override
     public int hashCode() {
-        // name().hashCode() and NOT Enum.hashCode() because the latter changes between runs
-        return (31 + kind.name().hashCode()) * 31 + (value == null ? 0 : value.hashCode());
+        return (31 + Integer.hashCode(((EnumWithProtoMetadata)kind).protoOrdinal())) * 31
+                + (value == null ? 0 : value.hashCode());
     }
 
     @SuppressWarnings("unchecked")
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ComparableOneOf.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ComparableOneOf.java
@@ -22,6 +22,7 @@ public record ComparableOneOf<E extends Enum<E>>(E kind, Comparable value) imple
         if (kind == null) {
             throw new NullPointerException("An enum 'kind' must be supplied");
         }
+        assert kind instanceof EnumWithProtoMetadata : "OneOf 'kind' must implement EnumWithProtoMetadata";
     }
 
     /**
--- a/pbj-core/pbj-runtime/src/test/java/tests/ComparableOneOfTest.java
+++ b/pbj-core/pbj-runtime/src/test/java/tests/ComparableOneOfTest.java
@@ -1,6 +1,7 @@
 package tests;
 
 import com.hedera.pbj.runtime.ComparableOneOf;
+import com.hedera.pbj.runtime.EnumWithProtoMetadata;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -28,7 +29,8 @@ class ComparableOneOfTest {
     @Test
     void hashCodeReturnsHashCode() {
         final var oneOf = new ComparableOneOf<>(TestEnum.KIND1, "Value");
-        assertEquals((31 + TestEnum.KIND1.name().hashCode()) * 31 + "Value".hashCode(), oneOf.hashCode());
+        assertEquals((31 + Integer.hashCode(TestEnum.KIND1.protoOrdinal())) * 31
+                + "Value".hashCode(), oneOf.hashCode());
     }
 
     @Test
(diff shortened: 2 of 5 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ComparableOneOf.java
```diff
@@ -22,6 +22,7 @@ public record ComparableOneOf<E extends Enum<E>>(E kind, Comparable value) imple
         if (kind == null) {
             throw new NullPointerException("An enum 'kind' must be supplied");
         }
+        assert kind instanceof EnumWithProtoMetadata : "OneOf 'kind' must implement EnumWithProtoMetadata";
```
netopyr, line 25: Is an `assert` the right approach? Why not a regular check?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ComparableOneOf.java
```diff
@@ -44,8 +45,8 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        // name().hashCode() and NOT Enum.hashCode() because the latter changes between runs
-        return (31 + kind.name().hashCode()) * 31 + (value == null ? 0 : value.hashCode());
+        return (31 + Integer.hashCode(((EnumWithProtoMetadata)kind).protoOrdinal())) * 31
```
netopyr, line 48: Nit: I don't think adding 31 provides a lot of value. It basically just adds a constant to the result. But it does not hurt.

pbj-core/pbj-runtime/src/test/java/tests/ComparableOneOfTest.java
```diff
@@ -46,9 +48,19 @@ void equalsWorks() {
...
 
-    public enum TestEnum {
+    public enum TestEnum implements EnumWithProtoMetadata {
         KIND1,
-        KIND2
+        KIND2;
+
+        @Override
+        public int protoOrdinal() {
+            return ordinal();
```
netopyr, line 57: The implementation of `protoOrdinal()` should be more complex IMO, resulting in negative values and gaps.

pbj-core/pbj-runtime/src/test/java/tests/ComparableOneOfTest.java
```diff
@@ -28,7 +29,8 @@ void asReturnsValue() {
     @Test
     void hashCodeReturnsHashCode() {
```
netopyr, line 30: I suggest to implement this as a parameterized tests that checks all values of `TestEnum`.

netopyr (approved):
LGTM

Just some minor comments that can be fixed later.

jsync-swirlds (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ComparableOneOf.java
anthony-swirldslabs: Yes, it should suffice. The `hashCode()` implementation will error out with a class cast exception if this assertion is not satisfied. So as far as the correctness is concerned, the current implementation is solid.

Using an `assert` here helps improve the performance a little bit because we really don't expect to see anything but `EnumWithProtoMetadata` here. But like mentioned above, even if that happens, we'll error out later on then.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ComparableOneOf.java
anthony-swirldslabs: The `31` comes from an implementation of `Objects.hashCode([])` which we removed nad replaced with this code because it created a new array and hurt the performance. However, we wanted to keep the hash codes values intact, hence the `31`.

pbj-core/pbj-runtime/src/test/java/tests/ComparableOneOfTest.java
anthony-swirldslabs: This is a test implementation and we currently don't have any tests that would benefit from a more complex implementation here.

pbj-core/pbj-runtime/src/test/java/tests/ComparableOneOfTest.java
anthony-swirldslabs: Good idea. I filed https://github.com/hashgraph/pbj/issues/232 to implement this at a later time.

PR #239, opened 2024-04-04
feat: make repeated fields @NonNull and default to empty lists

### LINKED ISSUE #238: Default repeated values to empty lists
Per the Protobuf spec at https://protobuf.dev/programming-guides/proto3/#default , repeated values should default to empty lists.

The current PBJ implementation mostly enforces this in the parser and the Builder of generated models. However, List fields are still marked `@Nullable` and the model constructor still allows them to be set to null.

We should modify the generated constructor to default List fields to empty lists if a null is passed to the constructor.

anthony-swirldslabs:
**Description**:
PBJ already defaults repeated fields to empty lists in the parser and the builder. This fix makes model constructors do the same, and generally makes the repeated fields non-nullable.

**Related issue(s)**:

Fixes #238

**Notes for reviewer**:
* Unit test
* `gr test` at hedera-services in hedera-file-service-impl which uses the FeeSchedule which was the main testing model.

The generated FeeSchedule ctor:
```
    public FeeSchedule(List<TransactionFeeSchedule> transactionFeeSchedule, TimestampSeconds expiryTime) {
        this.transactionFeeSchedule = transactionFeeSchedule == null ? Collections.emptyList() : transactionFeeSchedule;
        this.expiryTime = expiryTime;
    }
```

The generated Builder convenience method:
```
        public Builder transactionFeeSchedule(TransactionFeeSchedule ... values) {
            this.transactionFeeSchedule = values == null ? Collections.emptyList() :  List.of(values) ;
            return this;
        }
```

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
@@ -769,7 +769,8 @@ public final class ModelGenerator implements Generator {
 		if (field.repeated()) {
 			// Need to re-define the prefix and postfix for repeated fields because they don't use `values` directly
 			// but wrap it in List.of(values) instead, so the simple definitions above don't work here.
-			final String repeatedPrefix, repeatedPostfix;
+			final String repeatedPrefix;
+			final String repeatedPostfix;
 			if (parentOneOfField != null) {
 				repeatedPrefix = prefix + " values == null ? " + getDefaultValue(field, msgDef, lookupHelper) + " : ";
 				repeatedPostfix = postfix;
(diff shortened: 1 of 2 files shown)


jasperpotts:
It would be really nice to have a comment on PRs that change generation with a code snippet of generated code. It is very hard to quickly read PR and know what its effect on generated code is.

jasperpotts:
What is the effect on hasXXX() methods? as they were checking null for lists they will never be false, that might break code

anthony-swirldslabs:
> It would be really nice to have a comment on PRs that change generation with a code snippet of generated code. It is very hard to quickly read PR and know what its effect on generated code is.

That's fair. I've added the affected code snippets in the Description of this PR at the top.

anthony-swirldslabs:
> What is the effect on hasXXX() methods? as they were checking null for lists they will never be false, that might break code

So, the hasXXX() is a part of a collection of "convenience" methods that the PBJ compiler adds for every field of type Message. Since a scalar Message field is nullable per the Protobuf specification, these convenience methods allow you to check if it's present, or throw, or run some code only if it's present, as can be seen at: https://github.com/hashgraph/pbj/blob/main/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java#L517-L560

However, as stated in the comment where I've added an additional condition `!repeated` for this code block, repeatable fields of type Message technically always have a value, so the convenience methods dealing with nulls aren't necessary. The client code could use standard methods that work with collections in order to differentiate between empty and non-empty repeated Message fields.

I don't believe these convenience methods are used anywhere though, at least certainly not for repeated Message fields - if there were usages, my hedera-services build would fail.

imalygin (approved):

PR #249, opened 2024-04-26
chore: add more support for signatures

lpetrovic05:
closes https://github.com/hashgraph/hedera-services/issues/13015

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
@@ -306,13 +306,51 @@ public final class Bytes implements RandomAccessData, Comparable<Bytes> {
     }
 
     /**
-     * A helper method for efficient copy of our data into a Signature without creating a defensive copy of the data.
-     * The implementation relies on a well-behaved Signature that doesn't modify the buffer data.
-     *
-     * @param signature the Signature to copy into
+     * Same as {@link #updateSignature(Signature, int, int)} with offset 0 and length equal to the length of this
+     * {@link Bytes} object.
      */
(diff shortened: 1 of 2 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -305,14 +305,52 @@ public void writeTo(@NonNull final MessageDigest digest, final int offset, final
...
+     * @param signature The Signature to update
+     * @param offset    The offset from the start of this {@link Bytes} object to get the bytes from
+     * @param length    The number of bytes to extract
+     * @throws SignatureException If the Signature instance throws this exception
      */
-    public void writeTo(@NonNull final Signature signature) throws SignatureException {
-        signature.update(buffer, start, length);
+    public void updateSignature(@NonNull final Signature signature, final int offset, final int length)
+            throws SignatureException {
+        signature.update(buffer, Math.toIntExact(start + offset), length);
```
anthony-swirldslabs, line 329: Minor: might want to validate the provided `offset` and `length` to ensure they are within this Bytes buffer's length and (start+length), and throw AIOOB if they aren't. The same comment for the `verifySignature` below, so perhaps a helper private `validateBounds` method is warranted here, which could be useful in other places in Bytes as well.

anthony-swirldslabs (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
lpetrovic05: I added such a method. It seems that it could be useful in other places as well

anthony-swirldslabs (approved):

PR #258, opened 2024-05-28
feat: add support for maps

### LINKED ISSUE #254: Add support of `map` field type to PBJ
### Problem

Previously we didn't need it this much but with https://github.com/hashgraph/hedera-services/issues/10988, we realized that there is a need in a such type to represent some parts of `PlatformState`

### Solution

Add support for `map` type to PBJ

### Alternatives

_No response_

anthony-swirldslabs:
**Description**:
Introducing support for map fields in PBJ.

I apologize for the size of the PR but it would be close to impossible to break it up into reasonable pieces because all the code is deeply interconnected and needs to be updated all at once when adding a new field type.

Short summary:
1. The compiler is updated to generate Protobuf and JSON models/codecs for messages with map fields.
2. In models, the maps are present as instances of a new `PbjMap` class that implements an immutable map and provides access to a list of sorted keys to support deterministic iteration over the map entries.
3. Per the spec and the implementation of maps in protoc, they are... strange. In general, on the wire they're presented as repeated entry messages with key/value fields. However, unlike normal protobuf messages, the entries always serialize key and value values, even when the values are "default" (e.g. "" for a string, or 0 for an int, etc.) This peculiarity required quite a bit of refactoring in the codec generator. All those `skipDefault`-related changes are here to support just this.

**Related issue(s)**:

Fixes #254 

**Notes for reviewer**:
1. All the tests, including the integration and fuzz tests pass.
2. A new test model `map.proto` is added to test various maps.

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Field.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Field.java
@@ -194,44 +194,48 @@ public interface Field {
 	 */
 	enum FieldType {
 		/** Protobuf message field type */
-		MESSAGE("Object", "null", TYPE_LENGTH_DELIMITED),
+		MESSAGE("Object", "Object", "null", TYPE_LENGTH_DELIMITED),
 		/** Protobuf enum(unsigned varint encoded int of ordinal) field type */
-		ENUM("int", "null", TYPE_VARINT),
+		ENUM("int", "Integer", "null", TYPE_VARINT),
 		/** Protobuf int32(signed varint encoded int) field type */
-		INT32("int", "0", TYPE_VARINT),
+		INT32("int", "Integer", "0", TYPE_VARINT),
 		/** Protobuf uint32(unsigned varint encoded int) field type */
-		UINT32("int", "0", TYPE_VARINT),
+		UINT32("int", "Integer", "0", TYPE_VARINT),
 		/** Protobuf sint32(signed zigzag varint encoded int) field type */
-		SINT32("int", "0", TYPE_VARINT),
+		SINT32("int", "Integer", "0", TYPE_VARINT),
 		/** Protobuf int64(signed varint encoded long) field type */
-		INT64("long", "0", TYPE_VARINT),
+		INT64("long", "Long", "0", TYPE_VARINT),
 		/** Protobuf uint64(unsigned varint encoded long)  field type */
-		UINT64("long", "0", TYPE_VARINT),
+		UINT64("long", "Long", "0", TYPE_VARINT),
 		/** Protobuf sint64(signed zigzag varint encoded long) field type */
-		SINT64("long", "0", TYPE_VARINT),
+		SINT64("long", "Long", "0", TYPE_VARINT),
 		/** Protobuf float field type */
-		FLOAT("float", "0", TYPE_FIXED32),
+		FLOAT("float", "Float", "0", TYPE_FIXED32),
 		/** Protobuf fixed int32(fixed encoding int) field type */
-		FIXED32("int", "0", TYPE_FIXED32),
+		FIXED32("int", "Integer", "0", TYPE_FIXED32),
 		/** Protobuf sfixed int32(signed fixed encoding int) field type */
-		SFIXED32("int", "0", TYPE_FIXED32),
+		SFIXED32("int", "Integer", "0", TYPE_FIXED32),
 		/** Protobuf double field type */
-		DOUBLE("double", "0", TYPE_FIXED64),
+		DOUBLE("double", "Double", "0", TYPE_FIXED64),
 		/** Protobuf sfixed64(fixed encoding long) field type */
-		FIXED64("long", "0", TYPE_FIXED64),
+		FIXED64("long", "Long", "0", TYPE_FIXED64),
 		/** Protobuf sfixed64(signed fixed encoding long) field type */
-		SFIXED64("long", "0", TYPE_FIXED64),
+		SFIXED64("long", "Long", "0", TYPE_FIXED64),
 		/** Protobuf string field type */
-		STRING("String", "\"\"", TYPE_LENGTH_DELIMITED),
+		STRING("String", "String", "\"\"", TYPE_LENGTH_DELIMITED),
 		/** Protobuf bool(boolean) field type */
-		BOOL("boolean", "false", TYPE_VARINT),
+		BOOL("boolean", "Boolean", "false", TYPE_VARINT),
 		/** Protobuf bytes field type */
-		BYTES("Bytes", "Bytes.EMPTY", TYPE_LENGTH_DELIMITED),
+		BYTES("Bytes", "Bytes", "Bytes.EMPTY", TYPE_LENGTH_DELIMITED),
 		/** Protobuf oneof field type, this is not a true field type in protobuf. Needed here for a few edge cases */
-		ONE_OF("OneOf", "null", 0 );// BAD TYPE
+		ONE_OF("OneOf", "OneOf", "null", 0 ),// BAD TYPE
+		// On the wire, a map is a repeated Message {key, value}, sorted in the natural order of keys for determinism.
+		MAP("Map", "Map", "Collections.EMPTY_MAP", TYPE_LENGTH_DELIMITED );
 
 		/** The type of field type in Java code */
 		public final String javaType;
+		/** The type of boxed field type in Java code */
+		public final String boxedType;
 		/** The field type default value in Java code */
 		public final String javaDefault;
 		/** The protobuf wire type for field type */
@@ -244,8 +248,9 @@ public interface Field {
 		 * @param javaDefault The field type default value in Java code
 		 * @param wireType The protobuf wire type for field type
 		 */
-		FieldType(String javaType, final String javaDefault, int wireType) {
+		FieldType(String javaType, final String boxedType, final String javaDefault, int wireType) {
 			this.javaType = javaType;
+			this.boxedType = boxedType;
 			this.javaDefault = javaDefault;
 			this.wireType = wireType;
 		}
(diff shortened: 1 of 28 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/MapField.java
```diff
@@ -0,0 +1,153 @@
...
+                        null,
+                        "An internal, private map entry key for " + mapContext.mapName().getText(),
+                        false,
+                        null),
+                new SingleField(
+                        false,
+                        FieldType.of(mapContext.type_(), lookupHelper),
+                        2,
+                        "___" + mapContext.mapName().getText() + "__value",
+                        mapContext.type_().messageType() == null ? null : mapContext.type_().messageType().messageName().getText(),
```
imalygin, line 61: Please, refactor these parameters into method like `extractMessageTypeName` to make the mode more readable.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -87,15 +88,16 @@ private static RuntimeException unsupported() {
      * @param out The data output to write to
      * @param field the descriptor for the field we are writing
      * @param value the int value to write
+     * @param skipDefault default value results in no-op for non-oneOf
      */
-    public static void writeInteger(WritableSequentialData out, FieldDefinition field, int value) {
+    public static void writeInteger(WritableSequentialData out, FieldDefinition field, int value, boolean skipDefault) {
```
imalygin, line 93: It makes sense to create overloaded method 
```
public static void writeInteger(WritableSequentialData out, FieldDefinition field, int value) {
   writeInteger(out, field, value, true);
}
```
This will reduce your changeset.

imalygin (approved):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -74,33 +98,33 @@ static String generateParseMethod(final String modelClassName, final List<Field>
                             // So we catch this exception here and **only** here, because an EOFException
                             // anywhere else suggests that we're processing malformed data and so
                             // we must re-throw the exception then.
-                            final int tag;
+                            final int $prefixtag;
```
jasperpotts, line 101: Seems a bit confusing if I am reading this right. We so far have been using `$xxxx` as placeholders for replacement in strings. Looks like you are now introducing using `$` prefixed variables in generated code as well. This makes the code generation harder to read in my opinion as you do not know if it is a runtime variable or a string substitution.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -109,38 +133,27 @@ static String generateParseMethod(final String modelClassName, final List<Field>
...
-                    } catch (final Exception anyException) {
-                        if (anyException instanceof ParseException parseException) {
-                            throw parseException;
-                        }
-                        throw new ParseException(anyException);
-                    }
-                }
                 """
-        .replace("$modelClassName",modelClassName)
-        .replace("$fieldDefs",fields.stream().map(field -> "    %s temp_%s = %s;".formatted(field.javaFieldType(),
```
jasperpotts, line 138: Seems like lots more complex replaces here than new code but I do not see them used in old code. Were they just old?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -303,6 +316,46 @@ private static void generateFieldCaseStatement(final StringBuilder sb, final Fie
                     .replace("$maxSize", String.valueOf(field.maxSize()))
                     .indent(DEFAULT_INDENT)
             );
+        } else if (field.type() == Field.FieldType.MAP) {
+            // This is almost like reading a message above because that's how Protobuf encodes map entries.
+            // However(!), we read the key and value fields explicitly to avoid creating temporary entry objects.
+            final MapField mapField = (MapField) field;
+            final List<Field> mapEntryFields = List.of(mapField.keyField(), mapField.valueField());
+            sb.append("""
+						final var __map_messageLength = input.readVarInt(false);
```
jasperpotts, line 325: Here you seem to use a double underscore prefix. `__` to variable names. Would nice to be consistent.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/PbjMap.java
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Implements an immutable map that exposes a list of keys sorted in their natural order.
+ * <p>
+ * This Map implementation allows one to iterate the entries in a deterministic order
+ * which is useful for serializing, hash computation, etc.
+ *
+ * @param <K> key type
+ * @param <V> value type
+ */
+public class PbjMap<K, V> implements Map<K, V> {
```
jasperpotts, line 19: What is the advantage of this over `Collections.unmodifiableSortedMap()` ?

jasperpotts (changes requested):
Impressive work, that was a lot of work :-)

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/MapField.java
anthony-swirldslabs: Good idea. I implemented them as static members in Field and refactored both the MapField and the SingleField to use them.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
anthony-swirldslabs: All right. This should also help reduce any conflicts with code in hedera-services, I suppose. So I introduced the overloaded helper methods as you suggest.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Field.java
```diff
@@ -244,8 +248,9 @@ enum FieldType {
 		 * @param javaDefault The field type default value in Java code
 		 * @param wireType The protobuf wire type for field type
 		 */
-		FieldType(String javaType, final String javaDefault, int wireType) {
+		FieldType(String javaType, final String boxedType, final String javaDefault, int wireType) {
```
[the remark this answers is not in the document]
anthony-swirldslabs: Good catch. I've added the missing javadoc.

The reason for the `boxedType` is the fact that Java generics don't support primitive types. So for example, if one defines `map<int32, string>` in protobuf, we cannot translate it to `Map<int, String>` in Java, we have to write `Map<Integer, String>` instead because a map key is an object and not a primitive value. The current Field/FieldType implementation in PBJ doesn't provide a way to get the boxed type, so I had to add one.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
anthony-swirldslabs: You're misreading this part a bit. Here I'm introducing a placeholder named `$prefix` that is being `replace()`'d at line 154 below. So the `$prefixtag` is in fact a concatenation of the `$prefix` value replaced at line 154 and a literal string `tag`. So this code doesn't introduce any actual variable names starting with `$` in the actual generated code - they all get replaced.

Note that if the `$prefix` value at line 154 happens to be an empty string, then this particular variable will be named just `tag` as it's always been previously. The reason I need to introduce the `$prefix` is to support nesting the parsing code. This is necessary because, as stated elsewhere, for performance reason we don't model map entries as actual messages, and hence we cannot reuse the regular message parsing code path which would delegate to the message's PROTOBUF.parse() method. Instead we simply nest the map parsing loop inside the main parsing loop. But this creates a name conflict for Java variables such as this `tag` and javac simply fails (even though this nested loop is a separate `{}` block - it still won't let me overload a local variable name, and I think for a good reason.) So I just added this `$prefix` to resolve the conflict.

However, I agree that this looks a bit confusing, and the confusion mainly arises from the fact that "prefixtag" reads as a single word from human perspective while in fact it's two separate tokens from this code perspective.

If you wish, I could change the case of the prefix and change this to e.g. `$PREFIXtag`, but I'm unsure if this is too very helpful.

For now, I'm leaving this as is. But if you have any concrete suggestion on how to improve this part, I'm happy to implement it. Please let me know.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
anthony-swirldslabs: This is just an odd way how git displays diffs. All the old replacement are still alive and well, and they reside where they always have - above at lines 80-90. I just refactored that above code and extracted the `parseLoop` generator into a dedicated method and here in this method we only need a few replacements. But the old replacements are still there - just a little above.

Again, this refactoring is done so that maps could generate their own `parseLoop` w/o me having to duplicate almost identical code generation code here.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
anthony-swirldslabs: Yes, I do, and I do so by design. This particular change is to support parsing maps with values of type message.

Maps in protobuf are serialized as repeated entry messages, so I'm using a code that is very similar to parsing a message (but not quite the same) in order to parse the map - it's the code we're looking at now. However, when the map value type is a message, this code generator would insert a code snippet related to parsing that message inside of this map parsing code, and this will introduce an inner code in this parsing loop that would again use a variable name `messageLength`, which javac wouldn't allow us to do.

So I'm adding a prefix to the map-related variables and making them look distinct enough so that a human reader of the generated code can quickly spot what `messageLength` the code is dealing with. I thought a double underscore prefix is distinct enough. (I also use triple underscore prefix for map synthetic field names, btw, and again, I do so by design - to try and avoid any conflicts with regular field names.)

Please let me know if you're strongly opposed to multiple underscores and if so, please provide suggestions on how you'd want this to look instead.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/PbjMap.java
anthony-swirldslabs: Performance and code simplicity.

The `Collections.unmodifiableSortedMap()` accepts a `SortedMap` already (it cannot take a regular `Map`), so for starters we'd have to build one when we're parsing a message. Building a SortedMap (e.g. a TreeMap) that keeps its whole entries in a particular order may be a bit less efficient than building a simple HashMap and then only sorting a set of its keys.

Further, regular map operations (e.g. `get(key)`) would be O(log N) for TreeMap instead of O(1) for HashMap, further degrading the performance of the client code that uses the parsed model object and queries the map.

So I went with the new `PbjMap` instead.

imalygin (approved):

jasperpotts (approved):

Issue #255, opened 2024-05-24
Generate service stubs


rbair23:
### Problem

PBJ generates Java objects for each `message` it encounters in the protobuf schema definitions, but it does not generate anything for `service` definitions. Right now the only way to use PBJ as a library in a web server is to use a very low-level API provided by the gRPC libraries that basically deliver you a byte array from which you can parse. An example of how this can be done can be found in the [Hashgraph consensus node repo](https://github.com/hashgraph/hedera-services/tree/3f7390a8862073f9aef8ed2f6b25ce87efc6aa99/hedera-node/hedera-app/src/main/java/com/hedera/node/app/grpc/impl/netty).

### Solution

Create two new files:
  - `ServiceInterface`: this represents the `service` defined in the protobuf schema files. In addition to some helper methods, it contains the actual interface definition for the different methods defined on the `service`, such that a user can implement this interface in their code. It also contains default methods for various aspects of the actual handling of the call.
  - `ServiceMethod`: a simple definition of the method on the `service` -- what kind of method it is (unary, server streaming, client streaming, bidi streaming), and what it is named.

```java
/**
 * Defines a common interface for all implementations of a gRPC {@code service}. PBJ will generate a subinterface
 * for each {@code service} in the protobuf schema definition files, with default implementations of each of the
 * given methods in this interface.
 *
 * <p>For example, suppose I have the following protobuf file:
 * <pre>
 * {@code
 * package example;
 *
 * service HelloService {
 *   rpc SayHello (HelloRequest) returns (HelloResponse);
 * }
 *
 * message HelloRequest {
 *   string greeting = 1;
 * }
 *
 * message HelloResponse {
 *   string reply = 1;
 * }
 * }
 * </pre>
 *
 * <p>From this file, PBJ will generate a {@code HelloService} interface that extends {@code ServiceInterface}:
 * <pre>
 * {@code
 * public interface HelloService extends ServiceInterface {
 *    // ...
 *
 *    @NonNull
 *    HelloResponse sayHello(final @NonNull HelloRequest request);
 *
 *    default String serviceName() { return "HelloService"; }
 *    default String fullName() { return "example.HelloService"; }
 *
 *    // ...
 * }
 * </pre>
 *
 * In the application code, you will simply create a new class implementing the {@code HelloService} interface, and
 * register it with your webserver in whatever way is appropriate for your webserver.
 */
public interface ServiceInterface {
    /** Defines a method on the {@code service} IDL definition */
    interface Method {
[...]

PR #259, opened 2024-05-29
Initial upload of a GRPC implementation for Helidon

rbair23:
This PR introduces a new module, `pbj-grpc-helidon`. This is a [Helidon](https://helidon.io/) specific module (for version 4.0.8+) that provides the glue between HTTP2 (implemented by Helidon) and PBJ service endpoints (generated by PBJ based on a service definition in the protobuf schema files).

This PR does not introduce changes to the compiler to generate service interfaces or stubs. Rather, it assumes some other PR will do that, and it focuses on defining the runtime API and the Helidon module. Changes to the compiler will be done by #255.

In this PR:
- The build is upgraded to Java 21. The consensus node uses Java 21, as does Helidon.
- The module `pbj-grpc-helidon` is added
- The module `pbj-grpc-helidon-config` is added
- The interface `ServiceInterface` is added to `pbj-runtime` along with other gRPC interfaces

# gRPC

gRPC is an HTTP2-based protocol for message passing. The [specification](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) and the [documentation](https://grpc.io/docs/) show how this is done. Both `application/grpc+proto` (protobuf based messages) and `application/grpc+json` (JSON-encoded protobuf) are supported.

A separate specification for [web-grpc](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) exists, but is not supported by this PR.

# `pbj-grpc-helidon`

Public API in this module used directly by users include `PbjRouting`, and `GrpcStatus`.

Public API in this module used by Helidon directly include `PbjProtocolSelector`, `PbjProtocolProvider`, `PbjProtocolConfigProvider`, and `PbjConfig`.

All other entities are private to the module.

## Design

Helidon uses the `ServiceLoader` to identify the `ProtocolConfigProvider` (used for configuration) and the `Http2SubProtocolProvider` to delegate HTTP2 calls to. Our implementation of `Http2SubProtocolProvider`, `PbjProtocolProvider`, is simply used to create (based on some provided configuration) the `Http2SubProtocolSelector`. This selector is where all the real works happens. For this module, that is `PbjProtocolSelector`.

If you look into the `PbjProtocolSelector`, you will find the `subProtocol` method. This method is called by Helidon. It is the main entry point for handling HTTP2 routes supported by this system. After some preliminary checks, it delegates to a "handler" (`PbjErrorProtocolHandler` or `PbjProtocolHandler`) to handle appropriate details (reading from and writing to HTTP2 streams).

The `PbjProtocolHandler` will delegate to a `ServiceInterface` to handle a particular method invocation, in such a way as to avoid all reflection. It provides a callback to the `ServiceInterface` through which the `ServiceInterface` can send start a request, stream messages, and close a request. The implementation of this callback handles all the glue with HTTP2.

That is, the classes in `pbj-grpc-helidon` are the glue between the HTTP code of Helidon (that knows nothing about gRPC or PBJ), and the PBJ `ServiceInterface`s, which know nothing about HTTP2.


jsync-swirlds:
A couple questions we should probably answer in the documentation for this PR:
What is the advantage of this approach over the existing [Helidon gRPC](https://helidon.io/docs/latest/se/grpc/server) implementation?
Could (or does) this work leverage the Helidon implementation to reduce the amount we must build ourselves?

pbj-core/pbj-grpc-helidon/src/test/java/http/HttpTest.java
```diff
@@ -0,0 +1,23 @@
...
+        final var latch = new CountDownLatch(1000);
+        for (int i=0; i<1000; i++) {
+            pool.submit(() -> {
+                final var client = WebClient.builder().baseUri("http://localhost:8080").build();
+                System.out.println(client.get().path("/greet").request().as(String.class));
+                latch.countDown();
+            });
+        }
+
+        latch.await();
```
artemananiev, line 21: `await(timeout, unit)` is preferable, so if something goes wrong, the test wouldn't hang forever

rbair23:
> A couple questions we should probably answer in the documentation for this PR: What is the advantage of this approach over the existing [Helidon gRPC](https://helidon.io/docs/latest/se/grpc/server) implementation? Could (or does) this work leverage the Helidon implementation to reduce the amount we must build ourselves?

@jsync-swirlds, I'll add these to the issue itself (since it is more about the motivation rather than the solution)

rbair23:
> Could (or does) this work leverage the Helidon implementation to reduce the amount we must build ourselves?

I have put a message on their issue: https://github.com/helidon-io/helidon/issues/5418. I will try to create a PR for Helidon that creates the right abstraction for us, such that our module could be truly minimal and hopefully avoid having to duplicate any work with what they're doing.

pbj-core/gradle/modules.properties
```diff
@@ -1,3 +1,4 @@
 com.github.spotbugs.annotations=com.github.spotbugs:spotbugs-annotations
 com.google.protobuf=com.google.protobuf:protobuf-java
 org.antlr.antlr4.runtime=org.antlr:antlr4-runtime
+com.hedera.node.hapi=com.hedera.hashgraph:hapi
```
anthony-swirldslabs, line 4: I'm not an expert in Gradle, so I've no idea if this adds a dependency on HAPI, but if it does and the HAPI happens to want to use PBJ, too, wouldn't this create a circular dependency?

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjConfigBlueprint.java
```diff
@@ -0,0 +1,54 @@
...
+@Prototype.Blueprint
+@Prototype.Configured
+@Prototype.Provides(ProtocolConfig.class)
+interface PbjConfigBlueprint extends ProtocolConfig {
+    /**
+     * Default maximum message size in bytes ({@value}).
+     *
+     * @see #maxMessageSize()
+     */
+    int DEFAULT_MAX_MESSAGE_SIZE = 1024*10; // 10KB
```
anthony-swirldslabs, line 17: It's best to include the units suffix to the name of the constant itself, e.g. `...SIZE_BYTES`. This way, when using this constant, it's always clear what the value actually represents.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjErrorProtocolHandler.java
```diff
@@ -0,0 +1,63 @@
...
+    PbjErrorProtocolHandler(final Http2StreamWriter streamWriter,
+                            final int streamId,
+                            final Http2StreamState currentStreamState,
+                            final Consumer<WritableHeaders<?>> headerCallback) {
+            this.streamWriter = streamWriter;
+            this.streamId = streamId;
+            this.currentStreamState = currentStreamState;
+            this.headerCallback = headerCallback;
+        }
+
```
anthony-swirldslabs, line 31: Nit: please reformat all this.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,340 @@
...
+                    final var num = Integer.parseInt(matcher.group(0));
+                    final var unit = matcher.group(1);
+                    final var deadline = System.nanoTime() + num * switch (unit) {
+                        case "H" -> 3600_000_000_000L;
+                        case "M" -> 60_000_000_000L;
+                        case "S" -> 1_000_000_000L;
+                        case "m" -> 1_000_000L;
+                        case "u" -> 1_000L;
+                        case "n" -> 1L;
+                        default -> throw new IllegalArgumentException("Invalid unit: " + unit);
```
anthony-swirldslabs, line 128: Might want to use the `TimeUnit.convert()` instead of hard-coding a bunch of magic numbers.

pbj-core/gradle/modules.properties
rbair23: It would have been OK in this case, as the project that HAPI depends on is `runtime` and the gradle plugin, but not the Helidon module. But anyway this was wrong, it was leftover from early stages where I was using HAPI as a shortcut to get PBJ objects for testing, but then I switched to protoc objects for testing.

I will switch back to PBJ object, but before I can do that, I need to work on the gradle build. To prevent this PR from losing focus, I decided to just use protoc for now and do a follow on PR to address the gradle build more holistically

pbj-core/config/detekt/detekt.yml
```diff
@@ -0,0 +1,807 @@
...
+  # - 'ClassCountProcessor'
+  # - 'FunctionCountProcessor'
+  # - 'PropertyCountProcessor'
+  # - 'ProjectComplexityProcessor'
+  # - 'ProjectCognitiveComplexityProcessor'
+  # - 'ProjectLLOCProcessor'
+  # - 'ProjectCLOCProcessor'
+  # - 'ProjectLOCProcessor'
+  # - 'ProjectSLOCProcessor'
+  # - 'LicenseHeaderLoaderExtension'
```
anthony-swirldslabs, line 26: Minor: could we remove commented-out lines here and in a few places below? And/or if this file comes from somewhere externally, can we add a comment at the top of the file with a reference to the original, similar to how checkstyle configs above mention Palantir?

PS. I'm also unsure how much of this config is relevant to PBJ (all those native code mentions below, js, android, and whatnot) and perhaps most lines could be removed to fall back to defaults? Or maybe nuke this entire file? What will break if we do?

pbj-core/gradle.properties
```diff
@@ -1,5 +1,5 @@
 # Version number
-version=0.8.7-SNAPSHOT
+version=0.9.0-SNAPSHOT
```
anthony-swirldslabs, line 2: The `0.9.0` introduced maps. You're introducing more new stuff. I suggest to bump up the version number. I don't have a strong opinion if this should be 0.9.1 or 0.10.0 given that your changes don't really break any existing APIs.

pbj-core/gradle/plugins/src/main/kotlin/com.hedera.pbj.protoc.gradle.kts
```diff
@@ -0,0 +1,38 @@
...
+    plugins {
+        create("grpc") {
+            artifact = "io.grpc:protoc-gen-grpc-java:1.64.0"
+        }
+    }
+
+    generateProtoTasks {
+        all().forEach {
+            it.plugins {
+                create("grpc")
```
anthony-swirldslabs, line 34: You may want to ask @jjohannes to review Gradle changes.

pbj-core/gradle/scripts/clear-gradle-cache.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# The location were this script can be found.
+SCRIPT_PATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 || exit ; pwd -P )"
+
+{
+  cd "$SCRIPT_PATH"/../..
+
+  pkill -9 -f gradle
+
+  rm -rvf ~/.gradle
```
anthony-swirldslabs, line 13: :) Something tells me this script belongs to your personal `~/bin/` :) But I don't mind having it here.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -254,6 +254,8 @@ public static String getFieldsHashCode(final List<Field> fields, String generate
                                 result = 31 * result + Integer.hashCode($fieldName.protoOrdinal());
                              }
                              """).replace("$fieldName", f.nameCamelFirstLower());
+				} else if (f.type() == Field.FieldType.MAP) {
+					generatedCodeSoFar += getMapHashCodeGeneration(generatedCodeSoFar, f);
```
anthony-swirldslabs, line 258: This is a part of the maps implementation that has been in mainline for a long time already. I suggest to remerge/rebase your branch so that the PR only shows your actual changes.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/DeadlineDetector.java
```diff
@@ -0,0 +1,25 @@
...
+     * Schedule a callback to be invoked when the deadline has been exceeded. Please note that no operating system
+     * can actually measure elapsed time with nanosecond precision, so the actual deadline may be exceeded by a small
+     * amount of time measuring in the microseconds or even milliseconds.
+     *
+     * @param deadline The deadline, in nanoseconds, from now.
+     * @param onDeadlineExceeded The callback to invoke when the deadline has been exceeded.
+     * @return A {@link ScheduledFuture} that can be used to cancel the deadline.
+     */
+    @NonNull
+    ScheduledFuture<?> scheduleDeadline(long deadline, @NonNull Runnable onDeadlineExceeded);
```
anthony-swirldslabs, line 24: Minor: `s/deadline/deadlineNanos/`

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoParserToolsTest.java
```diff
@@ -196,7 +196,7 @@ void testReadString_maxSize() throws IOException {
 
     @Test
     void testReadString_incomplete() throws IOException {
-        final int length = rng.nextInt(0, 100);
+        final int length = rng.nextInt(1, 100);
```
anthony-swirldslabs, line 199: This is a huge PR. And this lowers the quality of code review that can be provided, leading to potential hidden bugs.

You could've split it into multiple small manageable parts, for example:
1. Upgrade to Java 21
2. Update the checkstyle and similar configs
3. Introduce documentation
4. Introduce small utilities
5. Introduce core interfaces/implementations (perhaps still split even further)
6. Introduce a test service

Yes, this would require that many PRs. However, the quality of reviews, and as a result, the quality of the code could be improved this way. Also, small PRs should (in theory) lead to faster turn around times.

(Yes, a huge PR can also get a very fast `Approved`, with a caveat that nothing has actually been reviewed properly or at all. And this is not the right thing to do.)

rbair23: Ah, it is actually not that big, but the PR has a bunch of diffs that it shouldn't. Time for some git magic.

pbj-core/gradle/scripts/clear-gradle-cache.sh
rbair23: I just brought it across from services because I hit a spot where gradle was unhappy and needed to get wiped, but I can just remove it.

pbj-core/config/detekt/detekt.yml
rbair23: These, honestly, I think are useless. They're identical from the consensus node repo, but don't seem to get picked up

rbair23: OK I just removed them.

pbj-core/gradle/scripts/clear-gradle-cache.sh
rbair23: OK, that dropped the number of files in the diff by 40.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoParserToolsTest.java
rbair23: Ok, that should be better, all the noise is gone now. I still don't know why github does that, because if I do a diff from the branch (before I "fixed" it) and main, it shows me only these files as different. But the PR shows a ton more changes, even though they aren't actually different from main. I'm not sure why it does that, but I've got it happy now (although I had to force-push to the branch. Blah).

I could factor out the change to Java 21 into another PR, but that's almost the only change that isn't actually related to the helidon project now.

rbair23: Prepare for a deluge of small PRs 😄 😄

rbair23: https://github.com/hashgraph/pbj/pull/264

rbair23: https://github.com/hashgraph/pbj/pull/265

rbair23: https://github.com/hashgraph/pbj/pull/266

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+            new TrailerBuilder().send();
+            close();
+        }
+    }
+
+    /**
+     * Simple implementation of the {@link ServiceInterface.RequestOptions} interface.
+     */
+    private record Options(
+            String authority,
```
jsync-swirlds, line 565: Should this be `Optional<String>` to match the interface being implemented?

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+            }
+
+            // The end of the stream has been reached! It is possible that a bad client will send end of stream before
+            // all the message data we sent. In that case, it is as if the message were never sent.
+            if (header.flags(Http2FrameTypes.DATA).endOfStream()) {
+                entityBytesIndex = 0;
+                entityBytes = null;
+                currentStreamState = Http2StreamState.HALF_CLOSED_LOCAL;
+                incoming.onComplete();
+            }
```
anthony-swirldslabs, line 306: What about the `else` branch? Should it re-initialize the variables? Should it log an error of some sort? Or emit a metric? Or maybe keep spinning the loop above? Or tell the `incoming` anything?  Or...?


BTW, should the loop above exit gracefully if the state changes while it's spinning (like if someone calls close() from another thread or something?)

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+                // This should NEVER be reachable, because the matcher would not have matched.
+                default -> throw new GrpcException(GrpcStatus.INTERNAL, "Invalid unit: " + unit);
+            });
+            return deadlineDetector.scheduleDeadline(deadline, () -> {
+                route.deadlineExceededCounter().increment();
+                incoming.onError(new GrpcException(GrpcStatus.DEADLINE_EXCEEDED));
+            });
+        }
+
+        return new NoopScheduledFuture();
```
anthony-swirldslabs, line 369: This means the given `timeout` is invalid, indicating that the client is "bad". On one hand, it would be nice to log something or emit a metric. On the other hand, doing so could worsen the effects of DoS attacks. So, I'm unsure. Thoughts?

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+     * <p>The Response-Headers are normally sent, followed by the data, followed by trailers. But if the request fails
+     * right away, before any handling, then it is possible to send Trailers-Only instead.
+     */
+    private void sendResponseHeaders(
+            @Nullable final Header messageEncoding,
+            @NonNull final HttpMediaType contentType,
+            @NonNull final List<Header> customMetadata) {
+
+        // Some headers are http2 specific, the rest are used for the grpc protocol
+        final var grpcHeaders = WritableHeaders.create();
```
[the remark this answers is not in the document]
anthony-swirldslabs: Please see the very next line - number 389.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+        grpcHeaders.set(GRPC_ACCEPT_ENCODING, IDENTITY);
+        customMetadata.forEach(grpcHeaders::set);
+        if (messageEncoding != null) {
+            grpcHeaders.set(messageEncoding);
+        }
+
+        streamWriter.writeHeaders(http2Headers,
+                streamId,
+                Http2Flag.HeaderFlags.create(Http2Flag.END_OF_HEADERS),
+                flowControl.outbound());
```
anthony-swirldslabs, line 405: Kind of seconding to Austin above, are the http2Headers connected to the grpcHeaders at line 389, so that whatever you set at line 399 is available in the http2Headers? Because you never use the grpcHeaders directly after line 399.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+     * <pre>
+     *     Trailers → Status [Status-Message] *Custom-Metadata
+     *     Status → "grpc-status" 1*DIGIT ; 0-9
+     *     Status-Message → "grpc-message" Percent-Encoded
+     *     Percent-Encoded → 1*(Percent-Byte-Unencoded / Percent-Byte-Encoded)
+     *     Percent-Byte-Unencoded → 1*( %x20-%x24 / %x26-%x7E ) ; space and VCHAR, except %
+     *     Percent-Byte-Encoded → "%" 2HEXDIGIT ; 0-9 A-F
+     * </pre>
+     */
+    private class TrailerBuilder {
```
anthony-swirldslabs, line 420: Ah, the magic of nested classes... I see...
Given that the usage above looks a bit confusing, I'd prefer if this was an outer (or static inner) class. But I'm okay if you really want the code to look like magic.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+        }
+
+        @Override
+        public int compareTo(@NonNull final Delayed o) {
+            // Since all NoopScheduledFuture instances have "0" as the delay, any other Delayed instance with a non-0
+            // delay will come after this one.
+            return (int) (o.getDelay(TimeUnit.NANOSECONDS));
+        }
+    }
+}
```
anthony-swirldslabs, line 590: This class is really-really huge. I'd suggest to move all the inner classes to the outer level. Not only will this eliminate the magic mentioned above, but this would also reduce the size of this file and hence improve its maintainability.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolSelector.java
```diff
@@ -0,0 +1,160 @@
...
+                .description("The number of failed gRPC requests"));
+    }
+
+    /**
+     * Called by Helidon to create the sub-protocol for PBJ gRPC requests. The {@link SubProtocolResult} returned
+     * will be responsible for handling the request.
+     */
+    @Override
+    public SubProtocolResult subProtocol(ConnectionContext ctx,
+                                         HttpPrologue prologue,
```
anthony-swirldslabs, line 70: Minor: I suggest to move the first argument to a new line with a regular continuation indentation of 8 spaces and avoid a huge block of vertical whitespace here for readability.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolSelector.java
```diff
@@ -0,0 +1,160 @@
...
+                                         Http2Settings clientSettings, // unused
+                                         StreamFlowControl flowControl,
+                                         Http2StreamState currentStreamState,
+                                         Router router) {
+        this.requestCounter.increment();
+
+        // As per the specification, only POST requests are supported. I would have thought that the code here should
+        // return a response code of 405 (Method Not Allowed) if the method is not POST, but the code here just returns
+        // NOT_SUPPORTED. I'm not sure if this is technically correct.
+        if (prologue.method() != Method.POST) {
```
anthony-swirldslabs, line 84: This code is owned by us though and we can do what we want (in fact we could support GET, I think), so I'm unsure about the reasoning. If this particular behavior of "UNSUPPORTED" vs 405 is specified in the Helidon spec somewhere, please add a link to that doc here in code for reference.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolSelector.java
```diff
@@ -0,0 +1,160 @@
...
+                        currentStreamState,
+                        config,
+                        route,
+                        deadlineDetector));
+    }
+
+    /**
+     * A handler for the case where the path is not found.
+     */
+    private static final class RouteNotFoundHandler implements Http2SubProtocolSelector.SubProtocolHandler {
```
anthony-swirldslabs, line 114: Minor: where it makes sense, please consider extracting nested classes to the outer level for code maintainability reasons.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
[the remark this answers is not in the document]
litt3: Yup, I was missing the obvious, thanks

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
litt3: Fixed

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjMethodRoute.java
```diff
@@ -0,0 +1,133 @@
...
+                        .scope(SCOPE)
+                        .addTag(Tag.create(SERVICE_TAG, serviceName))
+                        .addTag(Tag.create(METHOD_TAG, methodName))
+                        .description("The number of gRPC requests that exceeded their deadline"));
+    }
+
+    @Override
+    @NonNull
+    PbjMethodRoute toPbjMethodRoute(@NonNull final HttpPrologue grpcPrologue) {
+        return this;
```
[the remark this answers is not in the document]
litt3: The method is required by the abstract parent class. Since a sibling implementation uses the parameter, it can't be removed.

I don't think there are any checks that need to be done here. The method is `toPbjMethodRoute`, on the `PbjMethodRoute` class. So it can just reflexively return itself.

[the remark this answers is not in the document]
anthony-swirldslabs: The provided `grpcPrologue` may describe a route that differs from the route of the `this` object, so there's some room for verification.

Thanks for clarifying about the sibling implementation.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+    /** The regular expression used to parse the grpc-timeout header. */
+    private static final String GRPC_TIMEOUT_REGEX = "(\\d{1,8})([HMSmun])";
+    private static final Pattern GRPC_TIMEOUT_PATTERN = Pattern.compile(GRPC_TIMEOUT_REGEX);
+
+    // Helidon-specific fields related to the connection itself
+    private final Http2Headers headers;
+    private final Http2StreamWriter streamWriter;
+    private final int streamId;
+    private final StreamFlowControl flowControl;
+    private Http2StreamState currentStreamState;
```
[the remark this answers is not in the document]
litt3: I changed this to be an atomic reference, and removed the `synchronization` keyword from the `close()` method, since the only other thing in `close` apart from setting `currentStreamState` is threadsafe

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+    private ScheduledFuture<?> deadlineFuture;
+    /**
+     * The bytes of the next incoming message. This is created dynamically as a message is received, and is never
+     * larger than the system configured {@link PbjConfig#maxMessageSizeBytes()}.
+     */
+    private byte[] entityBytes = null;
+    /** The current index into {@link #entityBytes} into which data is to be read. */
+    private int entityBytesIndex = 0;
+    /** The subscriber that will receive incoming messages from the client. */
+    private Flow.Subscriber<? super Bytes> incoming;
```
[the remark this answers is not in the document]
litt3: For now, I expanded the javadocs on the non-final member variables, to address threading considerations.

I'm not opposed to a more strict level of enforcement, but I'm not sure if it's necessary in this case.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+import java.util.concurrent.Flow;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
+
+/**
+ * Implementation of gRPC based on PBJ. This class specifically contains the glue logic for bridging between
+ * Helidon and the generated PBJ service handler endpoints. An instance of this class is created for each new
+ * connection, and each connection is made to a specific method endpoint.
+ */
```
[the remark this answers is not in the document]
litt3: @rbair23 From what I've seen looking at `Http2ServerStream`, it appears to me like everything in this protocol handler is called sequentially, apart from `close` which may execute as a scheduled callback. Does this coincide with your understanding?

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+
+            // FUTURE: If the client sends a "grpc-accept-encoding", and if we support one of them, then we should pick
+            // one and use it in the response. Otherwise, we should not have any compression.
+
+            // If the grpc-timeout header is present, determine when that timeout would occur, or default to a future
+            // that is so far in the future it will never happen.
+            final var timeout = requestHeaders.value(GRPC_TIMEOUT);
+            deadlineFuture = timeout.isPresent()
+                    ? scheduleDeadline(timeout.get())
+                    : new NoopScheduledFuture();
```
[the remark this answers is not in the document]
litt3: Changed to use `.map().orElse()`

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
litt3: @rbair23 I've reordered how these headers are set, so that the GRPC headers are all set, and then copied into http headers.

I'm assuming for now that the previous ordering was just a mistake, so let me know if there was some intention here that I'm not understanding.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
litt3: > What about the else branch

I don't think any else branch is necessary. The `else` is the "standard" case, where the method call isn't complete, and we're just waiting on more data from the caller. In this case, the `data` method is going to be called again at some point in the future, and we need to continue holding these variables to keep track of what we've read so far.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -0,0 +1,590 @@
...
+
+            // Setup the subscribers. The "outgoing" subscriber will send messages to the client. This
+            // is given to the "open" method on the service to allow it to send messages to the client.
+            final var outgoing = new SendToClientSubscriber();
+            incoming = route.service().open(route.method(), options, outgoing);
+        } catch (final GrpcException grpcException) {
+            route.failedGrpcRequestCounter().increment();
+            new TrailerOnlyBuilder().grpcStatus(grpcException.status())
+                    .statusMessage(grpcException.getMessage())
+                    .send();
```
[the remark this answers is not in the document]
litt3: I agree that the nested classes can be confusing.

For now, I think I'll leave it as-is, since Richard is wanting this code to merge as a baseline, and then be modified / improved in the future. Since I'm taking over a bit to try to get this merged, it will allow me to avoid doing a lot of rewrites of Richard's code

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
litt3: See my comment [above](https://github.com/hashgraph/pbj/pull/259#discussion_r1765501729), I'm leaning towards leaving the magic for now.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
litt3: See my comment [above](https://github.com/hashgraph/pbj/pull/259#discussion_r1765501729)

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolSelector.java
litt3: I've not managed to find any helidon docs specifying this, @rbair23 do you have a source?

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolSelector.java
litt3: I extracted this internal class. I didn't see any other candidates for extraction that aren't wrapped up in magic, though

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
anthony-swirldslabs: Ah, I missed the idea of the `data()` method being called multiple times for the same request. This makes sense then. Thanks for the clarification.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
anthony-swirldslabs: The updated code looks good to me now.

anthony-swirldslabs (approved):

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolSelector.java
jsync-swirlds: Might be to meet the requirements of the [gRPC specification](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)

litt3: This does appear to be the case. Adding a link in the comment.

pbj-core/pbj-grpc-helidon-config/README.md
```diff
@@ -0,0 +1,10 @@
+# pbj-grpc-helidon-config
+
+This project contains just the configuration definition for the
+`pbj-grpc-helidon` module.
+
+Helidon modules require a "config blueprint". An annotation processor takes that
+"blueprint" and generates some metadata in META-INF and some code. The module
+then needs to compile against the generated code. Since Gradle is not capable of
+doing this in a single build, we have to split the configuration into a separate
```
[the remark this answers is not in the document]
litt3: I don't think I have the necessary insight to clarify this point. @rbair23 any thoughts here, that could help expand on this comment?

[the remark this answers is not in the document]
rbair23: I don't know. I agree @anthony-swirldslabs, annotation processors are common and it seems like this should work. I'm not sure why it is a problem in this case.

PR #261, opened 2024-06-04
Design document and diagrams for the PBJ gRPC Helidon module.

rbair23:
Design / architecture documentation for this new module. When approved, subsequent PRs will contain the implementation in phases.


pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,450 @@
...
+            }
+
+            @Override
+            public void onError(Throwable throwable) {
+                replies.onError(throwable);
+            }
+
+            @Override
+            public void onComplete() {
+                replies.onComplete();
```
anthony-swirldslabs, line 373: `s/replies/request/` at lines 361, 368, and 373? Or where does the `replies` come from otherwise?

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,450 @@
...
+7. If `Greeter.sayHelloStreamReply` is called and sends `onError`, the client receives an error and the connection is terminated
+8. If `Greeter.sayHelloStreamRequest` is called with invalid input, a `GrpcException` is thrown and turns into the expected gRPC error on the client
+9. If `Greeter.sayHelloStreamRequest` is called and sends `onError`, the client receives an error and the connection is terminated
+10. If `Greeter.sayHelloStreamBidi` is called with invalid input, a `GrpcException` is thrown and turns into the expected gRPC error on the client
+11. If `Greeter.sayHelloStreamBidi` is called and sends `onError`, the client receives an error and the connection is terminated
+12. The server passes the standard set of [gRPC tests](https://github.com/grpc/test-infra?tab=readme-ov-files)
+
+## Performance Testing
+
+1. Performance on the server is comparable to other solutions (see [Performance Testing](https://github.com/LesnyRumcajs/grpc_bench/blob/master/java_vertx_grpc_bench/Dockerfile)
```
anthony-swirldslabs, line 404: Nit: the final closing `)` is missing.

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,450 @@
+# Architectural Design
```
mattp-swirldslabs, line 1: Are there any plans to support [gRPC Reflection](https://grpc.io/docs/guides/reflection/) to promote discovery?  Java Greeter Reflection [example](https://github.com/grpc/grpc-java/tree/master/examples/example-reflection).  It looks like there was [support in Helidon v3](https://helidon.io/docs/v3/apidocs/io.helidon.grpc.server/io/helidon/grpc/server/ProtoReflectionService.html) but I don't see the same class in v4 at this time.

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,450 @@
...
+        };
+    }
+}
+```
+
+The bidirectional streaming pipeline is similar to the client streaming pipeline, except that the `replies` subscriber
+may be called more than once.
+
+![Bidirectional Streaming Pipeline](images/bidi-streaming.png)
+
```
mattp-swirldslabs, line 384: Where do you anticipate handling backpressure?

pbj-core/pbj-grpc-helidon/docs/design.md
rbair23: Great question! I think we should add it into the "Future Work" section.

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,450 @@
...
+A unary call is a simple request/response call. This is modeled in PBJ as a simple method call, where the method takes
+the request object, and returns the response object. It may optionally throw a `GrpcException` to indicate an error with
+a gRPC error code, or a `RuntimeException` to indicate an internal server error.
+
+#### Streaming Calls
+
+Each of the remaining three types of RPC calls involve streaming. Streaming requires the application developer to deal
+with connection termination, cancellation, and error handling. The approach taken by PBJ is similar to that taken in
+`grpc.io`, but instead of using a library-specific `StreamObserver` object, PBJ uses implementations of the new reactive
+`Flow` API in Java 21 and associated interfaces.
```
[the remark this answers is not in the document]
rbair23: Ah, you are probably right. I didn't look up where it was introduced.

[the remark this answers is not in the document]
rbair23: Fixed

pbj-core/pbj-grpc-helidon/docs/design.md
rbair23: Yes. However, I feel like the application code on the server needs a better API for working with streams, than creating a Flow.Subscriber from scratch. Maybe there should be a "GrpcStreamBuilder" or something with convenience API. I figure the right API will stare us in the face once we get some experience building with this API (it'll show us the pain points and we'll figure out how to smooth it out)

pbj-core/pbj-grpc-helidon/docs/design.md
rbair23: Good o'l CoPilot, innovating with variable names :). Will fix.

pbj-core/pbj-grpc-helidon/docs/design.md
rbair23: Done

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+`grpc.io`, but instead of using a library-specific `StreamObserver` object, PBJ uses implementations of the new reactive
+`Flow` API in Java 9 and associated interfaces.
+
+The `Flow` API has three primary methods on a `Flow.Subscriber`:
+  - `onNext`: called when a new message is available
+  - `onError`: called when an error occurs
+  - `onComplete`: called when the processing for this subscriber is complete
+
+It also includes some API for managing flow control, hence the name `Flow`. Each subscriber indicates to the publisher
+how many messages it is prepared to receive, and the publisher can then push that many messages (one at a time).
```
jsync-swirlds, line 103:
```suggestion
It also provides APIs for managing flow control. Each subscriber indicates to the publisher how many messages it is
prepared to receive in the next window, and the publisher can then push that many messages (one at a time).
```

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+  this module with the Helidon configuration system.
+- `com.hedera.pbj.grpc.helidon.PbjProtocolProvider`: the primary integration point, handles HTTP2 requests
+
+The `grpc` package of the PBJ runtime contains two classes to support grpc. These are the `ServiceInterface` and a
+utility class called `Pipelines`.
+
+`ServiceInterface` defines common methods related to all service interface implementations. Given any
+`ServiceInterface` implementation, you can introspect the service name, the "full name" (that is, the service name
+plus whatever "package" it is part of, such as would be used as the HTTP path to the service endpoint), and the list
+of methods (along with their names) that make up the service.
```
jsync-swirlds, line 120:
```suggestion
`ServiceInterface` defines common methods related to all service interface implementations. Given any
`ServiceInterface` implementation, one may introspect the service name, the "full name" (that is, the service name
plus whatever "package" it is part of, as might be used for the HTTP path to the service endpoint), and the list
of methods (along with their names) that make up the service.
```

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+
+Most importantly, the `ServiceInterface` defines the `open` method. This is called at the very beginning of the gRPC
+session creation. The `open` method is responsible for creating a `Flow.Subscriber` that will receive messages from
+the client. The method also receives the method that will be called, any request options, and the `Flow.Subscriber`
+that should be called for `onNext` (sending messages back to the client), `onError` (if there are errors), or
+`onComplete` (if the response is complete).
+
+This method may be called many times concurrently from different threads. Because this gRPC implementation is based
+on Helidon Nima, each request comes in on a single distinct Virtual Thread.
```
jsync-swirlds, line 129:
```suggestion
on Helidon Nima, each request is dispatched on a single distinct Virtual Thread.
```

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+
+## Pipelines
+
+The `Pipelines` class is a utility class that provides a way to create a "pipeline" of steps involved in processing
+requests and responses. This class is not intended to be used by the application developer. It is used by the code
+generated by the PBJ compiler.
```
jsync-swirlds, line 135: This might be confused with the platform pipelines.  Do we want to either rename this class or possibly join the two concepts?

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+are four types of pipelines: `unary`, `clientStreaming`, `serverStreaming`, and `bidiStreaming`. Each type has its
+own static method on `Pipelines` for creating that kind of pipeline.
+
+Internally, the pipelines use `Flow` classes, primarily `Flow.Subcriber` and `Flow.Subscription`. The `Flow` classes
+have support for flow control, cancellation, and error handling.
+
+## Unary Pipeline
+
+A unary call (which is a simple request/response call, or basic function), the application developer could implement a
+very simple method like this:
```
jsync-swirlds, line 209:
```suggestion
A unary call (which is a simple request/response call, or basic function), could be a
very simple method like this:
```

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+    }
+    
+    // ... other methods here
+}
+```
+
+As you can see, the method is trivial. It looks like a normal Java method. The PBJ runtime will handle all the work
+to marshal the request and response messages to and from the wire format, using PBJ codecs to do the work. The request
+is completed immediately upon return of the method. If any `RuntimeException` is thrown, then an appropriate error
+code will be returned to the client.
```
jsync-swirlds, line 225:
```suggestion
This method is trivial; it looks like a normal Java method with no special calls. The PBJ runtime will handle all the work
to marshal the request and response messages to and from the wire format, using PBJ codecs to do the work. The request
is completed immediately upon return of the method. If any `RuntimeException` is thrown, then an appropriate error
code will be returned to the client.
```

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+
+To have better control over error responses, the application developer can throw a `GrpcException`.
```
jsync-swirlds, line 227:
```suggestion
To have better control over error responses, the application developer should throw a `GrpcException`.
```

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+
+## Server Streaming Pipeline
+
+The "server streaming" RPC method is one where a single message is sent from the client to the server, and the server
+responds with a stream of messages.
+
+Here is an example, where after receiving a message, the server responds with 10 messages. For this case, the method
+signature contains the request object and a `Flow.Subscriber`. PBJ handles converting the request object from `Bytes`
+into the appropriate type. The `Flow.Subscriber` is used to send messages back to the client, or terminate the
+connection due to errors, or to complete the connection gracefully.
```
jsync-swirlds, line 259:
```suggestion
into the appropriate type. The `Flow.Subscriber` is used to send messages back to the client, to terminate the
connection due to errors, or to complete the connection gracefully.
```

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+
+The server streaming pipeline is almost identical to the unary pipeline, except that the `replies` subscriber may be
+called multiple times, and is actually exposed to the application developer.
+
+![Server Streaming Pipeline](images/server-streaming.png)
+
+## Client Streaming Pipeline
+
+The "client streaming" RPC method is one where many messages are sent from the client to the server, but only a single
+message is sent back from the server to the client at the end.
```
jsync-swirlds, line 284:
```suggestion
The "client streaming" RPC method supports many messages sent from the client to the server, but only a single
message sent back from the server to the client at the end.
```

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+the caller. This is the subscription to which the caller will push each message as it arrives from the client. Since
+we can only return a single object from the method, we need a callback into which to push the response message. We
+therefore see a `Flow.Subscription` supplied as an argument to the method, even though only a single message can
+be sent to the client. Sending more than one message will result in an error.
+
+```java
+@Override
+@NonNull
+public Flow.Subscriber<HelloRequest> sayHelloStreamRequest(@NonNull final Flow.Subscriber<HelloRequest> response) {
+    return new Flow.Subscriber<>() {
```
jsync-swirlds, line 297: nit: This might actually be easier to follow if the subscriber was not an anonymous inner class and response was passed via the constructor, it would certainly better highlight the simplicity of the stream request method in this case.

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+            response.onComplete();
+        }
+    };
+}
+```
+
+When the client streaming handler is built, it immediately calls the application method for that RPC call. That method
+return a `Flow.Subscriber`, F1. Then, as each message is received from the client, the `onNext` method of the request
+convert is called, which then calls `onNext` on F1 after converting from `Bytes` to the request object. When F1 calls
+any method on the `response` subscriber, the response is converted back to `Bytes` and sent back to the client.
```
jsync-swirlds, line 329:
```suggestion
When the client streaming handler is initiated, it immediately calls the application method for that RPC call. That method
returns a `Flow.Subscriber`, F1. Then, as each message is received from the client, the `onNext` method of the request
converter is called, which then calls `onNext` on F1 after converting from `Bytes` to the request object. When F1 calls
one of the methods on the `response` subscriber, the response is converted back to `Bytes` and sent back to the client.
```

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+            @Override
+            public void onComplete() {
+                replies.onComplete();
+            }
+        };
+    }
+}
+```
+
+The bidirectional streaming pipeline is similar to the client streaming pipeline, except that the `replies` subscriber
```
jsync-swirlds, line 380:
```suggestion
The bidirectional streaming pipeline is, therefore, very similar to the client streaming pipeline, except that the `replies` subscriber
```

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+2. `Greeter.sayHelloStreamReply` can be successfully called
+3. `Greeter.sayHelloStreamRequest` can be successfully called
+4. `Greeter.sayHelloStreamBidi` can be successfully called
+5. If `Greeter.sayHello` is called with invalid input, a `GrpcException` is thrown and turns into the expected gRPC error on the client
+6. If `Greeter.sayHelloStreamReply` is called with invalid input, a `GrpcException` is thrown and turns into the expected gRPC error on the client
+7. If `Greeter.sayHelloStreamReply` is called and sends `onError`, the client receives an error and the connection is terminated
+8. If `Greeter.sayHelloStreamRequest` is called with invalid input, a `GrpcException` is thrown and turns into the expected gRPC error on the client
+9. If `Greeter.sayHelloStreamRequest` is called and sends `onError`, the client receives an error and the connection is terminated
+10. If `Greeter.sayHelloStreamBidi` is called with invalid input, a `GrpcException` is thrown and turns into the expected gRPC error on the client
+11. If `Greeter.sayHelloStreamBidi` is called and sends `onError`, the client receives an error and the connection is terminated
```
jsync-swirlds, line 399: It may not be required to terminate the connection in this case.
Example:
Streaming validation server receives a stream of data items, and either acknowledges or errors for each one.
If an item is ack'd, that item is valid
If an item is error'd, that item is invalid
If the server sends "complete", then the server is terminating the connection (perhaps due to maximum single-stream duration, server shutdown, etc...)

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+
+1. If a client attempts to send a massive number of requests to a server, the server will only read as many as flow
+   control allows. The server will not crash or run out of memory.
+2. The server will limit the number of concurrent requests to a level where it remains stable and responses remain within
+   acceptable tolerances.
+
+## Malicious Testing
+
+1. If an attacker attempts to send a massive message during an RPC call, the server will reject the message outright
+   and close the connection because it restricts the number of bytes per message.
```
jsync-swirlds, line 416:
```suggestion
1. If an attacker attempts to send a massive message during an RPC call, the server will reject the message outright
   and close the connection after reading only enough bytes to exceed a configured limit. The server will not read the entire message.
```

anthony-swirldslabs (dismissed):

mattp-swirldslabs (approved):
Looks good

jsync-swirlds (commented):
A bunch of wording and nitpicks.
Mostly looks good.

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+message HelloRequest {
+  string name = 1;
+}
+
+message HelloReply {
+  string message = 1;
+}
+```
+
+The PBJ compiler will generate classes for `HelloRequest` and `HelloResponse`, and a `GreeterService` interface that
```
rbair23, line 76:
```suggestion
The PBJ compiler will generate classes for `HelloRequest` and `HelloReply`, and a `GreeterService` interface that
```

pbj-core/pbj-grpc-helidon/docs/design.md
rbair23: I'm probably forgetting, but I don't remember pipelines in the platform (are you thinking of the workflows in services, or the wiring framework in platform?). In any event, I think this is OK as they are internal to the PBJ module, and are not public or exposed to anyone.

pbj-core/pbj-grpc-helidon/docs/design.md
rbair23: I'm not sure I follow. Can you post a code example of what you have in mind?

pbj-core/pbj-grpc-helidon/docs/design.md
rbair23: In the gRPC spec, the only way to send a gRPC error is by sending the trailers with endStream. The application can create their own messages that represent application level errors, rather than gRPC errors, and it would work as you describe. But every onError results in a gRPC error.

pbj-core/pbj-grpc-helidon/docs/design.md
```diff
@@ -0,0 +1,455 @@
...
+
+## Malicious Testing
+
+1. If an attacker attempts to send a massive message during an RPC call, the server will reject the message outright
+   and close the connection because it restricts the number of bytes per message.
+
+# Future Work
+## Support for gRPC-Web
+
+Web clients are not able to use HTTP2 directly. They must use HTTP1.1. The `grpc.io` project promotes the use of an
```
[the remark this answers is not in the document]
rbair23: Ah, I should have said web browsers.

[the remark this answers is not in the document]
rbair23: The web browser supports http2, but as far as I know there is no standard javascript library you can use in a browser in order to make http2 calls. You just use normal XHR type calls and the browser may use HTTP1 or HTTP2. But it doesn't give the JS code the chance to interact with HTTP2 streams, server push, etc.

anthony-swirldslabs (dismissed):
+1

pbj-core/pbj-grpc-helidon/docs/design.md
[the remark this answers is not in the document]
jsync-swirlds: That makes more sense. Web browsers do control the transport mechanisms somewhat tightly. I don't know of an option to require/control HTTP2 transport in browsers either, but I'm not expert enough in Javascript to assert that there isn't one.

pbj-core/pbj-grpc-helidon/docs/design.md
jsync-swirlds: Interesting, I did not know that.
That seems like something we may wish to detail as a limitation of the module and consider carefully in server design. Having connections terminate for an application error can be far more disruptive than we'd prefer, depending on situation.

pbj-core/pbj-grpc-helidon/docs/design.md
jsync-swirlds: I was referring to platform wiring framework.
I agree that as long as the pipelines remain internal it's probably OK.

pbj-core/pbj-grpc-helidon/docs/design.md
jsync-swirlds: Just extract the new subscriber class to a static inner class so it's clearer how simple the subscriber method can be, and that the subscriber classes can (should) be reusable.
Something like this:
```java
@Override
@NonNull
public Flow.Subscriber<HelloRequest> sayHelloStreamRequest(@NonNull final Flow.Subscriber<HelloRequest> response) {
    return new JoinedEchoSubscriber(response);
}

/**
 * Simple subscriber that joins all of the streamed names together and
 * responds with a single "Hello" to all of those names joined together at
 * the end of the stream.
 */
private Static class JoinedEchoSubscriber extends Flow.Subscriber<HelloRequest> {
     private final List names = new ArrayList();
     private final Flow.Subscriber response;

     private JoinedEchoSubscriber(@NonNull final Flow.Subscriber<HelloRequest> response) {
         this.response = Objects.requireNonNull(response);
     }

     @Override
     public void onSubscribe(Flow.Subscription subscription) {
         subscription.request(Long.MAX_VALUE); // turn off flow control if you want
     }

     @Override
     public void onNext(HelloRequest item) {
         names.add(item.getName());
     }

     @Override
     public void onError(Throwable throwable) {
         response.onError(throwable);
     }

     @Override
     public void onComplete() {
         response.onNext(HelloReply.newBuilder()
                 .setMessage("Hello " + String.join(", ", names))
                 .build());
         response.onComplete();
     }
}
```

jsync-swirlds (dismissed):

pbj-core/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,445 @@
...
+        <property name="message" value="Found (&gt;&gt;&gt;&gt;&gt;&gt;&gt;), so it looks like you had a merge conflict that compiles. Please fix it."/>
+    </module>
+    <module name="RegexpSingleline">
+        <property name="format" value="\s+$"/>
+        <property name="message" value="Whitespace at end-of-line"/>
+    </module>
+    <module name="RegexpMultiline"> <!-- Java Style Guide: Vertical Whitespace -->
+        <property name="fileExtensions" value="java"/>
+        <property name="format" value="^\n\n$"/>
+        <property name="message" value="Two consecutive blank lines are not permitted."/>
```
anthony-swirldslabs, line 31: Minor: this is a bit too strict. An extra blank line might be useful to visually separate blocks of code. When used wisely, they don't hurt.

anthony-swirldslabs (approved):

PR #266, opened 2024-06-28
feat: Add API for gRPC. Compiler support later.

rbair23:
An API which the compiler will use to generate new types of files for supporting grpc Services.

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/GrpcStatus.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/GrpcStatus.java
@@ -17,7 +17,9 @@
 package com.hedera.pbj.runtime.grpc;
 
 /**
- * Status codes for gRPC. These are added to a {@link GrpcException} when an error occurs.
+ * Status codes for gRPC. These are added to a {@link GrpcException} when an error occurs. The ordinal of the enum
+ * <strong>exactly matches</strong> the status code in the gRPC protocol. The order in which these enums are defined
+ * is critical.
  *
  * @see <a href="https://grpc.github.io/grpc/core/md_doc_statuscodes.html">GRPC Status codes</a>
  */
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/GrpcStatus.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/GrpcStatus.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.hedera.pbj.runtime.grpc;
+
+/**
+ * Status codes for gRPC. These are added to a {@link GrpcException} when an error occurs.
+ *
+ * @see <a href="https://grpc.github.io/grpc/core/md_doc_statuscodes.html">GRPC Status codes</a>
+ */
+public enum GrpcStatus {
+    /**
+     * The operation completed successfully.
+     */
+    OK, // 0
+    /**
+     * The operation was cancelled (typically by the caller).
+     */
+    CANCELLED, // 1
+    /**
+     * Unknown error. An example of where this error may be returned is if a Status value received from another
+     * address space belongs to an error-space that is not known in this address space. Also, errors raised by APIs
+     * that do not return enough error information may be converted to this error.
+     */
+    UNKNOWN, // 2
+    /**
+     * Client specified an invalid argument. Note that this differs from FAILED_PRECONDITION. INVALID_ARGUMENT
+     * indicates arguments that are problematic regardless of the state of the system (e.g., a malformed file name).
+     */
+    INVALID_ARGUMENT, // 3
+    /**
+     * Deadline expired before operation could complete. For operations that change the state of the system, this
+     * error may be returned even if the operation has completed successfully. For example, a successful response
+     * from a server could have been delayed long enough for the deadline to expire.
+     */
+    DEADLINE_EXCEEDED, // 4
+    /**
+     * Some requested entity (e.g., file or directory) was not found.
+     */
+    NOT_FOUND, // 5
+    /**
+     * Some entity that we attempted to create (e.g., file or directory) already exists.
+     */
+    ALREADY_EXISTS, // 6
+    /**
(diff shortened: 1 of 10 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/GrpcStatus.java
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Status codes for gRPC. These are added to a {@link GrpcException} when an error occurs.
+ *
+ * @see <a href="https://grpc.github.io/grpc/core/md_doc_statuscodes.html">GRPC Status codes</a>
+ */
+public enum GrpcStatus {
+    /**
+     * The operation completed successfully.
+     */
+    OK, // 0
```
anthony-swirldslabs, line 30: Minor: not sure if the numbers in comments are necessary. Should you need to modify the enum, it would be difficult to update them properly. If they have to be equal to the values in the comments then maybe they should be explicitly assigned to a `private final` field via an enum constructor. Or, if this enum is already fixed in terms of its members and their positional numbers, then it's easier to simply mention that explicitly in the javadoc for the enum class itself rather than maintain these individual comments with ordinal numbers.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/ServiceInterface.java
```diff
@@ -0,0 +1,111 @@
...
+    interface RequestOptions {
+        String APPLICATION_GRPC = "application/grpc";
+        String APPLICATION_GRPC_PROTO = "application/grpc+proto";
+        String APPLICATION_GRPC_JSON = "application/grpc+json";
+
+        String authority();
+
+        boolean isProtobuf();
+
+        boolean isJson();
```
anthony-swirldslabs: Minor:
1. This is a bit limiting as more content types could become useful in the future (YAML, gzip, etc.)
2. This is error prone because an implementation of this interface could mistakenly return `true` for both the methods, and you then have to write code to handle this situation.

I think it could be better modeled with a `enum ContentType { ...; String type; }` and then have only a single method here, like `ContentType contentType()`. This enum could (now or in the future) help handle the different types, so as to hide these details from the code that actually has to parse the bytes. This would make the content type concept strongly typed, prevent potential bugs described above, and allow us to easily extend the support to more content types in the future.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/GrpcStatus.java
```diff
@@ -17,7 +17,9 @@
 package com.hedera.pbj.runtime.grpc;
 
 /**
- * Status codes for gRPC. These are added to a {@link GrpcException} when an error occurs.
+ * Status codes for gRPC. These are added to a {@link GrpcException} when an error occurs. The ordinal of the enum
+ * <strong>exactly matches</strong> the status code in the gRPC protocol. The order in which these enums are defined
+ * is critical.
```
anthony-swirldslabs, line 22: Awesome! Please see my comment for a previous revision - the individual comments could be removed then probably.

anthony-swirldslabs (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/GrpcStatus.java
rbair23: I should also have a unit test that fixes them in place.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/ServiceInterface.java
rbair23: The only problem I have with an enum is that it requires a code change to support a new content type, whereas this implementation would allow a new content type to be supported without any change in the library code. Otherwise I love the suggestion. 🤔

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/GrpcStatus.java
anthony-swirldslabs: Good idea.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/ServiceInterface.java
anthony-swirldslabs: I'm not sure how this is possible given these binary getters above? Any code that uses them would have to be updated if/when a new type is introduced.

However, if the content type logic is fully encapsulated in the enum implementation, then no code would need changing at all, apart from the enum itself by means of adding one more constant with proper initialization arguments to implement the support of the new content type.

anthony-swirldslabs: Although I agree that this would complicate the compiler a bit in order to generate the enum. So depending on how important/urgent it is for us to support more content types, we may decide to do this right now or not.

However, I disagree with the point of being able to add a new content type w/o any changes to PBJ. Your current sample code would simply fall into the !isProtobuf() branch which is equal to JSON unconditionally. So it wouldn't really support a random new content type out of the box anyway, unless the compiler is updated to support it properly.

anthony-swirldslabs (dismissed):

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/grpc/PipelinesTest.java
```diff
@@ -0,0 +1,585 @@
...
+            final var builder = Pipelines.<String, String>unary()
+                    .method(String::toUpperCase)
+                    .mapResponse(Bytes::wrap)
+                    .respondTo(replies);
+
+            assertThatThrownBy(builder::build).isInstanceOf(IllegalStateException.class);
+        }
+
+        @Test
+        void methodIsRequired() {
```
anthony-swirldslabs, line 87: Might want to add `.hasMessage(...)` checks to these tests to differentiate between various IllegalStateExceptions.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/grpc/PipelinesTest.java
```diff
@@ -0,0 +1,585 @@
...
+                    .mapResponse(Bytes::wrap)
+                    .respondTo(replies)
+                    .build();
+
+            pipeline.onSubscribe(mock(Flow.Subscription.class));
+            pipeline.onNext(Bytes.wrap("hello"));
+            pipeline.onNext(Bytes.wrap("world"));
+            pipeline.onComplete();
+            pipeline.onNext(Bytes.wrap("!!!"));
+            verify(replies).onError(any(IllegalStateException.class));
```
anthony-swirldslabs, line 256: While I realize that the purpose of the test is to verify the behavior after the `onComplete` is invoked, I'm wondering whatever happens on the `replies` side before or immediately after the `onComplete` call? It might be worth `verify()`'ing that too. Otherwise, the behavior remains undefined because the pipeline implementation might invoke `replies.onNext()`, or it might not, and nobody would know for sure if it does (and whether it should or shouldn't do that) just before the third `onNext()` call at line 255.)

anthony-swirldslabs (approved):

anthony-swirldslabs (approved):

Issue #284, opened 2024-09-06
writeByte() may be improved for byte array based BufferedData


artemananiev:
`ByteArrayBufferedData.writeByte()` method looks like this:

```java
    public void writeByte(final byte b) {
        validateCanWrite(1);
        final int pos = buffer.position();
        array[arrayOffset + pos] = b;
        buffer.position(pos + 1);
    }
```

Besides putting the byte to the byte array, it has a check and two calls to buffer.position() When multiple bytes are written one by one, this overhead becomes noticeable. For example, this `writeByte()` method is used in `Utf8Tools`, which is used to write all strings.

PR #285, opened 2024-09-09
fix: 284: writeByte() may be improved for byte array based BufferedData

artemananiev:
Fix summary:

* Introduced `writeByte2()`, `writeByte3()`, and `writeByte4()` methods to `WritableSequentialData`
* By default, the methods just call `writeByte()` the right number of times
* In `ByteArrayBufferedData` the methods are implemented more efficiently than N calls to `writeByte()`

Performance:

* In `ProtobufObjectBench`, writing `Everything` to byte array output is about 12% faster than previously

Fixes: https://github.com/hashgraph/pbj/issues/284

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/WritableSequentialData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/WritableSequentialData.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/WritableSequentialData.java
```diff
@@ -293,14 +327,8 @@ default void writeLong(final long value) throws BufferOverflowException, Uncheck
...
-        writeByte((byte)(value >>> 56));
-        writeByte((byte)(value >>> 48));
-        writeByte((byte)(value >>> 40));
-        writeByte((byte)(value >>> 32));
-        writeByte((byte)(value >>> 24));
-        writeByte((byte)(value >>> 16));
-        writeByte((byte)(value >>>  8));
-        writeByte((byte)(value));
+        writeByte4((byte) (value >>> 56), (byte) (value >>> 48), (byte) (value >>> 40), (byte) (value >>> 32));
+        writeByte4((byte) (value >>> 24), (byte) (value >>> 16), (byte) (value >>> 8), (byte) (value));
```
anthony-swirldslabs, line 331: Minor: might as well add `writeByte8`.

anthony-swirldslabs (approved):

imalygin (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/WritableSequentialData.java
artemananiev: It can be done, but looks like too corner case to me :)

PR #288, opened 2024-09-17
fix: throw on not enough bytes

### LINKED ISSUE #286: Codec returns partially read record
In a specific situation, a `Codec` can stop reading bytes and return a partially constructed object if no bytes are available. This happens when the record being parsed has a `bytes` field. If while reading these bytes the source has no more data, the codec will simply return whatever was read up to that point. 

The problem seems to be `ReadableSequentialData.readBytes(int)`:
```
    /**
     * Read {@code length} bytes from this sequence, returning them as a {@link Bytes} buffer of
     * the read data. The returned bytes will be immutable. The {@link #position()} of this sequence will be
     * incremented by {@code length} bytes.
     *
     * <p>Bytes are read from the sequence one at a time. If there are not {@code length} bytes remaining in this
     * sequence, then a {@link BufferUnderflowException} will be thrown. The {@link #position()} will be
     * incremented by the number of bytes read prior to the exception.
     *
     * @param length The non-negative length in bytes to read
     * @return new {@link Bytes} containing the read data
     * @throws IllegalArgumentException If {@code length} is negative
     * @throws BufferUnderflowException If there are not {@code length} bytes remaining in this sequence
     * @throws UncheckedIOException If an I/O error occurs
     */
    default @NonNull Bytes readBytes(final int length) throws BufferUnderflowException, UncheckedIOException {
        if (length < 0) {
            throw new IllegalArgumentException("Negative length not allowed");
        }

        if (remaining() < length) {
        }

        final var bytes = new byte[length];
        readBytes(bytes, 0, length);
        return Bytes.wrap(bytes);
    }
```

The javadoc says it is supposed to throw a `BufferUnderflowException` when it cannot read enough bytes. But the method 
`ReadableSequentialData.readBytes(byte[], int, int)` does not throw, it just reads as many bytes as it can and returns. 

Stack trace from debugger:
```
readBytes:192, ReadableStreamingData (com.hedera.pbj.runtime.io.stream)
readBytes:217, ReadableSequentialData (com.hedera.pbj.runtime.io)
readBytes:300, ProtoParserTools (com.hedera.pbj.runtime)
parse:114, GossipEventProtoCodec (com.hedera.hapi.platform.event.codec)
parse:28, GossipEventProtoCodec (com.hedera.hapi.platform.event.codec)
parse:74, Codec (com.hedera.pbj.runtime)
```

lpetrovic05:
closes #286

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ReadableSequentialData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ReadableSequentialData.java
@@ -214,7 +214,10 @@ public interface ReadableSequentialData extends SequentialData {
         }
 
         final var bytes = new byte[length];
-        readBytes(bytes, 0, length);
+        final long bytesRead = readBytes(bytes, 0, length);
+        if (bytesRead != length) {
+            throw new BufferUnderflowException();
+        }
         return Bytes.wrap(bytes);
     }
 
--- a/pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoParserToolsTest.java
+++ b/pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoParserToolsTest.java
@@ -250,8 +250,10 @@ class ProtoParserToolsTest {
         final byte[] bytes = data.toInputStream().readAllBytes();
         final byte[] incompleteCopy = new byte[bytes.length - 1];
         System.arraycopy(bytes, 0, incompleteCopy, 0, bytes.length - 1);
-        final ReadableStreamingData streamingData = new ReadableStreamingData(new ByteArrayInputStream(incompleteCopy));
-        assertThrows(BufferUnderflowException.class, () -> ProtoParserTools.readString(streamingData));
+        assertThrows(BufferUnderflowException.class,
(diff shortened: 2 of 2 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ReadableSequentialData.java
```diff
@@ -214,7 +214,10 @@ default long readBytes(@NonNull final BufferedData dst) throws UncheckedIOExcept
         }
 
         final var bytes = new byte[length];
-        readBytes(bytes, 0, length);
+        final long bytesRead = readBytes(bytes, 0, length);
+        if (bytesRead != length) {
+            throw new BufferUnderflowException();
```
anthony-swirldslabs, line 219: Minor: might want to provide some context in the exception message, like the number of expected bytes to read and the actual number read. That could help debugging this exception in the future.

anthony-swirldslabs (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ReadableSequentialData.java
lpetrovic05: `BufferUnderflowException` does not accept any arguments unfortunately. In order to add a message, we would have to change the exception type, but this would be a big change, since all the javadocs declare that this exception is thrown.

PR #311, opened 2024-10-24
fix: Propagate outgoing stream errors to incoming pipeline

### LINKED ISSUE #304: Helidon exceptions are not propagated up through PbjProtocolHandler to the application via `onError`
### Description

An exception thrown in `io.helidon.common.buffers.ReadOnlyArrayData` invokes the `onError()` method in PbjProtocolHandler but that method does not signal an error to the application layer via `Flow.Subscriber<? super Bytes> incoming`. This prevents the application layer from taking action and cleaning up, etc.

### Steps to reproduce

1. Put a breakpoint in the `PbjProtocolHandler.onError()` method on the running server: https://github.com/hashgraph/pbj/blob/a2e54fa0bd7b18afbeb88187af5ad30490eb7c94/pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java#L643-L654
2. Force an exception and note that it's caught in the `PbjProtocolHandler.onError()` method.
3. Observe that this method never calls `incoming.onError()` or an equivalent to notify the application layer there was a problem.

### Additional context

image

### Hedera network

_No response_

### Version

v0.9.9

### Operating system

macOS

litt3:
- closes #304

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,40 @@
+###################################
+##### Global Protection Rule ######
+###################################
+# NOTE: This rule is overriden by the more specific rules below. This is the catch-all rule for all files not covered by the more specific rules below.
+*                                               @hashgraph/devops-ci @hashgraph/release-engineering-managers
+
+#########################
+##### Source Files  #####
+#########################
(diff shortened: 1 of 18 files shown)


rbair23:
Closed by commit PR #328

PR #325, opened 2024-11-15
Convert from Flow.Subscriber to Pipeline

rbair23:
Change the usage from Flow.Subscriber to Pipeline


mattp-swirldslabs (approved):
Looks good

PR #327, opened 2024-11-15
Add support for testing with metrics

rbair23:
Adds a new dependency when **testing**, so as to have a real implementation of metrics.


mattp-swirldslabs (approved):
LGTM

PR #344, opened 2025-01-09
Eraly PR for Add compute on construction hashcode and protobuf size

### LINKED ISSUE #343: Add caching of hash code and protobuf size to model objects
### Problem

We calculate the hashcode and protobuf size of PBJ model objects many times over. Both are reasonably costly operations especially for large object trees. It would ideal if they were only calculated once and reused. As the model objects are immutable they should only need to be calculated once.

### Solution

As the model objects are Java records there is an issue that they can't have private fields or mutable fields. So we can't have lazy computation. There are 3 options I considered:

1. **Change Model Objects from Records to Classes** - This would be the most elegant API wise but would mean that model objects can not become value types with future Java project [Valhalla](https://openjdk.org/projects/valhalla/). That should offer a huge performance gain and reduction in short term garbage generation. So is very hard to give up on the promise.
2. **Cheat on immutability by adding reference field to a mutable object** - There are two sub ideas here a `int[]` field of size 2 with one for hash code and one for size. The other is a custom object with two int fields. Both would allow lazy computation but I am concerned how they will impact future value types. The second big issue is it means every model object has one more object creation. As we create millions of model objects a second this is a big garbage problem.
3. **Add two int fields to every model object and computation values on construction** - This is simple and the cheapest memory/garbage wise. It is ugly API-wise as there is no way to hide those fields. It also changes the timing of when computation cost is paid from use to object construction time. That may or may not be a problem.

The current plan is to try out option 3 and see if it is acceptable.

### Alternatives

_No response_

jasperpotts:
PR to fix #343


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -399,64 +487,51 @@ private static String generateConstructor(
...
+								+ ")";
+						sb.append("this.precomputedHashCode = "+computeHashCode+";");
+					} else if (field.nameCamelFirstLower().equals("protobufEncodedSize")) {
+						final String computeProtobufSize = "computeProtobufSize("
+								+ fields.stream()
+								.filter(f -> f.fieldNumber() != -1)
+								.map(f -> "this."+f.nameCamelFirstLower())
+								.collect(Collectors.joining(", "))
+								+ ")";
+						sb.append("this.protobufEncodedSize = "+computeProtobufSize+";");
```
artemananiev, line 505: I wonder if it should be done lazily rather than in the constructor. Are all protobuf objects measured sooner or later?

artemananiev: Ah, I see, these are Java records, so all fields must be initialized in the constructor and never changed in the future...

jasperpotts: Yep, trying changing to class

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -230,13 +263,30 @@ public record $javaRecordName(
...
-								+ field.javaFieldType() + " " + field.nameCamelFirstLower()
-				).collect(Collectors.joining(",\n")).indent(DEFAULT_INDENT))
 				.replace("$bodyContent", bodyContent);
 	}
 
+	private static String generateRecordStyleGetters(final List<Field> fields) {
+		return fields.stream().map(field -> {
+			String fieldComment = field.comment();
+			String fieldCommentLowerFirst = fieldComment.substring(0, 1).toLowerCase() + fieldComment.substring(1);
+            return """
```
jsync-swirlds, line 273: It generally helps to add comments to suppress spotless on these code blocks.

Also, we probably shouldn't use tabs for indentation in the generated classes; it is helpful if the generated classes at least partially follow the same code styles as the rest of the codebase...

&nbsp;

---
Github tip:  If you set tab size in github (Settings|Appearance|Tab size preference) to 12, it makes tab indentation _much_ easier to see in code review.

jsync-swirlds (commented):
A couple suggestions to use `printf` instead of string concatenation in `println`.

Also noticed all of the generated code blocks now use tabs instead of spaces, which actually makes it pretty unreadable, particularly the larger code blocks which mix tabs and spaces a lot.

rbair23 (approved):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -236,37 +448,50 @@ private static String generateHashCode(final List<Field> fields) {
...
                     long hashCode = result;
-                $hashCodeManipulation
-                    return (int)hashCode;
+            $hashCodeManipulation
+                    $hashCode = (int)hashCode;
                 }
-                """.replace("$hashCodeManipulation", HASH_CODE_MANIPULATION).indent(DEFAULT_INDENT);
+                return $hashCode;
+            }
+            """.replace("$hashCodeManipulation", HASH_CODE_MANIPULATION)
```
artemananiev, line 480: Should `$hashCode` be replaced with a field name, too?

artemananiev: Nevermind. I see `$hashCode` is a real field name (which is confusing, btw)

PR #348, opened 2025-01-10
Fixed all compiler and doc warnings

jasperpotts:
I had enough of all the warnings, this cleans them all up, YAY :-) Fixes #345

PR #381, opened 2025-01-24
bug: Fix a couple errors in pbj-integration that slipped in recent PRs

jsync-swirlds:
* Fixed TestHashFunctions class that had trivial warnings which fail compilation due to Heiro "Warnings are errors" setting.
   * Introduced in PR #109, but only became an error recently.
* Fixed a random annotation in NonSynchronizedByteArrayInputStream that was referencing the Jetbrains annotation and _between_ `byte` and `[]` in a `byte[]` return type.
   * Introduced in PR #348 and not detected because pbj-integration-tests is not running in the PBJ CI suite. ([example build scan](https://scans.gradle.com/s/3cjan7qcca5ny); the gradle cache skips the entire run).

@jjohannes Could you look into why the gradle tasks are skipping everything in most CI runs?

### DIFF
--- a/pbj-integration-tests/src/main/java/com/hedera/pbj/integration/NonSynchronizedByteArrayInputStream.java
+++ b/pbj-integration-tests/src/main/java/com/hedera/pbj/integration/NonSynchronizedByteArrayInputStream.java
@@ -6,7 +6,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Arrays;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


anthony-swirldslabs (approved):

jsync-swirlds:
@jjohannes Could you look into why the gradle tasks are skipping everything in most CI runs?

jjohannes:
https://github.com/hashgraph/pbj/pull/383

PR #302, opened 2024-10-09
build: use common Hiero Gradle config

jjohannes:
**Description**:

Before this can be integrated:

- [x] Publish _hiero-gradle-conventions_ plugins to Gradle Plugin Portal and use from there
- [x] https://github.com/hiero-ledger/hiero-gradle-conventions/issues/50

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/14255

PR #383, opened 2025-01-29
ci: re-enable integration tests on CI and align Java version on CI

jjohannes:
**Description**:
- Make sure 'PBJ Integration' steps are no longer skipped (issue likely introduced in #302)
- Set Java version on CI to same version defined in 'toolchain-versions.properties' files
- Fix exist quality check issues in `pbj-integration-tests`

**Related issue(s)**:

#302

### DIFF
--- a/.github/workflows/zxc-compile-pbj-code.yaml
+++ b/.github/workflows/zxc-compile-pbj-code.yaml
@@ -26,7 +26,7 @@ on:
         description: "Java JDK Version:"
         type: string
         required: false
-        default: "21.0.3"
+        default: "21.0.5"
       gradle-version:
         description: "Gradle Version:"
         type: string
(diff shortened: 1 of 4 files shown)


.github/workflows/zxc-compile-pbj-code.yaml
```diff
@@ -122,7 +122,7 @@ jobs:
 
       - name: Gradle Assemble (PBJ Integration)
         id: gradle-integration-build
-        if: ${{ (inputs.enable-integration-tests || inputs.enable-jmh-tests) && steps.gradle-publish-local.conclusion == 'success' && !cancelled() }}
+        if: ${{ (inputs.enable-integration-tests || inputs.enable-jmh-tests) && !cancelled() }}
```
jjohannes, line 125: This was the culprit: `gradle-publish-local` step no longer exists as we do things with `includeBuild` now.

Apparently, GH actions just silently evaluates this to some value (null, false, ... ?) and I missed it #302.

jsync-swirlds: Not sure if this is _expected_, but the pbj-core compile was, effectively, skipped over in _this_ PR (https://scans.gradle.com/s/qjwrhmpjtze3q).

anthony-swirldslabs (approved):

.github/workflows/zxc-compile-pbj-code.yaml
jjohannes: Yes the difference is that in this case Gradle was called but then was able to restore build results from the cache. Because this PR does not change anything in `pbj-core`, it was built before in this state by other CI builds.

The issue with the _integration-tests_ project is that (without this change) GH actions always skips the step and Gradle is not called at all.

imalygin (approved):

andrewb1269hg (approved):

Issue #392, opened 2025-02-13
Need a way to access raw field bytes during parsing


artemananiev:
Sample protobuf record:

```
message Outer {

    int64 longField = 1;

    Inner messageField = 2;

}

message Inner {

    // some fields here

}
```

In compile time, Java records are generated for both `Outer` and `Inner` messages, as well as PBJ codecs. The codecs provide `parse()` methods like this:

```java
public Outer parse(ReadableSequentialData in, boolean strictMode, int maxDepth) { ... }
```

Given a byte array, such `parse()` methods can be used to deserialize an `Outer` record from the array. The deserialized object will have the `longField` field of type long, and the `messageField` field of type `Inner`. So far so good.

In some cases, there is a need to get raw bytes for message fields, in this case, for `longField` or `messageField`. Primitive field types are easy: it's enough to serialize them back to bytes. Message fields are more complicated. Current PBJ codecs serialize all objects in field number order (ascending), while the original byte array received from a user may have arbitrary field order.

This ticket is to provide API to get access to original field bytes during parsing. A few random ideas:

* Introduce an optional parameter to `Codec.parse()` method to accept a listener, which is notified when a field is deserialized
* Provide a separate method to extract field bytes from given bytes. It can accept a `ReadableSequentialData` input similar to `parse()`, and a "path" to the field to extract bytes for. For example, if `Inner` message above has a message field `InnerInner innerMessage = 1`, then its path in the original `Outer` bytes will be `messageField.innerMessage` (field names) or `2.1` (field IDs)

PR #396, opened 2025-02-18
fix: 392: PBJ codecs to provide access to raw field bytes during parsing

artemananiev:
Fix summary:
* A new method is provided in `ProtoParserTools`: `extractRawFieldBytes()`
* This method doesn't depend on any codecs, this is why it's in `ProtoParserTools`, not in generated codec classes
* The method can't work with nested fields, e.g. there is no way to read field A in field B in the input. However, it's trivial to implement this in user code by calling the method twice:

```java
final Bytes externalFieldBytes = ProtoParserTools.extractRawFieldBytes(input, fieldB);
final Bytes internalFieldBytes = ProtoParserTools.extractRawFieldBytes(externalFieldBytes.toReadableSequentialData(), fieldA);
```

* If needed, a new utility method can be provided to cover this scenario. It will accept a list of field definitions and will chain the calls appropriately
* Unit tests for the new method are added

Fixes: https://github.com/hashgraph/pbj/issues/392


netopyr:
The API looks good to me. Thanks a lot @artemananiev!

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -305,6 +308,61 @@ public static Bytes readBytes(final ReadableSequentialData input, final long max
...
+     * @param input The input to read from
+     * @param field Field definition to extract bytes for
+     * @return Field bytes without tag or length, or {@code null} if the field is not found
+     *      in the input
+     * @throws IOException If an I/O error occurred
+     * @throws ParseException If there is a mismatch between the requested field and the field
+     *      in the input with the same field ID
+     */
+    @Nullable
+    public static Bytes extractRawFieldBytes(
```
anthony-swirldslabs, line 328: Minor: `extractFieldRawBytes` might be more precise, no? We're trying to extract raw bytes of a field, rather than extract bytes of a raw field?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -305,6 +308,61 @@ public static Bytes readBytes(final ReadableSequentialData input, final long max
...
 
+    /**
+     * Reads a requested length-delimited protobuf field from the input and returns it as a
+     * {@link Bytes} object. If the requested field is repeated or not length-delimited, this
+     * method throws an {@link IllegalArgumentException}. If the field is not found in the
+     * input, {@code null} is returned.
+     *
+     * <p>The returned Bytes object, if not null, will not contain the tag or the length.
+     *
+     * @param input The input to read from
```
anthony-swirldslabs, line 319: You may want to describe the constraints for the `input` as well. E.g. to state that it should represent a protobuf-encoded message, and/or it should start at a field boundary, etc. Also, it may be worth mentioning that the input stream must end where the object being parsed ends, i.e. the input stream must contain a single message only.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -305,6 +308,61 @@ public static Bytes readBytes(final ReadableSequentialData input, final long max
...
+     * @param field Field definition to extract bytes for
+     * @return Field bytes without tag or length, or {@code null} if the field is not found
+     *      in the input
```
anthony-swirldslabs, line 322: Personally, I like `null`. But we also had a "Bytes are never null" saga in the past. We might want to agree whether we do want to return `null` here, or we want to return the Bytes.EMPTY instead.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoParserToolsTest.java
```diff
@@ -342,4 +445,91 @@ private static <T> void testRead(
...
+        @Override
+        public boolean equals(Object obj) {
+            if (!(obj instanceof TestMessage other)) {
+                return false;
+            }
+            return Objects.equals(value, other.value);
+        }
+    }
+
+    private static final class TestMessageCodec implements Codec<TestMessage> {
```
anthony-swirldslabs, line 475: Minor: that's okayish for the test. But I'd prefer if that was an actual test .proto file with an actual generated codec. This is to account for any and all potential changes in the PBJ implementation, w/o having to remember that this hard-coded model and codec exist.

anthony-swirldslabs (dismissed):
LGTM functionally, but I left a few comments.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
artemananiev: Maybe just `extractFieldBytes()`?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
artemananiev: OK, I'll add this to the JavaDoc

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
anthony-swirldslabs: Works for me!

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
artemananiev: I don't think `Bytes.EMPTY` would be the right choice here. First, this trick to return `Bytes.EMPTY` instead of `null` is used at a higher level (codec level), while this new method is low-level utility method in `ProtoParserTools`. Second, there may actually be an empty field in the input (length == 0), this is not common, but possible. I would use `Bytes.EMPTY` in this case, but use `null` when the field is completely missing

anthony-swirldslabs: Sounds good to me.

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/ProtoParserToolsTest.java
artemananiev: Well, this is a unit test, not an integration test, right? There are no generated message / codec / schema classes in pbj-runtime currently, as far as I can see, and I don't think they should be added. The new API method works with any protobuf input, not necessarily produced by PBJ codecs. In fact, the test uses a codec only for the internal field, and only because our current api, `ProtoWriterTools.writeMessage()`, requires a codec, otherwise I would use low-level ProtoWriterTools / ProtoParserTools methods only. No need to overcomplicate the tests

anthony-swirldslabs: Makes sense. That looks good then. Thanks for the details.

tinker-michaelj (dismissed):
LGTM, tyvm @artemananiev !

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
artemananiev: Done

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
artemananiev: Renamed to `extractFieldBytes()`

anthony-swirldslabs (approved):
LGTM. Thank you!

anthony-swirldslabs:
> LGTM. Thank you!

Although, you may want to check tests in the PR checks.

imalygin (approved):

PR #412, opened 2025-03-06
Updating and expansion of README documentation for PBJ.

jasperpotts:
Cleanup docs, update where needed and add more.


README.md
```diff
@@ -1,91 +1,61 @@
...
+  * ### **PBJ Core** `pbj-core` which has 4 subprojects
+    * ### [**PBJ Protobuf Schema Compiler Gradle Plugin** `pbj-compiler`](pbj-core/pbj-compiler/README.md)
+    * ### [**PBJ Runtime Library** `pbj-runtime`](pbj-core/pbj-runtime/README.md)
+    * ### [**Grpc Helidon** `pbj-grpc-helidon`](pbj-core/pbj-grpc-helidon/README.md)
+    * ### [**Grpc Helidon Config** `pbj-grpc-helidon-config](pbj-core/pbj-grpc-helidon-config/README.md)
+  * ### [**Integration Tests** `pbj-integration-tests`](pbj-integration-tests/README.md) 
 
 ## Build Libraries
-Running `gradle build` in `pbj-core` directory will build the 2 libraries into the local maven repository for compiler and runtime.
+Running `gradle build` in `pbj-core` directory will build the libraries for compiler, runtime & grpc.
```
jsync-swirlds, line 44: Consistency
```suggestion
Running `gradle build` in `pbj-core` directory will build the libraries for compiler, runtime, and grpc.
```

README.md
```diff
@@ -1,91 +1,61 @@
...
 run all the 100k+ generated unit tests which takes a few minutes. The hedera protobufs provide a good coverage of
-protobuf features and insure all the generated code is tested for hedera node use case.
-
-**_These tests should be run before committing any new code for PBJ._**
\ No newline at end of file
+protobuf features and insure all the generated code is tested for hiero node use case.
+
+**_These tests should be run before committing any new code for PBJ._**
+
+## Long Term Goals
```
jsync-swirlds, line 54: Some of these are also described above. Should we unify the goals above, or possibly move those more detailed descriptions here?

pbj-core/pbj-compiler/README.md
```diff
@@ -0,0 +1,127 @@
...
+    main {
+        pbj {
+            srcDir(...)
+        }
+    }
+}
+```
+
+## Build Libraries
+Running `gradle build` in `pbj-core` directory will build the libraries for compiler, runtime and produce
```
jsync-swirlds, line 33: Consistency
```suggestion
Running `gradle build` in `pbj-core` directory will build the libraries for compiler and runtime to produce 
```

pbj-core/pbj-compiler/README.md
```diff
@@ -0,0 +1,127 @@
...
+}
+```
+
+## Build Libraries
+Running `gradle build` in `pbj-core` directory will build the libraries for compiler, runtime and produce 
+`hedera.pbj-compiler.gradle.plugin-VERSION.jar`
+
+## Records vs Classes
+The generated model objects are Java Classes, they used to be Java Records. This is a very tough decision and there are 
+many tread offs. The main reason for the change is that Java Records have no way to have lazy computed final fields. We
```
jsync-swirlds, line 38: Spelling
```suggestion
many tradeoffs. The main reason for the change is that Java Records have no way to have lazy computed final fields. We
```

pbj-core/pbj-compiler/README.md
```diff
@@ -0,0 +1,127 @@
...
+need to lazy compute and cache the hashCode() and protobufSize() values otherwise the performance impact is huge. In 
+hiero use cases not all model objects will have hashCode or protobufSize called on them so we need to avoid the cost of 
+computing if not needed. Also, it really hurts to compute in constructor and store in a final field as the thread that 
+constructs the model object is performance critical and these costs can be paid for on background threads. When we get 
+to Java versions with support for "[JEP 401: Value Classes and Objects](https://openjdk.org/jeps/401)" we will have to 
+review this decision as it maybe be better performance to compute in constructor so model objects can be value objects 
+and exist on stack rather than heap. It is also possible there will be a heuristic to choose which classes are one way 
+or the other, or we have to add an option in schema to choose on message by message basis.
```
jsync-swirlds, line 46: Grammar and spelling
```suggestion
need to lazy compute and cache the hashCode() and protobufSize() values, otherwise the performance impact is huge. In
hiero use cases, not all model objects will have hashCode or protobufSize called on them so we need to avoid the cost of
computing if not needed. Also, it is very costly to compute values in the constructor and store them in a final field as the thread that
constructs the model object is performance critical; these costs can be moved to background threads with lazy compute. When we get
to Java versions with support for "[JEP 401: Value Classes and Objects](https://openjdk.org/jeps/401)" we will need to
review this decision. It may be better performance to compute in the constructor then so that model objects can be value objects
and exist on stack rather than heap. It is also possible there will be a heuristic to choose which classes are processed one way
or the other. Alternatively, we may add an option in schema to choose the implementation on a message-by-message basis.
```

README.md
```diff
@@ -1,91 +1,61 @@
...
+ * **Modern Nice Java Objects** - parse the proto files into nice clean Java objects. With as clean as API as possible. 
+     Using the newer Java Record style getters and setters. Support latest LTS Java version features.
+ * **Explicit Error Handling** - when a message field is not on the wire, return null from getter not default object. 
+     This is like Java's choice of checked exceptions over unchecked exceptions. It forces the developer to handle 
+     the error case rather than silently ignoring it. To make this easier, the generated model objects have two versions 
+     of each getter method `foo()` and `fooOrElse(defaultValue)`.
  * **Performance Optimized** - be as fast or **faster** than the standard Google ProtoC generated code
  * **Minimal Garbage Generated** - produce the minimum amount of garbage Java objects as possible
+ * **Identical Binary Wire Encoding** - produce the same binary encoding as the standard Google ProtoC generated code
+ * **Deterministic Binary Encoding** - produce the same binary encoding for the same input every time. This means fields
```
anthony-swirldslabs, line 12: Taken together, these two points may contradict each other. For instance, Google ProtoC doesn't provide any guarantees for the map entries order while PBJ does. Therefore, the binary representation for these is almost certainly not identical. I'd suggest to rephrase, or expand on these two points, or perhaps combine them into a single one to avoid potential contradictions.

README.md
```diff
@@ -1,91 +1,61 @@
...
+  * ### [**Integration Tests** `pbj-integration-tests`](pbj-integration-tests/README.md) 
 
 ## Build Libraries
-Running `gradle build` in `pbj-core` directory will build the 2 libraries into the local maven repository for compiler and runtime.
+Running `gradle build` in `pbj-core` directory will build the libraries for compiler, runtime & grpc.
 
 ## Run Integration Tests
 Running `gradle build` in `pbj-integration-tests` will check out the latest proto source files from the
-[hedera-protobufs](https://github.com/hashgraph/hedera-protobufs) repository and generate code using `pbj-compiler` then 
+[hedera-protobufs](https://github.com/hashgraph/hedera-protobufs) repository as sample schemas and generate code using `pbj-compiler` then
```
anthony-swirldslabs, line 48: BTW, out of curiosity, what's our method for keeping models in `hedera-protobufs` in sync with what we have in `hiero-consensus-node` ? For example, to remove AddressBooks from PlatformState protobuf, I only modified the latter. Does the model in the former package even matter? If not, shouldn't we test against protobufs in `hiero-consensus-node` instead? (I'm not suggesting to change this in this fix, I'm just bringing up a discussion point.)

pbj-core/pbj-compiler/README.md
anthony-swirldslabs, line 38:
```suggestion
many trade-offs. The main reason for the change is that Java Records have no way to have lazy computed final fields. We
```

pbj-core/pbj-grpc-helidon-config/README.md
```diff
@@ -8,3 +8,4 @@ Helidon modules require a "config blueprint". An annotation processor takes that
 then needs to compile against the generated code. Since Gradle is not capable of
 doing this in a single build, we have to split the configuration into a separate
 module.
+` which produces `pbj-grpc-helidon-config-VERSION.jar`
```
anthony-swirldslabs, line 11:
```suggestion
module, which produces `pbj-grpc-helidon-config-VERSION.jar`.
```

pbj-core/pbj-runtime/README.md
```diff
@@ -0,0 +1,55 @@
...
+These are utility classes that provide some common functionality for the generated code. They also provide a low level 
+API for manually encoding and decoding protobuf messages.
+
+# PBJ Data IO Library `com.hedera.pbj.runtime.io`
+
+As well as protobuf code generation, PBJ also provides a data IO library that provides an alternative to the default
+java.io & java.nio libraries. The PBJ Data IO library has some key design decisions, of which some are alternatives
+to the standard Java IO libraries:
+* The starting point was to provide an API that could unify Streams and Buffers. So a single parser could work on
+  either a InputStream or a ByteBuffer. This was needed as we have both uses cases in the Hedera Node project and
```
anthony-swirldslabs, line 20:
```suggestion
  either a InputStream or a ByteBuffer. This was needed as we have both uses cases in the Hiero Consensus Node project and
```

pbj-integration-tests/README.md
```diff
@@ -0,0 +1,14 @@
+# PBJ Integration Tests
+The integration tests is a separate gradle project as it uses the `pbj-compiler` plugin to generate code from 
+test `.proto` files. The generated code is then compiled and tested as part of the project build. It uses the protobuf 
+schemas from the Hedera network as extensive examples to help insure correctness.
```
anthony-swirldslabs, line 4:
```suggestion
schemas from the Hiero Consensus network as extensive examples to help insure correctness.
```

or something like that.

pbj-integration-tests/README.md
```diff
@@ -0,0 +1,14 @@
...
+## Everything protobuf message
+For testing there is an `Everything` message that has every possible field type in it. This is used to test all the
+protobuf field types and insure they are correctly encoded and decoded. As support is added to PBJ for more Protobuf 
+features they should be added to this message.
+## JMH Benchmarks
+There are also JMH benchmarks that test the performance of the generated code. They also compare the performance of PBJ 
+generated code to the standard Google Protobuf generated code. The benchmarks are run by running the `jmh` gradle task.
+## Fuzz Testing
+There is a fuzz test that generates random protobuf messages and then serializes and deserializes them. It then compares
+the original message with the deserialized message to insure they are the same. This is run by running the `fuzzTest`
```
anthony-swirldslabs, line 14: You may want to also add that it tries to corrupt the serialized data and then parse it back, and ensures that parsing the corrupted data produces expected errors instead of crashing or silently swallowing them. This is actually the most essential part of our fuzz tests.

anthony-swirldslabs (commented):
Looks good overall. I left a few comments. Also, you may want to run spotless as it complains in PR checks.

jsync-swirlds (commented):
Thanks for updating these.
I noticed lots of little typos, grammar and spelling slips, and homonym swaps, and commented each with a suggestion to correct.

README.md
jasperpotts: I think @jsync-swirlds has been doing it manually but not sure we have a good process today yet. We need to have it automated at some point.

README.md
jasperpotts: We achieve both today but may not be able to in the future with new Protobuf versions that may not be possible. I know these design goes conflict. That is why it says under: "These design goals often compete with each other so this project tries to strike the right balance for use in the Hedera Node project."

README.md
jsync-swirlds: The release process for hiero-consensus-node includes a copy PR to update hedera-protobufs.
There is some work toward automating that.  At last check it is not complete.

README.md
anthony-swirldslabs: Sounds good.

README.md
anthony-swirldslabs: Sounds good. Thanks for confirming.

README.md
jasperpotts: Maybe opportunity for another PR. Feel like there is still loads that can be done to help docs. I just wanted to make sure it was at least more correct than it was. Also document a few design assumptions I had forgotten about before I forgot them again.

jsync-swirlds (approved):
Looks good to me.

anthony-swirldslabs (approved):
LGTM

jsync-swirlds (approved):

anthony-swirldslabs (approved):

Issue #263, opened 2024-06-05
PBJ promotes sub-type references to global scope.


jsync-swirlds:
PBJ compiler fails when it encounters a reference to a message contained in another message.  This makes it impossible to use intuitive message naming and well designed message scope.

## Example
File distro_response.proto
```proto
syntax = "proto3";

package com.zyxyl.distribution.partition;

message PartitionResponse {
    oneof response {
        FullResponse.Acknowledgement acknowledgement = 1;
        FullResponse.EndOfStream status = 2;
    }
}
message FullResponse {
    // several hundred lines elided here for clarity
    message Acknowledgement {
        oneof acknowledgements {
            FullResponse.ChildAcknowledgement child_ack = 271;
            FullResponse.ParentAcknowledgement parent_ack = 272;
        }
    }
    // several hundred lines elided here for clarity
    message ParentAcknowledgement {
        uint32 child_count = 872;
    }
    // several hundred lines elided here for clarity
    message ChildAcknowledgement {
        uint32 partition_count = 1438;
    }
    // several hundred lines elided here for clarity
    enum EndOfStream {
        UNKNOWN = 0;
        CANCELED = 1;
       // several hundred lines elided here for clarity
    }
}
```

The canonical protoc compiler compiles these and handles the references as expected (note, the file may contain several other messages with child messages named `Acknowledgement`, `EndOfStream`, etc...)
The PBJ compiler fails with an error similar to the following:
```text
Not found pbj package for message or enum [FullResponse.Acknowledgement] in file [protobufs/distribution/partition/distro_response.proto]
```

Worse, if there are two `EndOfStream` subtypes and both are used _only within their enclosing type_, PBJ will silently promote _one_ to global, and use the one type for all cases.  This causes the PBJ code to incorrectly create or interpret the messages (due to using the wrong subtype) and only fails at runtime (fortunately our extensive automated testing _does_ detect this).
This is a subtle data corruption bug, and it needs to be fixed.

jsync-swirlds:
PBJ also _promotes subtypes to global scope_.  This is incompatible with protoc, and is no longer something we can just ignore.  PBJ cannot compile the block node block service (nor many common protocol buffer/gRPC APIs!) without a fix for this bug.

jsync-swirlds:
Too be updated: Note here how this blocks implementing the "publish" protocol for Block Stream.

Nana-EC:
Waiting on @jasperpotts rewrite of pbj compiler.

[...]

PR #415, opened 2025-03-11
fix: handle partially-qualified inner message names

### LINKED ISSUE #414: Allow referencing inner messages
A sample .proto model in the parent ticket uses references to inner messages. PBJ is currently unable to compile this model unless all the references use fully-qualified names.

The ask is to allow using non-fully-qualified names when referencing inner messages.

anthony-swirldslabs:
**Description**:
Re-implementing fully qualified names look-up that now allows one to use partially qualified names to refer to inner and/or imported messages. This is the first part of a bigger issue described at https://github.com/hashgraph/pbj/issues/263.

**Related issue(s)**:

Fixes #414

**Notes for reviewer**:
A new test is added to verify the new behavior.

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
@@ -132,8 +132,7 @@ public final class LookupHelper {
         }
     }
 
-    private String getFullyQualifiedProtoMessageName(
-            final String messageType, final File protoSrcFile, final ParserRuleContext context) {
+    private String getFullyQualifiedProtoMessageName(final String messageType, final ParserRuleContext context) {
         // messageType can be a fully qualified name already, or it may be an unqualified, or partially qualified name.
         // Check all known fully qualified names, both local and imported:
         if (pbjPackageMap.containsKey(messageType)) {
@@ -151,8 +150,7 @@ public final class LookupHelper {
         while ((parentContext = parentContext.getParent()) != null) {
             if (parentContext instanceof MessageDefContext ctx) {
                 final String msgName = ctx.messageName().getText();
-                final String tryName =
-                        getFullyQualifiedProtoMessageName(msgName + "." + messageType, protoSrcFile, ctx);
+                final String tryName = getFullyQualifiedProtoMessageName(msgName + "." + messageType, ctx);
                 if (tryName != null) {
                     return tryName;
                 }
@@ -161,8 +159,10 @@ public final class LookupHelper {
                 // statements, so simply try each:
                 for (final Protobuf3Parser.PackageStatementContext psc : ctx.packageStatement()) {
                     final String packageName = psc.fullIdent().getText();
-                    final String prefix = (packageName == null || packageName.isBlank()) ? "" : (packageName + ".");
-                    final String tryName = getFullyQualifiedProtoMessageName(prefix + messageType, protoSrcFile, ctx);
+                    if (packageName == null || packageName.isBlank()) {
+                        continue;
+                    }
+                    final String tryName = getFullyQualifiedProtoMessageName(packageName + "." + messageType, ctx);
                     if (tryName != null) {
                         return tryName;
                     }
@@ -195,8 +195,7 @@ public final class LookupHelper {
                 return messageType;
             }
 
-            final String fullyQualifiedProtoMessageName =
-                    getFullyQualifiedProtoMessageName(messageType, protoSrcFile, context);
+            final String fullyQualifiedProtoMessageName = getFullyQualifiedProtoMessageName(messageType, context);
             if (fullyQualifiedProtoMessageName != null) {
                 return fullyQualifiedProtoMessageName;
             }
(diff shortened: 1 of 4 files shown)


jsync-swirlds (approved):
LGTM.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
```diff
@@ -132,6 +132,48 @@ public String getUnqualifiedProtoName(final ParserRuleContext context) {
         }
     }
 
+    private String getFullyQualifiedProtoMessageName(
+            final String messageType, final File protoSrcFile, final ParserRuleContext context) {
```
artemananiev, line 136: protoSrcFile seems redundant...

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
```diff
@@ -132,6 +132,48 @@ public String getUnqualifiedProtoName(final ParserRuleContext context) {
...
+        ParserRuleContext parentContext = context;
+        while ((parentContext = parentContext.getParent()) != null) {
+            if (parentContext instanceof MessageDefContext ctx) {
+                final String msgName = ctx.messageName().getText();
+                final String tryName =
+                        getFullyQualifiedProtoMessageName(msgName + "." + messageType, protoSrcFile, ctx);
+                if (tryName != null) {
+                    return tryName;
+                }
+            } else if (parentContext instanceof Protobuf3Parser.ProtoContext ctx) {
```
artemananiev, line 159: Can parentContext be anything else besides MessageDefContext or Protobuf3Parser.ProtoContext?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
```diff
@@ -132,6 +132,48 @@ public String getUnqualifiedProtoName(final ParserRuleContext context) {
...
+                // We've reached the top-level context. Weirdly, the grammar allows for multiple package
+                // statements, so simply try each:
+                for (final Protobuf3Parser.PackageStatementContext psc : ctx.packageStatement()) {
+                    final String packageName = psc.fullIdent().getText();
+                    final String prefix = (packageName == null || packageName.isBlank()) ? "" : (packageName + ".");
+                    final String tryName = getFullyQualifiedProtoMessageName(prefix + messageType, protoSrcFile, ctx);
```
artemananiev, line 165: I wonder if it makes sense to call this method recursively, if packageName is null/blank

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
anthony-swirldslabs: Good catch. Removing it.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
anthony-swirldslabs: Yes. I used a debugger to see the whole tree, and there's a lot of different contexts, like "one-of body", "message body", and half a dozens of others. However, as far as resolving names is concerned, only the `MessageDefContext` and `ProtoContext` seem to matter, at least in the sample .proto models that this fix is trying to support. So for now we simply skip all the others and keep walking upwards as they're irrelevant.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
anthony-swirldslabs: Probably not. Tests still seem to pass if I `continue` on a blank package name, so I'll skip that.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
jsync-swirlds: If enum has its own context, we might want to handle that here as well.

jsync-swirlds (approved):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/LookupHelper.java
anthony-swirldslabs: It may be unusual to define inner types inside an enum, at least in protobuf. Also, our grammar doesn't support this currently:

https://github.com/hashgraph/pbj/blob/d29a03004c897f96dcd84b1a580941e11bd465f1/pbj-core/pbj-compiler/src/main/antlr/com/hedera/hashgraph/protoparser/grammar/Protobuf3.g4#L159-L187

We can consider supporting this if/when we have a real use case for that, and after we confirm protobuf is supposed to support that. For now, I think we're already covering 99.(9)% of real-world use-cases.

Issue #416, opened 2025-03-12
Generate inner Java classes for inner Protobuf types


anthony-swirldslabs:
Generators in PBJ currently accept `File` objects referring to directories where to store main and test generated sources, and then create files and write generated code into them on their own.

In order to support generating inner classes, we'll need to introduce an abstraction "JavaFileGenerator" or similar that would expose an input stream for the actual generators as well as a facility to manage import statements.

Note that in reality we generate a model, a protobuf codec, a json codec, a schema, and a test - 5 files in total, each with its own set of imports. So perhaps we'll need an even higher-level abstraction, such as a "FileSetGenerator" that would essentially be a map of JavaFileGenerator objects for each generated file type.

The top-level loop at the PbjCompiler class would then be responsible for creating a new such FileSetGenerator object for each top-level element (message or enum) and would pass it to the generators.

When generating inner messages, the generators would pass its own FileSetGenerator to the nested generator calls, thus making them actual inner Java classes.

This requires a bit of a refactoring but should allow us to support inner classes generation. This may (or may not) require a few tricks to support nested unit tests as well.

An important detail is that this modification isn't 100% backward-compatible w.r.t. files organization on disk, and will require class references and/or imports statements fixing in client code.

anthony-swirldslabs:
All tasks are done. Resolving.

PR #418, opened 2025-03-12
feat: introduce JavaFileWriter and FileSetWriter

### LINKED ISSUE #417: Introduce JavaFileWriter and FileSetWriter

anthony-swirldslabs:
**Description**:
Under https://github.com/hashgraph/pbj/issues/416 , we want to start generating real inner classes for inner .proto sub-types.

Currently, PBJ generators accept `File` objects pointing to main/test source directories and create files for every model on their own. In order to support generating inner classes, we need to remove this responsibility from the generators to allow them to build their generated text continuously, including all the inner classes that they may need to generate. The responsibility of creating actual physical files on disk will be with the caller of the generators.

This PR introduces utilities that will help implement the above transfer of responsibilities in the future.

**Related issue(s)**:

Fixes #417

**Notes for reviewer**:
A new simple test is added to verify the file writer.

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/FileSetWriter.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/FileSetWriter.java
@@ -10,7 +10,7 @@ import java.io.IOException;
  * There's not a lot of logic around the set of JavaFileWriter objects, and for this reason they
  * are exposed as `public final` fields directly.
  */
-public class FileSetWriter {
+public final class FileSetWriter {
     public final JavaFileWriter modelWriter;
     public final JavaFileWriter schemaWriter;
     public final JavaFileWriter codecWriter;
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/FileSetWriter.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/FileSetWriter.java
@@ -4,34 +4,17 @@ package com.hedera.pbj.compiler.impl;
 import java.io.IOException;
 
 /**
- * This is essentially a set of JavaFileWriter objects that are necessary to support the generation
- * of a single top-level .proto model implementation in Java.
- *
- * There's not a lot of logic around the set of JavaFileWriter objects, and for this reason they
- * are exposed as `public final` fields directly.
+ * This is essentially a set of all the JavaFileWriter objects that are necessary to support the generation
+ * of a single top-level .proto model implementation in Java, such as a model writer, a codec writer, etc.
  */
-public final class FileSetWriter {
-    public final JavaFileWriter modelWriter;
-    public final JavaFileWriter schemaWriter;
-    public final JavaFileWriter codecWriter;
-    public final JavaFileWriter jsonCodecWriter;
-    public final JavaFileWriter testWriter;
+public record FileSetWriter(
+        JavaFileWriter modelWriter,
+        JavaFileWriter schemaWriter,
+        JavaFileWriter codecWriter,
+        JavaFileWriter jsonCodecWriter,
+        JavaFileWriter testWriter) {
 
-    /** Creates a new FileSetWriter object. */
-    public FileSetWriter(
-            final JavaFileWriter modelWriter,
-            final JavaFileWriter schemaWriter,
-            final JavaFileWriter codecWriter,
-            final JavaFileWriter jsonCodecWriter,
-            final JavaFileWriter testWriter) {
-        this.modelWriter = modelWriter;
-        this.schemaWriter = schemaWriter;
-        this.codecWriter = codecWriter;
-        this.jsonCodecWriter = jsonCodecWriter;
-        this.testWriter = testWriter;
-    }
-
-    /** A utility method to write all files at once. */
+    /** A utility method to write all the files at once. */
     public void writeFiles() throws IOException {
         modelWriter.writeFile();
         schemaWriter.writeFile();
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/FileSetWriter.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/FileSetWriter.java
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.pbj.compiler.impl;
+
+import java.io.IOException;
+
+/**
+ * This is essentially a set of JavaFileWriter objects that are necessary to support the generation
+ * of a single top-level .proto model implementation in Java.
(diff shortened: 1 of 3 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/JavaFileWriter.java
```diff
@@ -0,0 +1,91 @@
...
+     * although it will write all the latest updates to the imports and the text on each invocation.
+     * If the imports and/or text are updated and this `writeFile()` method is not invoked after the update,
+     * then the latest updates will not be persisted.
+     *
+     * @throws IOException if an I/O error occurs
+     */
+    public void writeFile() throws IOException {
+        try (Writer writer = new BufferedWriter(new FileWriter(javaFile))) {
+            // Hard-coding the license header for now. Can make it configurable in the future if need be.
+            writer.append("// SPDX-License-Identifier: Apache-2.0\n");
```
jsync-swirlds, line 81: It would probably help if this string was a constant.

jsync-swirlds (dismissed):
Looks good, one minor suggestion.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/FileSetWriter.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public final class FileSetWriter {
+    public final JavaFileWriter modelWriter;
+    public final JavaFileWriter schemaWriter;
+    public final JavaFileWriter codecWriter;
+    public final JavaFileWriter jsonCodecWriter;
+    public final JavaFileWriter testWriter;
+
+    /** Creates a new FileSetWriter object. */
+    public FileSetWriter(
```
artemananiev, line 21: Should it accept just a set of writers, i.e. `Set<JavaFileWriter>`?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/JavaFileWriter.java
```diff
@@ -0,0 +1,91 @@
...
+            writer.append("package ").append(javaPackage).append(";\n");
```
artemananiev, line 82: This would result in an incorrect Java source, so it makes sense to check that `javaPackage` is not null/blank.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/JavaFileWriter.java
```diff
@@ -0,0 +1,91 @@
...
+    private final String javaPackage;
+
+    /** All the imported symbols. */
+    private final Set<String> imports = new HashSet<>();
+
+    /** The actual text to be written, likely a top-level class or enum definition. */
+    private final StringBuilder stringBuilder = new StringBuilder();
+
+    /** Creates a new JavaFileWriter. */
+    public JavaFileWriter(final File javaFile, final String javaPackage) {
```
artemananiev, line 39: `@NonNull` for both args + `Objects.requireNonNull()`

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/JavaFileWriter.java
```diff
@@ -0,0 +1,91 @@
...
+     * @throws IOException if an I/O error occurs
+     */
+    public void writeFile() throws IOException {
+        try (Writer writer = new BufferedWriter(new FileWriter(javaFile))) {
+            // Hard-coding the license header for now. Can make it configurable in the future if need be.
+            writer.append("// SPDX-License-Identifier: Apache-2.0\n");
+            writer.append("package ").append(javaPackage).append(";\n");
+
+            if (!imports.isEmpty()) {
+                writer.append(imports.stream().sorted().collect(Collectors.joining(";\nimport ", "\nimport ", ";\n")));
```
artemananiev, line 85: This is correct, but not easy to parse. What about something like
```java
imports.stream().map(v -> "import " + v + ";\n").collect(Collectors.joining());
```
?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/JavaFileWriter.java
```diff
@@ -0,0 +1,91 @@
...
+            }
+
+            writer.append("\n").append(stringBuilder.toString()).append("\n");
```
artemananiev, line 88: Does it make sense to split it into three separate calls, so generated sources look more similar to this code? I mean, `writer.append("\n")` on its own line will make it very clear that a new line is added to the output.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/JavaFileWriter.java
anthony-swirldslabs: Added a constant.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/FileSetWriter.java
anthony-swirldslabs: No, because those are different writers (model, codec, etc.) I updated the javadoc to clarify, and also simplified this to be a record.

In theory we could get by using a `Map<Generator, JavaFileWriter>`. However, this wouldn't simplify the write-all-files operation (because of the IOException and need for sneaky-throw), and also, it would be an unnecessary complication because we don't expect the set of files to grow, and therefore we don't really need a map for this.

So it's just a small record with properly named fields.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/JavaFileWriter.java
anthony-swirldslabs: This is now being checked in the constructor.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/JavaFileWriter.java
anthony-swirldslabs: I'm not a fan of this boilerplate, but okay.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/JavaFileWriter.java
anthony-swirldslabs: I just took what I found in the existing model generators in PBJ. But I simplified this now per your suggestion.

artemananiev (approved):

PR #425, opened 2025-03-13
refactor: use JavaFileWriter in EnumGenerator

### LINKED ISSUE #419: Use JavaFileWriter in EnumGenerator

anthony-swirldslabs:
**Description**:
This is the very beginning of us using the new `JavaFileWriter` class. This PR simply refactors the `EnumGenerator` to use this class instead of creating a physical file on its own.

**Related issue(s)**:

Fixes #419

**Notes for reviewer**:
All tests must continue passing. This fix is not expected to change anything in the generated code per se.

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/EnumGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/EnumGenerator.java
@@ -6,9 +6,8 @@ import static com.hedera.pbj.compiler.impl.Common.DEFAULT_INDENT;
 
 import com.hedera.pbj.compiler.impl.ContextualLookupHelper;
 import com.hedera.pbj.compiler.impl.FileType;
+import com.hedera.pbj.compiler.impl.JavaFileWriter;
 import com.hedera.pbj.compiler.impl.grammar.Protobuf3Parser;
-import java.io.File;
-import java.io.FileWriter;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


jsync-swirlds (dismissed):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/EnumGenerator.java
```diff
@@ -29,12 +28,14 @@ record EnumValue(String name, boolean deprecated, String javaDoc) {}
...
      */
-    public static void generateEnumFile(
-            Protobuf3Parser.EnumDefContext enumDef, File destinationSrcDir, final ContextualLookupHelper lookupHelper)
+    public static void generateEnum(
+            final Protobuf3Parser.EnumDefContext enumDef,
+            final JavaFileWriter writer,
+            final ContextualLookupHelper lookupHelper)
             throws IOException {
         final String enumName = enumDef.enumName().getText();
         final String modelPackage = lookupHelper.getPackageForEnum(FileType.MODEL, enumDef);
```
artemananiev, line 41: This local var is no longer used

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/EnumGenerator.java
```diff
@@ -29,12 +28,14 @@ record EnumValue(String name, boolean deprecated, String javaDoc) {}
      * Generate a Java enum from protobuf enum
      *
      * @param enumDef the parsed enum def
-     * @param destinationSrcDir The destination source directory to generate into
+     * @param writer the writer to append the generated enum to
```
artemananiev, line 31: Does it make sense to add a doc that the generated enum sources should not contain a Java package statetement?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/EnumGenerator.java
anthony-swirldslabs: Good catch. Removed.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/EnumGenerator.java
anthony-swirldslabs: No. This is already documented in the javadoc for the JavaFileWriter, so it's implied. There's not a need to duplicate this fact in every generator, especially now when we're removing this responsibility from them.

artemananiev (approved):

jsync-swirlds (approved):

Issue #389, opened 2025-02-07
Bytes.hashCode() can be optimized


artemananiev:
`Bytes.hashCode()` is implemented in a sub-optimal way:

```java
    @Override
    public int hashCode() {
        int h = 1;
        for (long i = length() - 1; i >= 0; i--) {
            h = 31 * h + getByte(i);
        }
        return h;
    }
```

Instead, we can use `Arrays.hashCode()`, which leverages vector ops support in modern CPUs.

There is one issue, though. The existing `hashCode()` may not be changed, not yet. If it's changed, hash code values will be changed, too. If a hash code is persisted anywhere (e.g. in MerkleDb), this place will be broken. We've seen it in the past: a key was stored on disk with one hash code, then hashCode() implementation was changed, and the key couldn't be retrieved from disk any longer.

So I propose:
* Introduce a separate method in `Bytes`, for example, `fastHashCode()`. This is what this ticket is about
* When we switch to Virtual Mega Map, use this method instead of `hashCode()` to store keys/values on disk. Migration to the mega map doesn't have to be backwards compatible, all data is migrated anyway
* Some time later, `hashCode()` can be changed the same way as `fastHashCode()`, and the latter can be removed. All usages of `fastHashCode()` is the main repo will be replaced with `hashCode()`. Since implementation is the same, it will be backwards compatible
* In-memory usages of `hashCode()`, e.g. when an object is stored in a Java collection, should be fine, since they are not persisted across node runs (no traces in state snapshots)

artemananiev:
Some interesting facts:

* `Arrays.hashCode()` only works with full arrays, there is no way to hash a part of an array (offset + length). It makes this method unusable for sliced `Bytes` objects (created using `Bytes.slice()`)

* `Arrays.hashCode()` value is not equal to existing `Bytes.hashCode()`

* `Arrays.hashCode()` calls `ArraysSupport.vectorizedHashCode()`, which is annotated with `IntrinsicCandidate`. It means, this method may or may not be optimized by VM

* Optimized and unoptimized versions of `ArraysSupport.IntrinsicCandidate()` produce different values. We need stable and predictable `hashCode()` implementation for `Bytes`, so `Arrays.hashCode()` doesn't look like a viable option

And some numbers:

* On mac, `Arrays.hashCode()` is 50-60% faster than existing `Bytes.hashCode()`
[...]

PR #426, opened 2025-03-13
Try faster hashCode

### LINKED ISSUE #424: Bytes.hashCode() performance
### Description

Bytes.hashCode shows up as the biggest number of loop iterations in code base.
Image

### Steps to reproduce

measure performance of Bytes.hashCode()

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.9.20

### Operating system

None

jasperpotts:
@anthony-swirldslabs : A second revision of the fix on 3/31/2025. This time we always use our own hash code computation code and never rely on JDK directly. Also, the benchmark has been modified to actually test the `hashCode` method specifically. Results for small (16), medium (16K), and large (16M) Bytes objects:

**Before the fix:**
```
Benchmark                            Mode  Cnt          Score          Error  Units
BytesHashCodeBench.hashLargeBytes   thrpt    5         62.785 ±        2.616  ops/s
BytesHashCodeBench.hashMediumBytes  thrpt    5      64916.981 ±     1178.676  ops/s
  BytesHashCodeBench.hashSmallBytes   thrpt    5  156480715.770 ± 10117792.960  ops/s
```

**With the fix:**
```
Benchmark                            Mode  Cnt            Score           Error  Units
BytesHashCodeBench.hashLargeBytes   thrpt    5  17794927753.539 ± 120647815.286  ops/s
BytesHashCodeBench.hashMediumBytes  thrpt    5  17793757339.364 ± 137804766.981  ops/s
BytesHashCodeBench.hashSmallBytes   thrpt    5  17711207197.438 ± 541245113.180  ops/s
```

A 100x improvement even for a small Bytes of size 16. A much larger improvement for larger Bytes objects.

-------------------------
Below is the description from the original revision of the fix from @jasperpotts :
-------------------------

Aims to fix #424 , bellow are results on mac laptop.

hashCodeResults


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -466,11 +471,20 @@ public boolean equals(@Nullable final Object o) {
...
     @Override
     public int hashCode() {
-        int h = 1;
-        for (int i = start + length - 1; i >= start; i--) {
-            h = 31 * h + UnsafeUtils.getArrayByteNoChecks(buffer, i);
+        if (hashCode == 0 && length > 0) {
+            if (length == 1) {
+                hashCode = 31 + (int) buffer[start];
+            } else if (start == 0 && length == buffer.length) {
+                hashCode = Arrays.hashCode(buffer);
```
anthony-swirldslabs, line 478: I'm concerned about using JDK-provided methods for hashing. We've already encountered a case where the Java hash code could be stored in state. That code obviously breaks whenever the JDK changes its internal implementation. Further, some implementations may even return a different hash code value between JVM runs.

Could you please benchmark the cached version using the original loop only and see if the results are dramatically worse than using the `Arrays.hashCode`?

artemananiev: I second to that. Using `Arrays.hashCode()` was discussed in https://github.com/hashgraph/pbj/issues/389, and a decision was made to not use it, basically for the reasons outlined above by Anthony. Backwards compatibility with current implementation can be solved (e.g. switching to this new PBJ version in mega map with full data migration). However, the concern about future Java versions cannot be addressed

imalygin (dismissed):

imalygin (approved):

alex-kuzmin-hg (approved):
LGTM. I'll perf-test after merge

OlegMazurov (approved):

PR #428, opened 2025-03-14
refactor: use JavaFileWriter in message generators

### LINKED ISSUE #420: Use JavaFileWriter in message generators

anthony-swirldslabs:
**Description**:
In this fix we use the new `JavaFileWriter` in all the remaining message generator classes. With this fix, all the generators are no longer responsible for creating physical files for its own code that they generate - they simply emit all the generated data into the supplied writer, and the caller of the generator is responsible for calling `writeFile` on the writer to actually write a file on disk.

Please note that inner types are still written into separate files/classes. In general, as far as the actual generated code is concerned, it should be the same as before. This fix simply replaces the core mechanism of creating files on disk with using the new utilities that we've introduced recently. A switch to producing actual inner classes will happen in a separate fix in the future.

**Related issue(s)**:

Fixes #420 

**Notes for reviewer**:
All tests should continue to pass as before.

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecGenerator.java
@@ -3,23 +3,19 @@ package com.hedera.pbj.compiler.impl.generators.protobuf;
 
 import static com.hedera.pbj.compiler.impl.generators.protobuf.CodecDefaultInstanceMethodGenerator.generateGetDefaultInstanceMethod;
 
-import com.hedera.pbj.compiler.impl.Common;
 import com.hedera.pbj.compiler.impl.ContextualLookupHelper;
 import com.hedera.pbj.compiler.impl.Field;
 import com.hedera.pbj.compiler.impl.FileType;
+import com.hedera.pbj.compiler.impl.JavaFileWriter;
 import com.hedera.pbj.compiler.impl.MapField;
 import com.hedera.pbj.compiler.impl.OneOfField;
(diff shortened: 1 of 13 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecGenerator.java
```diff
@@ -32,101 +28,92 @@ public final class CodecGenerator implements Generator {
...
+                .replace("$codecClass", codecClassName)
+                .replace("$unsetOneOfConstants", CodecParseMethodGenerator.generateUnsetOneOfConstants(fields))
+                .replace("$parseMethod", CodecParseMethodGenerator.generateParseMethod(modelClassName, fields))
+                .replace("$writeMethod", writeMethod)
+                .replace("$measureDataMethod", CodecMeasureDataMethodGenerator.generateMeasureMethod(modelClassName, fields))
+                .replace("$measureRecordMethod", CodecMeasureRecordMethodGenerator.generateMeasureMethod(modelClassName, fields))
+                .replace("$fastEqualsMethod", CodecFastEqualsMethodGenerator.generateFastEqualsMethod(modelClassName, fields))
+                .replace("$getDefaultInstanceMethod", generateGetDefaultInstanceMethod(modelClassName))
+        );
         // spotless:on
```
artemananiev, line 117: `// spotless:off` is removed, so I assume this `// spotless:on` should be as well

anthony-swirldslabs: Nope. It's still up there at line 84 above.

PR #442, opened 2025-03-25
feat: add javaPackageSuffix gradle plugin parameter

### LINKED ISSUE #435: Add PBJ Gradle plugin parameter to enable adding package name suffix

anthony-swirldslabs:
**Description**:
Introducing a new PBJ Gradle plugin parameter to specify a Java package suffix that PBJ compiler will add to generated classes if the input protobuf models don't specify the `pbj.java_package` option explicitly. This allows the PBJ integration tests to compile such models using both the PBJ compiler and Google protoc and not have a clash between the generated models sharing the same Java package - with this additional parameter specified, the PBJ models will be placed in a differently named Java package.

I'm specifying this additional suffix in the integration tests' `build.gradle.kts` and adding a test model that would otherwise fail the compilation of the PBJ integration tests today. But thanks to the new parameter, the compilation succeeds and so do all the PBJ integration tests.

I'm also updating the snapshot version to 0.10.1 and will release it once the PR is merged.

So we can now easily import 3rd-party protobuf models in the Consensus Node software, and PBJ integration tests would still be able to compile them all and hopefully succeed each time.

**Related issue(s)**:

Fixes #435

**Notes for reviewer**:
A new model is added to the integration tests `main` sources to verify the new behavior. If the `javaPackageSuffix` isn't specified in `build.gradle.kts` or the entire `pbj` section is missing there, then the build with this new model fails with:
```
Execution failed for task ':compileJava'.
> Compilation failed; see the compiler output below.
  /Users/anthony/ws/53-pbjInner/pbj/pbj-integration-tests/build/generated/source/pbj-proto/main/java/com/hedera/pbj/test/proto/MessageInAPackage.java:22: error: duplicate class: com.hedera.pbj.test.proto.MessageInAPackage
  public final class MessageInAPackage {
               ^
  10 errors
```

**Checklist**


pbj-integration-tests/build.gradle.kts
```diff
@@ -61,6 +61,9 @@ dependencyAnalysis { issues { all { onAny { exclude("com.hedera.pbj:pbj-compiler
 // IMPROVE: JMH code should not depend on test code
 jmh { includeTests = true }
 
+// Avoid a clash with Google protoc models when .ptoto files don't specify `pbj.java_package`:
```
jsync-swirlds, line 64:
```suggestion
// Avoid a clash with Google protoc models when .proto files don't specify `pbj.java_package`:
```

jsync-swirlds (dismissed):
Looks good, one minor typo.

jsync-swirlds (approved):

imalygin (approved):

PR #446, opened 2025-03-26
feat: support inner enums

### LINKED ISSUE #445: Enable enum to be nested within a message.
## Problem

The Protobuf spec and ProtoC compiler both permit an `enum` as a nested type within a message.
Because the ProtoC compiler uses C++ scope rules for enum values (the individual enums are "siblings" of the enum type, not "children"), this is the only mechanism to reuse enumerated names within the same _package_.

PBJ, however, does not support an `enum` as a nested type. This limitation has led to a habit of naming every enum in HAPI with a usage prefix, even if the enum serves the same purpose across services or within the same service (e.g. TokenFreezeStatus, TokenKycStatus, and TokenPauseStatus).
The lack of nested `enum` definitions has also led to including within many enum value names a reference to the enum type (e.g. ResponseCodeEnum.UNKNOWN vs. FreezeType.UNKNOWN_FREEZE_TYPE, or HashAlgorithm.HASH_ALGORITHM_UNKNOWN) because only one enum in all of the HAPI can use the simpler name (partly because HAPI uses the package `proto` for all but some of the very newest definitions).

## Solution

We would like to have support for defining an `enum` as a nested message within a `message` type (itself possibly nested).

The requirement for the enum value names to be parsed at the same scope as the enclosing `enum` is a compiler requirement in ProtoC, and need not _necessarily_ be imposed by PBJ. In hiero-consensus-node, the integration tests will, in effect, impose that requirement by compiling the proto with ProtoC, which should be sufficient.

If that scope requirement is to be enabled for PBJ, it would probably need to be applied as a compiler function when verifying the parsed content.

The following file (with all comments removed) is a realistic test file (it is how we _wanted to_ design the Block Node API file initially, before this limitation was encountered).

```protobuf
syntax = "proto3";

package com.hedera.hapi.block.node;

// SPDX-License-Identifier: Apache-2.0

option java_multiple_files = true;

import "services/basic_types.proto";
import "block/stream/block.proto";
import "block/stream/block_item.proto";

message PublishStreamRequest {
    oneof request {
        BlockItemSet block_items = 1;
    }
    enum StreamEndCode {
        UNKNOWN = 0;
        RESET = 1;
        TIMEOUT = 2;
        ERROR = 3;
        TOO_FAR_BEHIND = 4;
    }
}

message BlockItemSet {
    repeated com.hedera.hapi.block.stream.BlockItem block_items = 1;
}

message PublishStreamResponse {
    oneof response {
        Acknowledgement acknowledgement = 1;
        EndOfStream end_stream = 2;
        SkipBlock skip_block = 3;
        ResendBlock resend_block = 4;
    }
    message Acknowledgement {
        BlockAcknowledgement block_ack = 1;
    }
    message BlockAcknowledgement {
        uint64 block_number = 1;
        bytes block_root_hash = 2;
        bool block_already_exists = 3;
    }
    message SkipBlock {
        uint64 block_number = 2;
    }
    message ResendBlock {
        uint64 block_number = 2;
    }
    message EndOfStream {
        ResponseCode status = 1;
        uint64 block_number = 2;
        enum ResponseCode {
            UNKNOWN = 0;
            SUCCESS = 1;
            TIMEOUT = 2;
            OUT_OF_ORDER = 3;
            BAD_STATE_PROOF = 4;
            BEHIND = 5;
            INTERNAL_ERROR = 6;
            PERSISTENCE_FAILED = 7;
        }
    }
}

message SingleBlockRequest {
    uint64 block_number = 1;
    bool allow_unverified = 2;
    bool retrieve_latest = 3;
}

message SingleBlockResponse {
    ResponseCode status = 1;
    com.hedera.hapi.block.stream.Block block = 2;
    enum ResponseCode {
        UNKNOWN = 0;
        INSUFFICIENT_BALANCE = 1;
        SUCCESS = 2;
        NOT_FOUND = 3;
        NOT_AVAILABLE = 4;
        ALLOW_UNVERIFIED_REQUIRED = 5;
        VERIFIED_BLOCK_UNAVAILABLE = 6;
    }
}

message SubscribeStreamRequest {
    uint64 start_block_number = 1;
    uint64 end_block_number = 2;
    bool allow_unverified = 3;
}

message SubscribeStreamResponse {
    oneof response {
        ResponseCode status = 1;
        BlockItemSet block_items = 2;
    }
    enum ResponseCode {
        UNKNOWN = 0;
        INSUFFICIENT_BALANCE = 1;
        SUCCESS = 2;
        INVALID_START_BLOCK_NUMBER = 3;
        INVALID_END_BLOCK_NUMBER = 4;
        NOT_AVAILABLE = 5;
    }
}

message StateSnapshotRequest {
    uint64 last_block_number = 2;
    bool retrieve_latest = 3;
}

message StateSnapshotResponse {
    ResponseCode status = 1;
    uint64 last_block_number = 2;
    string snapshot_reference = 3;
    enum ResponseCode {
        UNKNOWN = 0;
        INSUFFICIENT_BALANCE = 1;
        SUCCESS = 2;
    }
}

message ServerStatusRequest {}

message ServerStatusResponse {
    uint64 first_available_block = 1;
    uint64 last_available_block = 2;
    bool only_latest_state = 3;
    BlockNodeVersions version_information = 4;
}

message BlockNodeVersions {
    proto.SemanticVersion address_book_version = 1;
    proto.SemanticVersion stream_proto_version = 2;
    proto.SemanticVersion software_version = 3;
}

service BlockNodeService {
    rpc serverStatus(ServerStatusRequest) returns (ServerStatusResponse);
}

service BlockAccessService {
    rpc singleBlock(SingleBlockRequest) returns (SingleBlockResponse);
}

service StateService {
    rpc stateSnapshot(StateSnapshotRequest) returns (StateSnapshotResponse);
}

service BlockStreamService {
    rpc publishBlockStream (stream PublishStreamRequest) returns (stream PublishStreamResponse);
    rpc subscribeBlockStream(SubscribeStreamRequest) returns (stream SubscribeStreamResponse);
}

```

## Alternatives

Leave the status as is and fail to correctly parse any proto file that uses an `enum` nested within a `message`.

anthony-swirldslabs:
**Description**:
Introducing support for inner enum definitions in PBJ which are now properly generated and usable.

Also, fixing a tiny bug in `TestGenerator.java` related to inner classes where their tests wouldn't compile if the inner class was never used as a real field type in a proto model.

**Related issue(s)**:

Fixes #445 

**Notes for reviewer**:
* A new model is added to PBJ integration tests that verifies various nested enums
* A model definition from the description of #445 can also be compiled and tested via the PBJ integration tests (and all tests pass), although I'm not including it in the fix since it's a bit more complex than what we want to have in tests.

**Checklist**


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/ContextualLookupHelper.java
```diff
@@ -99,14 +99,20 @@ public String getPackageFieldMessageType(final FileType fileType, final FieldCon
...
      */
-    public String getPackageFieldMessageType(final FileType fileType, final Type_Context typeContext) {
-        return lookupHelper.getPackage(srcProtoFileContext, fileType, typeContext.messageType());
+    public String getPackageFieldType(final FileType fileType, final Type_Context typeContext) {
+        if (typeContext.messageType() != null) {
+            return lookupHelper.getPackage(srcProtoFileContext, fileType, typeContext.messageType());
+        } else if (typeContext.enumType() != null) {
+            return lookupHelper.getPackage(srcProtoFileContext, fileType, typeContext.enumType());
+        } else {
+            return null;
```
imalygin, line 114: Is null a valid value here? Shouldn't we throw an `UnsupportedOperationException`?

anthony-swirldslabs: Yes, `null` is valid here. This particular method is only used in once place at 

https://github.com/hashgraph/pbj/blob/573939697dc2986dcdf3dceae1e99d9f33b90d2d/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Field.java#L238-L238

and as you can see in there, the `Field.extractTypePackage()` itself may return `null` as well. So anyone who uses the return value is prepared to handle nulls already.

imalygin (approved):

PR #467, opened 2025-04-15
perf: use $hashCode in equals()

### LINKED ISSUE #233: Caching of Hashcode in generated objects
### Problem

Computing the hashcode over and over is a performance problem. Specifically with the extra shifts to randomize that is done for better bucket distribution when objects are used as map keys.

### Solution

Move all PBJ model objects from Java Record to final Java Classes. Implement code generation for constructors, fields, getter methods and toString() method. Then add a private hashCode field that is computed at construction time and caches the hashCode as it is constant as objects are immutable. We should also use the new computed cached hashCode in the equals() method to fail fast if the hashCodes do not match.

### Alternatives

_No response_

anthony-swirldslabs:
**Description**:
Optimize generated models `equals()` method by checking the precomputed `$hashCode` before comparing all the fields individually.

**Related issue(s)**:

Fixes #233 

**Notes for reviewer**:
All test should pass.

**Checklist**


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -433,6 +433,9 @@ public boolean equals(Object that) {
                 return false;
             }
             $javaRecordName thatObj = ($javaRecordName)that;
+            if ($hashCode != -1 && thatObj.$hashCode != -1 && $hashCode != thatObj.$hashCode) {
```
imalygin, line 436: just for my understanding. where does this `-1` come from?

anthony-swirldslabs: This comes from the initial value of the internal private mutable `$hashCode` variable which means that it hasn't been precomputed yet. Here's how the `hashCode()` looks like:

```
    @Override
    public int hashCode() {
        // The $hashCode field is subject to a benign data race, making it crucial to ensure that any
        // observable result of the calculation in this method stays correct under any possible read of this
        // field. Necessary restrictions to allow this to be correct without explicit memory fences or similar
        // concurrency primitives is that we can ever only write to this field for a given Model object
        // instance, and that the computation is idempotent and derived from immutable state.
        // This is the same trick used in java.lang.String.hashCode() to avoid synchronization.
    
        if($hashCode == -1) {
```

and it then goes on and precomputes this value and stores the precomputed value in the `$hashCode` variable so that the next time it can return it immediately.

anthony-swirldslabs: And this is how this field is defined in every model:
```
    /** Computed hash code, manual input ignored. */
    private int $hashCode = -1;
```

artemananiev (approved):

PR #471, opened 2025-04-17
feat: support unknown fields

### LINKED ISSUE #456: Support unknown fields
### Problem

The mirror node needs to be both backwards and future compatible with all versions of the record and block stream protobufs. If any field is removed, we still need to parse historical versions of it. If any new transaction type is added but we have not updated the protobuf version to add support, we still store generic transaction info without failing.

### Solution

Add optional support for unknown fields when using `parse()`. In protoc, this takes the form of `Map<Integer, Field> fields`. `parseStrict()` should remain unchanged and throw an error if unknown fields are present.

### Alternatives

_No response_

anthony-swirldslabs:
**Description**:
Introducing support for unknown fields. The default behavior is unchanged and PBJ skips unknown fields by default. To activate the new behavior, the `parseUnknownFields` argument of the `Codec.parse()` method needs to be set to `true`. This enables the parser to collect all the known fields in a map in the model class. Later on, the codec can use this map to serialize all the unknown fields if they're present. Also, applications are able to examine the unknown fields by calling `model.getUnknownFields()` to get an unmodifiable map with the fields.

**Revision 2:**
* Addressing comments - mainly replacing the map with a list, adding docs, and an extra round-trip test.
* Fixing minor bugs: in `protobufSize()` to account for unknown fields, and in `copyBuilder()` to account for them as well.

**Revision 3:**
* Adding unknown fields support to equals/hashCode/toString/compareTo
* Sorting the fields when parsing to ensure determinism of the above, as well as of the write()
* Optimizing the initial size of the unknown fields array to help speed up parsing
* Adding a comprehensive round-trip test that uses the Everything model and hence verifies the round trip for all the types that PBJ supports.
**Also note:** in Google Protobuf, the JSON format officially doesn't support unknown fields. So PBJ won't support them in its JSONCodec as well. See https://protobuf.dev/programming-guides/proto3/#retaining

**Revision 4:**
* Updated the javadoc for all public methods in the `UnknownField` record.

**Related issue(s)**:

Fixes #456 

**Notes for reviewer**:
A new integ test is added.

**Checklist**


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -189,14 +192,36 @@ public void generate(
...
+                /**
+                 * Get an unmodifiable map from protobuf field numbers to UnknownField records for all unknown fields parsed
+                 * from the original data, i.e. the fields that are unknown to the .proto model which generated
+                 * this Java model class.
+                 * Note that by default, PBJ Codec discards unknown fields. The parse() method has to be invoked
+                 * with `parseUnknownFields = true` in order to populate this map.
+                 * @return a (potentially empty) map of unknown fields
+                 */
+                public @NonNull Map<Integer, UnknownField> getUnknownFields() {
+                    return $unknownFields == null ? Collections.EMPTY_MAP : $unknownFields;
```
artemananiev, line 220: I wonder if somebody will need to distinguish between "no unknown fields" and "unknown field parsing was not requested"? For example, if this method is called, but unknown fields aren't parsed, it may throw an illegal state exception

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ReadableSequentialData.java
```diff
@@ -485,6 +485,26 @@ default long readVarLong(final boolean zigZag) throws BufferUnderflowException,
...
+    /**
+     * Read the bytes of a 64bit protobuf varint at current {@link #position()}.
+     *
+     * @return bytes of the long read in var int format
+     * @throws BufferUnderflowException If the end of the sequence is reached before the final variable byte fragment
+     *                                  is read
+     * @throws UncheckedIOException if an I/O error occurs
+     * @throws DataEncodingException if the variable long cannot be decoded
+     */
+    default Bytes readVarBytes() throws BufferUnderflowException, UncheckedIOException {
```
artemananiev, line 497: The name is slightly confusing. What about `readVarIntBytes()`? Or `readVarLongBytes()`

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/UnknownField.java
```diff
@@ -0,0 +1,12 @@
...
+package com.hedera.pbj.runtime;
+
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import java.util.List;
+
+/**
+ * A record representing an unknown field.
+ * @param wireType the wire type of the field (e.g. varint, or delimited, etc.)
+ * @param bytes a list of the raw bytes of each occurrence of the field (e.g. for repeated fields)
+ */
```
artemananiev, line 11: Please, add a doc explaining how length-delimited fields are stored here (the Bytes objects contain both length and content)

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -145,9 +152,20 @@ static String generateParseLoop(
...
-                                            // We just need to read off the bytes for this field to skip it
-                                            // and move on to the next one.
-                                            skipField(input, ProtoConstants.get(wireType), $skipMaxSize);
+                                            if (parseUnknownFields) {
+                                                if ($unknownFields == null) {
+                                                    $unknownFields = new HashMap<>();
+                                                }
+                                                $unknownFields.merge(
+                                                    field,
+                                                    new UnknownField(ProtoConstants.get(wireType), List.of(extractField(input, ProtoConstants.get(wireType), $skipMaxSize))),
```
artemananiev, line 161: I wonder if code would be easier, if UnknownField contained a tag (field number + wire type) and a single Bytes object for the value, and models contained a list <UnknownField> rather than a map <field number, UnknownField>?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
jasperpotts: I question best way to store. The options that come to mind are: the way you have done it or you could have an array of `UnknownField` objects and add field number integer as a field in `UnknownField`. The advantage of that is lighter weight memory wise at runtime at the cost of more complicated access to unknown fields.  I think this question leads into the bigger question of the main use case of this feature. I think there are two possible use cases and they have differing design optimization criteria. 

1. For round trip retention of data. This is the block node use case. It wants to be able to parse objects for new schema than it has so they may have new fields. Then be able to serialize/write them back out without any data loss. This could include adding additional fields to a oneof. 
2. Code that wants to access fields it doesn't know. For this use case we want nice API to access field as various types, getAsLong() or getAsString() or something similar, or way to use with ProtobufTools class to have nice way to read including fields of type message. We do not have a customer for this yet but I would not be surprised for it to come up. 

For case (1) we want to minimize performance and memory overhead of extra data passing though without it ever being read other than when serializing back out in the end. For (2) then a map to access fields by field ID is nice to have. 

I know JDK did some optimizations for small Maps, so maybe they are not that expensive any more. In terms of number of objects created and RAM bytes used.

anthony-swirldslabs: I'm unaware of such a requirement.

The internal code (aka the Codec.write()) must be able to work successfully either way. So it would have to deal with exceptions or null's regardless of anything.

The external code (aka users of the protobuf model) are supposed to be aware how they've called the Codec.parse() previously because if their business logic needs to deal with unknown fields, then they know that they must call parse() passing `true` for `parseUnknownFields`. Therefore, they always know what to expect from `getUnknownFields()` and hence they don't need additional indications such as `null` or exceptions.

For this reason, unless we hear of such requirements, the API could be nice as it is in this fix - it never returns `null` or throws any exceptions, thus simplifying the client code that wants to use this API.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ReadableSequentialData.java
anthony-swirldslabs: Good catch. `readVarLongBytes` seems to be perfect here, so I'm renaming it accordingly.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -566,6 +593,12 @@ private static String generateConstructor(
                 .replace("$constructorParams",fieldsNoPrecomputed.stream().map(field ->
                         field.javaFieldType() + " " + field.nameCamelFirstLower()
                 ).collect(Collectors.joining(", ")))
+                .replace("$unknownFieldsParam", initUnknownFields
+                        ? ((fieldsNoPrecomputed.isEmpty() ? "" : ", ") + "final Map<Integer, UnknownField> $unknownFields")
+                        : "")
+                .replace("$unknownFieldsCode", "this.$unknownFields = " + (initUnknownFields
+                        ? "$unknownFields == null ? null : Collections.unmodifiableMap($unknownFields)"
```
jasperpotts, line 600: I wonder if there is a way to use `Map.of()` as it is immutable by default and uses an efficient array based implementation. Like what happens if the constructor throws a error if you do not pass an immutable map. Then parser uses Map.of() maybe.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
jasperpotts: This is the same trade-off debate I mentioned in earlier comment. Working out optimal way to store data. I guess most critical thing is how we expose in public API on generated model object. As if that is simple enough then we have flexibility to implement different ways over time as we optimize for different use cases. Current design here is `Map<Integer, UnknownField> getUnknownFields()` which leads to a map based API. Maybe a option that leaves more doors open is:

`UnknownField unknownField(int fieldID)`
`Stream<UnknownField> unknownFieldStream()` where UnknownField also contains field ID

With that we could use map or array based internal storage and work on finding optimal over time.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
jasperpotts: Another point is we are using record style method names with out the `get` and `set` so should be `unknownFields()`

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -430,13 +448,14 @@ static String readMethod(Field field) {
...
+                default ->
+                    throw new PbjCompilerException(
+                            "Optional message type [%s] not supported".formatted(field.messageType()));
             };
         }
         return switch (field.type()) {
-            case ENUM -> "%s.fromProtobufOrdinal(readEnum(input))"
-                    .formatted(Common.snakeToCamel(field.messageType(), true));
+            case ENUM ->
+                "%s.fromProtobufOrdinal(readEnum(input))".formatted(Common.snakeToCamel(field.messageType(), true));
```
jasperpotts, line 458: This made me wonder if we need to also handle new enum values added to be able to support full protobuf forward compatibility?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/UnknownField.java
```diff
@@ -0,0 +1,12 @@
...
+public record UnknownField(ProtoConstants wireType, List<Bytes> bytes) {}
```
jasperpotts, line 12: Should we have convenience methods to read as Integer, Long, String etc. Don't know if there is a way to do generics magic auto return type of single get method based on wire type.

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/UnknownFieldsTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.hedera.pbj.runtime.ProtoConstants;
+import com.hedera.pbj.runtime.UnknownField;
+import com.hedera.pbj.runtime.io.buffer.BufferedData;
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.hedera.pbj.test.proto.pbj.MessageWithBytes;
+import com.hedera.pbj.test.proto.pbj.MessageWithBytesAndString;
+import org.junit.jupiter.api.Test;
+
+public class UnknownFieldsTest {
```
jasperpotts, line 16: We need to test round tripping of protobuf objects with all wire types to make sure we can read an object as one with unknown fields, serialize it back out with identical bytes. Then parse as object with all fields and everything works seamlessly..

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
anthony-swirldslabs: I added the `get` prefix on purpose - to minimize the chance of this method colliding with an actual protobuf field named `unknownFields` in the .proto model, and also to distinguish the unknown fields getter from regular protobuf field getters.

While one can still define `int32 getUnknownFields = 1;` or similar in .proto and this will break the PBJ compilation, I think it's unlikely that someone would start their field name with a verb. So the `get` prefix should help us in 99.(9)% of cases.

However, I'm open to other suggestions that address both these points.

anthony-swirldslabs: As for the usage of the feature - we do in fact have a use-case for 2 - it's reported at https://github.com/hashgraph/pbj/issues/456 which this PR is trying to address (together with 1, of course).

I agree that the map looks a bit over-complicated. A list might be a simpler and more efficient data structure. I believe that we could expose the list directly in the API for now because the current use-case described in the ticket seems to simply iterate over all unknown fields w/o ever needing to fetch fields by their field IDs. We can consider adding a map API if there's such a requirement in the future.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/UnknownField.java
anthony-swirldslabs: This is a nice to have feature which we can easily add in the future if/when there's a need for this. For now, clients can use ProtoParserTools utility methods given the provided wireType and the Bytes in the UnknownField.

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/UnknownFieldsTest.java
anthony-swirldslabs: That's an excellent catch! I'm adding a test for just that.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
anthony-swirldslabs: I'm unsure about this. It would be odd to add an otherwise known field to a list of unknown fields just because it uses an unknown enum constant. Also, this wouldn't really help actual applications trying to use the model unless they all are willing to implement an incredible complexity of always checking unknown fields for enum values (+ requiring them to always parse with unknown fields, which would be a performance hit.) I don't believe Google Protobuf does that.

If we encounter a need to support this, we'd need a separate ticket with a reproducible test-case (that works with other protobuf implementations) and then we'd take it from there.

PS. This particular line was changed by `spotlessApply` btw. It's not a part of the fix.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
anthony-swirldslabs: Firstly, this has now been changed to a list.

Secondly, Map.of() and List.of() are very nice, but only when you know the number of elements (and the elements themselves) at compile time. Otherwise, you'd have to use Map.copyOf() (or List.copyOf()) which is essentially no different from the `Collections` methods.

Thirdly, before I write this code, I in fact went and looked at the implementation of the `Collections.unmodifiableMap` and I determined that it's reasonable, small, and performant for our use-case.

Finally, I really didn't want to make this a public constructor. I was about to make it private or package-private. But then I saw that PBJ generates codecs in a separate package (with the `.codec` suffix) and this constructor should only really be accessible from the codec code. But since it lives in a separate Java package, we can't have nice things - like use package-private members in the model. So this is the only reason this constructor is public. It's not really intended to be used by applications code.

But regardless, the main point is that I don't see any performance implications here. Besides, this code only activates when an application explicitly enables the parsing of unknown fields. By default this code sees only nulls here and should execute really fast.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
anthony-swirldslabs: Thanks for the comments. I've changed this to a list. In the API, I expose this list explicitly, rather that a stream over its items. I see your point about the possibility to change the internal implementation in the future. However, a stream API here seems a bit unusual to me. Also, given that we don't have too many real use-cases for this API (other than the one described in the ticket from the mirror node), I think we'll be able to change that in case we find that the list API is just horrible. But I don't think this will happen.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/UnknownField.java
anthony-swirldslabs: Good idea. I've added the doc.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
jasperpotts: For enums you need value enum 0 to be special UNKNOWN value, then when you get a known enum value you set the enum in main object to that UNKNOWN value and add the real value into unknown fields list. See https://chatgpt.com/share/68080a73-dfa0-800f-b96f-09893422125c  and https://protobuf.dev/programming-guides/proto3/#enum

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -189,14 +192,46 @@ public void generate(
...
+                 * in case a repeated field uses an unpacked data format on the wire. It's up to the application
+                 * to interpret these unknown fields correctly if necessary.
+                 * <p>
+                 * If the parsing of unknown fields was enabled when this model instance was parsed originally and
+                 * so this list isn't empty, then a subsequent `Codec.write()` call will preserve all the parsed
+                 * unknown fields as they were originally present on the wire.
+                 *
+                 * @return a (potentially empty) list of unknown fields
+                 */
+                public @NonNull List<UnknownField> getUnknownFields() {
```
jasperpotts, line 228: I am still not a fan of exposing list as public API as when people start using we will be stuck with it forever and I can imagine reasons we might want an alternative internal implementation. But to argue against my self this is also the question of are the new features designed in a way we could go back to being java records? We moved model from record to class to aid with lazy caching but wanted to keep the option of going back to record to be a "value class" when java supports them.  We want to performance test at that point if value class or lazy cache is better performance trade-off. We are hoping there will be a preview JDK with value type support soon.

jasperpotts (commented):
Also model generator,  toString, equals, compareTo and hashCode should all include unknown fields.

Also do we test unknown oneOf values?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
anthony-swirldslabs: I don't use AI, so I didn't open the first link and I don't know what it's saying (and I'm not interested in finding out.)

I've re-read the spec from the second link, and I didn't find references to unknown fields w.r.t. treating unrecognized enum values. It only says:
```
In languages with closed enum types such as Java, a case in the enum is used to represent an unrecognized value, and the underlying integer can be accessed with special accessors.
```
The "special accessor" in this case refers to a dedicated method in the model itself with a name `get<FieldName>Value()` which returns an integer.

So I went ahead and constructed the following test:
```proto
// SPDX-License-Identifier: Apache-2.0
syntax = "proto3";
package pbj.integ.test.enumeration.unknown.value.test;
option java_multiple_files = true;

enum TestEnum {
  TEST_ENUM_UNKNOWN = 0;
  A = 1;
  B = 2;
}

message MessageWithEnum {
  TestEnum value = 1;
}

// Protobuf uses a var int to persist an enum value.
// So on the wire, the MessageWithInt looks exactly the same as the MessageWithEnum above.
// But we can now put a random number into the enum field using this alternative model,
// thus emulating a currently-unknown, future enum element:
message MessageWithInt {
  int32 value = 1;
}
```

and Java part:

```java
package com.hedera.pbj.integration.test;

import org.junit.jupiter.api.Test;
import pbj.integ.test.enumeration.unknown.value.test.TestEnum;

import java.io.ByteArrayOutputStream;

import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertTrue;

public class UnknownEnumTest {
    @Test
    void testUnknownEnum() throws Exception {
        // Use Google Protoc models here and create MessageWithInt(666):
        pbj.integ.test.enumeration.unknown.value.test.MessageWithInt msg1 = pbj.integ.test.enumeration.unknown.value.test.MessageWithInt.newBuilder().setValue(666).build();

        // Serialize it into a byte array:
        ByteArrayOutputStream baos = new ByteArrayOutputStream();
        msg1.writeTo(baos);
        final byte[] byteArray = baos.toByteArray();

        // Parse it back as MessageWithEnum:
        pbj.integ.test.enumeration.unknown.value.test.MessageWithEnum msg2 = pbj.integ.test.enumeration.unknown.value.test.MessageWithEnum.parseFrom(byteArray);

        assertEquals(TestEnum.UNRECOGNIZED, msg2.getValue());
        assertEquals(666, msg2.getValueValue());
        assertTrue(msg2.getUnknownFields().isEmpty());
    }
}
```

And the test passes. As you can see, Google Protoc does not put unrecognized enum values into unknown fields. It instead provides a specialized getter, which is named `getValueValue()` in this case.

For this reason, I don't believe PBJ should add unrecognized enum values to unknown fields because they aren't unknown fields according to the Protobuf spec.

I'll re-iterate: I believe that improving handling of unrecognized enum values should be a separate task unrelated to this ticket or this PR. And the test above proves that unrecognized enum values have nothing to do with unknown fields as defined by Protobuf.

anthony-swirldslabs: I filed https://github.com/hashgraph/pbj/issues/476 because the enum issue is unrelated to the unknown fields issue specifically.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecWriteMethodGenerator.java
```diff
@@ -39,6 +39,14 @@ static String generateWriteMethod(
              */
             public void write(@NonNull $modelClass data, @NonNull final WritableSequentialData out) throws IOException {
                 $fieldWriteLines
+                // Check if not-empty to avoid creating a lambda if there's nothing to write.
```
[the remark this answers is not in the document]
anthony-swirldslabs: Google implementation doesn't maintain the fields order either. Here's how Google's `writeTo()` looks like:
```
  @java.lang.Override
  public void writeTo(com.google.protobuf.CodedOutputStream output)
                      throws java.io.IOException {
    if (value_ != pbj.integ.test.enumeration.unknown.value.test.TestEnum.TEST_ENUM_UNKNOWN.getNumber()) {
      output.writeEnum(1, value_);
    }
    getUnknownFields().writeTo(output);
  }
``` 

Ours is essentially the same. So there's not a problem.

Also, the high and low field numbers matter per se - for those individual fields representations on the wire. It doesn't matter whether you serialize low numbers first or high numbers first - the total record size will be exactly the same because each field is serialized completely independently of all the other fields.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -145,9 +152,20 @@ static String generateParseLoop(
                                             // Since we are parsing is strict mode, this is an exceptional condition.
                                             throw new UnknownFieldException($prefixfield);
                                         } else {
-                                            // We just need to read off the bytes for this field to skip it
-                                            // and move on to the next one.
-                                            skipField(input, ProtoConstants.get(wireType), $skipMaxSize);
+                                            if (parseUnknownFields) {
```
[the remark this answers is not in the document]
anthony-swirldslabs: Good idea. Done.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
anthony-swirldslabs: Alright. Let's expose a Stream then.

anthony-swirldslabs: On a second thought, no, let's not. If this is defined as a Stream API, then the Codec.wirte() method would always create a new stream at least once, or perhaps even twice (you'll see why in the next revision.) This would negatively affect the performance of the serialization for every model, even if it's never had any unknown fields at all and even if we find a way to invoke it just once instead of twice.

For this performance-related reason, let's continue using `List` here for now.

We'll work with the users of this API if we ever need to change it. Also, we'll be able to support this List API even if we change the internal implementation, although with a possible performance hit for users of this API only. Again, by working with the users, we'll be able to migrate them. And again, this would only be necessary if we actually have to replace the API.

steven-sheehy: I would prefer avoiding Stream as well as it's heavy weight and this would be used in our fast path code. We generally avoid Stream when processing a transaction as it creates way too many small objects.

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/UnknownFieldsTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+        assertEquals(2, uf.field());
+        assertEquals(ProtoConstants.WIRE_TYPE_DELIMITED, uf.wireType());
+        assertEquals(TEST_STRING.length(), uf.bytes().getVarInt(0, false));
+        assertEquals(TEST_STRING, uf.bytes().asUtf8String(1, uf.bytes().length() - 1));
+    }
+
+    @Test
+    void testUnknownFieldsRoundTrip() throws Exception {
+        // write MessageWithBytesAndString
```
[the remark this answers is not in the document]
anthony-swirldslabs: I'm working on a more comprehensive test - we'll employ our good buddy `Everything` for this.

Just wanted to note regarding the Google round trip wire representation compatibility: yes, I'm working on making it compatible, however, this now requires sorting by the field number because Google uses a `TreeMap<Integer, ...>` to store the fields. I don't want to use a sorted collection in our case for performance reasons. But the `write()` method performance will be hit because of the sorting if there's any unknown fields present.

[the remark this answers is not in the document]
anthony-swirldslabs: I've added a comprehensive test that uses `Everything` to test the round-trip for any and all types that we support. It works.

As mentioned above, unknown enum values is a completely independent issue, unrelated to the unknown fields support. Please find all the details in above comments, and please the issue that I filed for them: #476

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -516,6 +558,15 @@ public int hashCode() {
 
         bodyContent += statements;
 
+        bodyContent +=
+            """
+                    if ($unknownFields != null) {
```
jasperpotts, line 563: Not sure on answer but thinking of how will unknown fields effect stable hashcodes. I think it should be fine assuming unknown fields that have a default value are not in the `$unknownFields` list? Do we ensure that is true? Otherwise I have a model object as key in map and then parse a new key like AccountId with schema with new field and can no longer find the value in map.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -151,21 +155,19 @@ static String generateParseLoop(
...
-                                                ));
-                                            } else {
-                                                // We just need to read off the bytes for this field to skip it
-                                                // and move on to the next one.
-                                                skipField(input, ProtoConstants.get(wireType), $skipMaxSize);
+                                        } else if (parseUnknownFields) {
+                                            if ($unknownFields == null) {
+                                                $unknownFields = new ArrayList<>($initialSizeOfUnknownFieldsArray);
                                             }
+                                            $unknownFields.add(new UnknownField(
```
jasperpotts, line 162: I think we will not get here for default values as they are not written in the wire format. So we should not have fields with default values in unknownFields. Is that correct?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/UnknownField.java
```diff
@@ -10,9 +12,78 @@
...
+        hashCode ^= hashCode >>> 24;
+        hashCode += hashCode << 30;
+
+        return hashCode;
+    }
+
+    /**
+     * Prints a protobuf model.toString() representation for this unknown field into the given StringBuilder.
+     */
+    public void printToString(final StringBuilder sb) {
```
jasperpotts, line 79: I guess this needs to be public as called from generated code? It would be nice to add full JavaDoc comments to at least all public API. Also comment like the implementation of this will have to be stable over time as will be depended on.

jasperpotts (dismissed):
I think all good, just please check the handling of unknown fields with default values and think through stable hashcode and equals needed for use of PBJ model objects as keys in DB. While needing to be able to add fields to those objects over time and the keys still match.  Maybe a unit test to make sure that is true and stays true would be good.

jasperpotts:
It would be great to have Artem and Joseph review this PR as well because there are so many edge details to remember like stable hashcode.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecWriteMethodGenerator.java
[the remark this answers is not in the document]
jsync-swirlds: Field order is kind of important because one prime use case (block node forward compatibility) requires the ability to round-trip the protobuf object and obtain the exact same serialized bytes (otherwise we end up with a different hash and the block signature won't verify).

PBJ determinism is one of the things that _allows_ us to use it in state and block stream.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
jsync-swirlds: I don't think it's possible to have an unknown field object with default value, because it wouldn't be in the serialized form we parsed (possible interaction with the `optional` keyword here, however).
Basically, we don't know about unknown fields that do not exist in the input data...

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/UnknownFieldsTest.java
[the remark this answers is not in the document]
jsync-swirlds: I think this feature may be a point where we cannot maintain total wire format _match_ with ProtoC.
We can remain compatible, but not identical because ProtoC is non-deterministic (it serializes the object differently when the same fields are unknown or known fields).
PBJ, however, needs to preserve stable serialization so we can generate matching hash values...

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecWriteMethodGenerator.java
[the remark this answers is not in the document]
anthony-swirldslabs: Good point. The current version of this PR sorts unknown fields by the field number. When serializing, it writes them after all the known field in this same order.

While Google Protoc implementation uses a different data structure (a TreeMap), the end result is (supposed to be) the same: their TreeMap is sorted by the field number (it's the key of their map), and they write all the unknown fields in the end after writing known fields.

So I think that in this case we should be both deterministic and also still binary-compatible with what the Google implementation does.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
anthony-swirldslabs: PBJ can only see what it's parsed. By definition, it's unable to know about unknown fields that may have default values. I originally didn't include hashCode/equals and friends because I was considering a use case where an old application might consider the model equal as long as all the known fields are equal. But then again, there are applications where the business logic might fail because of this. So went ahead and implemented these methods as you suggested, because there doesn't seem to be the "right option" either way.

Besides, Google protoc includes the unknown fields in equal/hashCode, etc. So at least we're compatible with Google behavior here. And that's the best we can do here.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
anthony-swirldslabs: Yes, that's correct. Unless the serialized form was produced by a protobuf implementation that writes default values. We'd honestly parse them into our unknown fields then, and would write them when PBJ serializes the data. This would ensure a compatible round-trip with such an implementation.

However, I'm unaware of such protobuf implementations. So we can safely assume we don't have default values in our unknown fields.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/UnknownField.java
anthony-swirldslabs: I've updated the javadoc for all the public methods here in the latest commit.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecWriteMethodGenerator.java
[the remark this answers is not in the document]
jsync-swirlds: The issue is we have to match between an entity that knows the new schema, and an implementation that does not know the new schema.  Basically, if we move unknown fields to the end of the serialization, we're breaking the serialization compatibility for block node, because it won't get the same has value that consensus had (assuming CN has the new schema and BN has an older version).
Matching Google is not as important as matching between CN and BN without the requirement to upgrade both in lock-step.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -400,6 +439,29 @@ public int compareTo($javaRecordName thatObj) {
 
         bodyContent += Common.getFieldsCompareToStatements(fields, "");
 
+        bodyContent +=
+                """
+                // Treat null and empty lists as equal
+                if ($unknownFields != null && !$unknownFields.isEmpty()) {
+                    if (thatObj.$unknownFields == null || $unknownFields.size() != thatObj.$unknownFields.size()) {
+                        // This has unknown fields, that one doesn't. So we're greater:
+                        return 1;
```
artemananiev, line 448: This is right, when `thatObj.$unknownFields == null || thatObj.$unknownFields.isEmpty()`. However, if the the other obj does have unknown fields, but the size is different, such implementation would violate compareTo contract: if a.compareTo(b) is 1, then b.compareTo(a) must be -1, and vice versa. In this case, both may be 1, if `$unknownFields.size()`s are different

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -189,14 +192,50 @@ public void generate(
...
+                 * to interpret these unknown fields correctly if necessary.
+                 * <p>
+                 * If the parsing of unknown fields was enabled when this model instance was parsed originally and
+                 * the unknown fields were present, then a subsequent `Codec.write()` call will persist all the parsed
+                 * unknown fields.
+                 *
+                 * @return a (potentially empty) list of unknown fields
+                 */
+                public @NonNull List<UnknownField> getUnknownFields() {
+                    return $unknownFields == null ? Collections.EMPTY_LIST : $unknownFields;
```
artemananiev, line 233: The doc says the method returns an unmodifiable list. Here it returns the field as is, and it looks like the list is created as modifiable in the constructor

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -175,6 +175,20 @@ public static Bytes fromHex(@NonNull final String string) {
...
 
+    /**
+     * Creates a new Bytes object by merging the bytes from the given bytes1 and bytes2.
+     * @param bytes1 first Bytes
+     * @param bytes2 second Bytes
+     * @return a Bytes object with bytes of bytes1 followed by bytes of bytes2
+     */
+    @NonNull
+    public static Bytes merge(@NonNull final Bytes bytes1, @NonNull final Bytes bytes2) {
+        final byte[] array = new byte[bytes1.length + bytes2.length];
```
artemananiev, line 186: Is it worth checking if bytes1 / bytes2 is empty to avoid unnecessary array copies?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecWriteMethodGenerator.java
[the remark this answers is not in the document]
anthony-swirldslabs: I see what you're saying. However, preserving the exact fields order, for both known and unknown fields, would introduce a notable complexity because the model class would have to maintain the exact fields order as has been read from the input for all the fields. This would also make the write() method extremely slower when the unknown fields are present.

Secondly, PBJ currently implements a certain deterministic order of writing fields. If a CN instance uses a Protobuf implementation that emits fields in a different order, then your concern would be a huge problem regardless of the presence of unknown fields or whether PBJ supports them or not at all.

If the problem statement of preserving the original binary representation is the crux of the BN behavior, then it would be reasonable to consider a BN implementation that doesn't try to re-serialize a previously deserialized input object with whatever Protobuf implementation it happens to use at the moment. I believe you want to use the original bytes for the purpose of persisting the data or fanning it out to downstream consumers. You may deserialize it in memory to see what's inside, but you shouldn't try to serialize it back into the binary form and pretend that's what you've got originally from your source of truth, because it isn't, by definition.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
anthony-swirldslabs: We'd save one (or two) `System.arraycopy` calls at most, which would be no-ops for all intents and purposes anyway. So we're talking about saving way less than 1ms of time. However, we'd lose this very time on running the conditional statements, I think. For this reason, I don't believe this is worth optimizing right now, especially given that the current usages are unlikely to benefit from such an optimization. Once we start using this method elsewhere and notice its underperformance due to this edge case, we can always come back and tweak its internal implementation.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
anthony-swirldslabs: That's a good catch! I had an `isEmpty()` check originally but then over-optimized this implementation and missed this edge case. Thanks for noticing! I'm fixing this in the next commit.

PR #477, opened 2025-04-25
feat: implement PbjMessage interface

### LINKED ISSUE #287: Add interface to PBJ records
A PBJ record has a static `PROTOBUF` variable for encoding/decoding a record to/from protobuf. In order to create a tool that can encode an arbitrary PBJ record, it would be useful to introduce an interface to these records (`PbjRecord`). The interface would have a method like `getProtoCodec` that would return this static variable. The tool could then encode an arbitrary record without any additional data.

anthony-swirldslabs:
**Description**:
Introducing `PbjMessage<T>` interface that every generated message implements. The interface allows a code having a PBJ object to access its static fields (e.g. the `<ModelClass>.PROTOBUF` codec) w/o knowing the concrete class of the model at compile time. This allows for more robust and simple code: instead of passing both the model object and its codec object to a serialization method, the application can now pass the model object only, and the serialization method can retrieve the correct codec directly from the model.

A need for this was previously discovered at https://github.com/hiero-ledger/hiero-consensus-node/pull/15417 where the code currently has to ensure it passes the correct codec for a given model which may sometimes be error-prone.

**Related issue(s)**:

Fixes #287 

**Notes for reviewer**:
The changes are trivial as we add three simple getters to each model. All tests should continue to pass.

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
@@ -645,15 +645,13 @@ public final class ModelGenerator implements Generator {
         return """
                 /** Protobuf codec for reading and writing in protobuf format */
                 public static final Codec<$modelClass> PROTOBUF = new $qualifiedCodecClass();
-                /** {@inheritDoc} */ @Override public Codec<$modelClass> getProtobufCodec() { return PROTOBUF; }
+                /** {@inheritDoc} */ @Override public Codec<$modelClass> PROTOBUF() { return PROTOBUF; }
 
                 /** JSON codec for reading and writing in JSON format */
                 public static final JsonCodec<$modelClass> JSON = new $qualifiedJsonCodecClass();
-                /** {@inheritDoc} */ @Override public JsonCodec<$modelClass> getJsonCodec() { return JSON; }
 
                 /** Default instance with all fields set to default values */
                 public static final $modelClass DEFAULT = newBuilder().build();
-                /** {@inheritDoc} */ @Override public $modelClass getDefaultMessage() { return DEFAULT; }
                 """
                 .replace("$modelClass", javaRecordName)
                 .replace("$qualifiedCodecClass", lookupHelper.getFullyQualifiedMessageClassname(FileType.CODEC, msgDef))
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
@@ -85,6 +85,7 @@ public final class ModelGenerator implements Generator {
         final List<String> hasMethods = new ArrayList<>();
         // The generated Java import statements. We'll build this up as we go.
         writer.addImport("com.hedera.pbj.runtime.*");
+        writer.addImport("com.hedera.pbj.runtime.PbjMessage");
         writer.addImport("com.hedera.pbj.runtime.io.*");
         writer.addImport("com.hedera.pbj.runtime.io.buffer.*");
         writer.addImport("com.hedera.pbj.runtime.io.stream.*");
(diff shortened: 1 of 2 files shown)


jasperpotts:
I am torn on if this is the right way forward, it would be hard to remove once added. I work if it would impact being a `record` in future. Think it doesn't break that. I am torn on mixing `getX()` method style with `x()` style. Though can see the `get` style has less chance of clashing with field names. Just wondering if the gain is worth the cost in terms of making future movement maybe harder as more API to support.

anthony-swirldslabs:
> I am torn on if this is the right way forward, it would be hard to remove once added. I work if it would impact being a `record` in future. Think it doesn't break that. I am torn on mixing `getX()` method style with `x()` style. Though can see the `get` style has less chance of clashing with field names. Just wondering if the gain is worth the cost in terms of making future movement maybe harder as more API to support.

Could you please elaborate on what impact w.r.t. switching back to `record` this could possibly have? Records can implement interfaces, and we already use this with `Comparable`. This new interface is essentially no different. So I don't see anything breaking if we go back to records. Could you please clarify what you think could break with this new interface if we make them records again?

And yes, I chose to use `get` explicitly to underscore that these methods don't return any actual model fields, but instead return some "special", "system" values. I'm happy to replace `get` with something else. But I don't believe exposing them as `protobufCodec` etc would be the right thing to do as the names would look confusing when there's similarly named getters for normal model fields. Please suggest a replacement for `get` if you like. E.g. I'm happy to make it an underscore, or whatever else you have in mind for this.

To address the "it would be hard to remove once added", I'm happy to reduce the scope to the `getProtobufCodec()` only and remove the other two methods. I don't believe we'll ever ditch support for protobuf, after all its **PB**j, right? So this method is unlikely to need any changes in the future and could stay. And it's the only one that is really necessary. I've added the other two for completeness only, but I have absolutely no strong feelings about them. Do you want me to remove the other two methods?

anthony-swirldslabs:
> And yes, I chose to use `get` explicitly to underscore that these methods don't return any actual model fields, but instead return some "special", "system" values. I'm happy to replace `get` with something else. But I don't believe exposing them as `protobufCodec` etc would be the right thing to do as the names would look confusing when there's similarly named getters for normal model fields. Please suggest a replacement for `get` if you like. E.g. I'm happy to make it an underscore, or whatever else you have in mind for this.
> 
> To address the "it would be hard to remove once added", I'm happy to reduce the scope to the `getProtobufCodec()` only and remove the other two methods. I don't believe we'll ever ditch support for protobuf, after all its **PB**j, right? So this method is unlikely to need any changes in the future and could stay. And it's the only one that is really necessary. I've added the other two for completeness only, but I have absolutely no strong feelings about them. Do you want me to remove the other two methods?

How about we call it `PROTOBUF()` in all caps? Advantages:
1. Looks sort of like a record getter.
2. Looks distinctly different from normal field getters.
3. Suggests that it returns a static constant.
4. Its name is literally equal to the name of the constant in the same class, so there's no confusion as to what the method returns.

Would that resolve the concern?

jasperpotts:
I chatted to Richard and he had another thought. Would it be better to add utility class for getting codec from object instance. Could be generated and do instance based switch inside. It would provide similar convenance without adding more API to every model object. Something to think about. I like `PROTOBUF()` better just not a fan of mixing get style and record style if we can avoid it. Though agree on argument of avoiding name clash.

anthony-swirldslabs:
An instance based switch doesn't seem great. It would require all your PBJ models to be compiled in one go. There may be instances where you source your protobuf definitions from different locations and/or run the PBJ generator separately for those different sets of .proto files. So it would be impossible for all of them to contribute to a single instance-based-switch utility to account for all the models known to a given Java application (unless we resort to reflection, scanning the class path, and similar, which we probably won't do.) This would also cause a class name clash in case you try to use several .jar files compiled independently each of which compiled their own PBJ models. So this solution seems brittle.

I understand where you're coming from w.r.t. adding new APIs, and I fully agree with your point in general. However, the PBJ seems a little less widely used than, say, the JDK, so the same level of strictness may not fully apply here (although I agree it's very useful as a template and a bar rising standard.) We currently control and/or have pretty much direct access to all the consumers of PBJ, and we can make necessary changes if we find ourselves having to break our own APIs - we've been there already (e.g. when we added real nested classes.)

To address concerns, I've updated the interface to use `PROTOBUF()` in the latest commit, and removed the other two methods from the interface.

lpetrovic05 (approved):

abies:
This will trip off every style checker out there to have a method name with all caps...

anthony-swirldslabs:
@abies : we normally don't run style checkers on generated code, so this shouldn't be a problem.

Overall, however, I see that @jasperpotts isn't extremely happy with the solution, so I'm closing it for now.

PR #484, opened 2025-05-15
fix: handling empty msg in helidon plugin

### LINKED ISSUE #483: bug: Helidon Plugin handling empty unary requests
### Description

When a unary gRPC method is called with an empty request body, which translates to a 0-length Protobuf message, the current gRPC frame parsing logic in the `data()` method fails to call `pipeline.onNext()`. This effectively means the service method handler for that gRPC call is never invoked, and the request hangs or times out for the client.

The issue occurs because the entire 5-byte gRPC frame (1 byte for compression flag + 4 bytes for message length of 0) is consumed, leading to `data.available()` becoming 0. The `while (data.available() > 0)` loop in the `data()` method then terminates before the `ReadState.READ_ENTITY_BYTES` case can be processed for this zero-length message, which is where `pipeline.onNext()` would typically be called after reading the (non-existent) entity bytes.

### Steps to reproduce

1.  Define a unary gRPC service method that accepts an empty Protobuf message as a request (e.g., `message EmptyRequest {}`).
2.  Call this gRPC method from a client, sending an empty request (`{}`).
3.  Observe that the server-side handler for this method is not invoked, and the client request likely times out or hangs.
4.  Debugging the `data()` method on the server will show that `pipeline.onNext()` is not called for this request.

### Additional context

_No response_

### Hedera network

_No response_

### Version

main, 0.11.3

### Operating system

None

georgi-l95:
**Description**:
Adds handing of empty requests. In those cases we receive 5 bytes, where first one is compression flag and the rest 4 are the lenght of the message that usually follows. However in case of empty request, such as [ServerStatusRequest](https://github.com/hashgraph/hedera-protobufs/blob/v0.62.1/block/block_service.proto#L837) data() method receive exactly 5 bytes, which results in not reaching onNext call on the pipeline and the client hangs, as it's waiting for the respons.
In this PR we add a condition where length is 0, so that we can continue the flow corrrectly and reset the state for the next messages. More info in the issue #483 

**Related issue(s)**:

Fixes #483 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
+++ b/pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
@@ -380,8 +380,17 @@ final class PbjProtocolHandler implements Http2SubProtocolSelector.SubProtocolHa
                             // would break the application. So we need a new buffer each time :-(
                             entityBytes = new byte[(int) length];
                             entityBytesIndex = 0;
-                            // done with length now, so move on to next state
-                            currentReadState = ReadState.READ_ENTITY_BYTES;
+
+                            // In this case we have a request with no message, like an empty unary request
+                            if (length == 0) {
+                                final Bytes bytes = Bytes.wrap(entityBytes);
+                                pipeline.onNext(bytes);
+                                entityBytes = null;
+                                currentReadState = ReadState.START;
+                            } else {
+                                // done with length now, so move on to next state
+                                currentReadState = ReadState.READ_ENTITY_BYTES;
+                            }
                         }
                         break;
                     }
@@ -483,8 +492,8 @@ final class PbjProtocolHandler implements Http2SubProtocolSelector.SubProtocolHa
                                 case "m" -> TimeUnit.MILLISECONDS;
                                 case "u" -> TimeUnit.MICROSECONDS;
                                 case "n" -> TimeUnit.NANOSECONDS;
-                                    // This should NEVER be reachable, because the matcher
-                                    // would not have matched.
+                                // This should NEVER be reachable, because the matcher
+                                // would not have matched.
                                 default -> throw new GrpcException(GrpcStatus.INTERNAL, "Invalid unit: " + unit);
                             });
             return deadlineDetector.scheduleDeadline(deadline, () -> {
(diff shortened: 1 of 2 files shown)


imalygin (dismissed):

pbj-core/pbj-grpc-helidon/src/test/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandlerTest.java
```diff
@@ -155,6 +155,31 @@ void contentTypeIsNormalized() {
         assertThat(handler.streamState()).isEqualTo(Http2StreamState.HALF_CLOSED_REMOTE);
     }
 
+    /**
+     * Passing valid empty request is handled correctly. Such case is for example unary api asking for status, without passing anything in the request body.
+     */
+    @Test
+    void requestIsSuccessfulOnEmptyData() {
+        final var h = WritableHeaders.create();
```
anthony-swirldslabs, line 163: Minor: would be nice to use real types instead of `var`.

anthony-swirldslabs (dismissed):

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -380,8 +380,17 @@ public void data(@NonNull final Http2FrameHeader header, @NonNull final BufferDa
...
+                            // In this case we have a request with no message, like an empty unary request
+                            if (length == 0) {
+                                final Bytes bytes = Bytes.wrap(entityBytes);
+                                pipeline.onNext(bytes);
+                                entityBytes = null;
+                                currentReadState = ReadState.START;
+                            } else {
+                                // done with length now, so move on to next state
+                                currentReadState = ReadState.READ_ENTITY_BYTES;
+                            }
```
anthony-swirldslabs, line 393: Minor: we have `Bytes.EMPTY`, so we can avoid allocating a new bytes array and then a new Byte object in this case and only nullify the entityBytes in the positive branch. The `new byte[]` statement could be moved inside the else branch.

pbj-core/pbj-grpc-helidon/src/test/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandlerTest.java
georgi-l95: Done, thank you.

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
georgi-l95: Good point, thank you. Moved the allocations, please check again when time permits.

anthony-swirldslabs (approved):

anthony-swirldslabs:
I merged the fix, and it'll be available in PBJ v0.11.4 shortly: https://central.sonatype.com/artifact/com.hedera.pbj/pbj-runtime/versions

PR #532, opened 2025-06-27
test: more tests for GRPC client

### LINKED ISSUE #531: Add more GRPC client comprehensive tests

anthony-swirldslabs:
**Description**:
Adding more tests for PBJ GRPC client to verify various scenarios, especially failures, including a GRPC server dying. In addition to existing tests that run both the server and the client in the current JVM, 4 more tests are added that run the server in a separate JVM, allowing us to kill the server process w/o any clean up or a chance to close the call gracefully, and then see what happens on the client side.

I believe we're covering many if not all possible scenarios. If you have more ideas, please let me know in the PR comments and I'll file more issues and will open more PRs in the future. Otherwise, we'll consider this to be the final PR to complete the PBJ GRPC client testing for now.

**Related issue(s)**:

Fixes #531 

**Notes for reviewer**:
All tests pass.

**Checklist**

### DIFF
--- a/pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcCall.java
+++ b/pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcCall.java
@@ -4,13 +4,17 @@ package com.hedera.pbj.grpc.client.helidon;
 import com.hedera.pbj.runtime.Codec;
 import com.hedera.pbj.runtime.ParseException;
 import com.hedera.pbj.runtime.grpc.GrpcCall;
+import com.hedera.pbj.runtime.grpc.GrpcException;
+import com.hedera.pbj.runtime.grpc.GrpcStatus;
 import com.hedera.pbj.runtime.grpc.Pipeline;
 import com.hedera.pbj.runtime.grpc.ServiceInterface;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
 import io.helidon.common.buffers.BufferData;
 import io.helidon.common.socket.HelidonSocket;
+import io.helidon.http.HeaderName;
 import io.helidon.http.HeaderNames;
 import io.helidon.http.HeaderValues;
+import io.helidon.http.Headers;
 import io.helidon.http.WritableHeaders;
 import io.helidon.http.http2.Http2FrameData;
 import io.helidon.http.http2.Http2Headers;
@@ -23,6 +27,11 @@ import io.helidon.webclient.http2.Http2ClientStream;
 import io.helidon.webclient.http2.Http2StreamConfig;
 import io.helidon.webclient.http2.StreamTimeoutException;
 import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
 
 /**
  * A blocking, non-buffering wrapper around an HTTP2 stream that serves a single GRPC call.
@@ -44,6 +53,8 @@ import java.time.Duration;
 public class PbjGrpcCall<RequestT, ReplyT> implements GrpcCall<RequestT, ReplyT> {
 
     private static final BufferData EMPTY_BUFFER_DATA = BufferData.empty();
+    private static final HeaderName GRPC_STATUS = HeaderNames.createFromLowercase("grpc-status");
+    private static final HeaderName GRPC_MESSAGE = HeaderNames.createFromLowercase("grpc-message");
 
     private final PbjGrpcClient grpcClient;
     private final Codec<RequestT> requestCodec;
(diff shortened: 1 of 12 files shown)


pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcCall.java
```diff
@@ -194,19 +212,92 @@ private void receiveRepliesLoop() {
...
+                    pipeline.onError(new GrpcException(
+                            grpcStatus < GrpcStatus.values().length
+                                    ? GrpcStatus.values()[grpcStatus]
+                                    : GrpcStatus.UNKNOWN,
+                            message));
+                    onErrorCalled = true;
+                }
+            } catch (NumberFormatException ignored) {
+                // a bad server sent an invalid header. This shouldn't happen really.
+                System.err.println(String.format("Invalid GRPC_STATUS: %s", value));
```
jsync-swirlds, line 297: Two questions:
1. Should we call pipeline.onError to signal a communication issue?
1. Should we use the `System.Logger` facility (which can be redirected to a preferred logging implementation by the user of the generated client) rather than `System.err.println` for diagnostic output?

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/grpc/GoogleProtobufGrpcServerGreeterHandle.java
```diff
@@ -166,6 +170,17 @@ public synchronized void stop() {
         }
     }
 
+    @Override
+    public synchronized void stopNow() {
+        if (server != null) {
+            try {
+                server.shutdownNow().awaitTermination();
+            } catch (InterruptedException ignored) {
+            }
```
jsync-swirlds, line 179: Rather than completely ignoring this, should we reset the interrupted flag (with `Thread.currentThread().interrupt()`)?  That would, at least, help ensure other possibly blocking code on this thread has a chance to exit.

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/grpc/GrpcClientComprehensiveTest.java
```diff
@@ -59,27 +50,34 @@ public GrpcClientComprehensiveTest(Function<Integer, GrpcServerGreeterHandle> se
...
+    private static <T extends Throwable> T assertThrowsCause(Class<T> expectedType, Executable executable) {
+        try {
+            executable.execute();
+        } catch (final Throwable t) {
+            for (Throwable cause = t; cause != null; cause = cause.getCause()) {
+                if (cause.getClass().equals(expectedType)) {
+                    return (T) cause;
+                }
+            }
+            throw new AssertionFailedError(
```
jsync-swirlds, line 62: Before throwing here, would it make sense to also check for suppressed exceptions that match the expected type?

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/grpc/GrpcTestUtils.java
```diff
@@ -0,0 +1,42 @@
...
+    public static final Options PROTO_OPTIONS =
+            new Options(Optional.empty(), ServiceInterface.RequestOptions.APPLICATION_GRPC);
+
+    /** sleep() for non-blocking GRPC calls to wait for results to arrive. */
+    public static void sleep(final GrpcClient grpcClient) {
+        // Wait a tad longer than the read timeout:
+        try {
+            Thread.sleep(((PbjGrpcClient) grpcClient).getConfig().readTimeout().plusSeconds(2));
+        } catch (InterruptedException e) {
+        }
```
jsync-swirlds, line 29: If you're just ignoring the interrupt; it might be better to use `parkNanos` or `parkUntil` on LockSupport (which are calls `sleep` uses under the hood).
Example:
```suggestion
        LockSupport.parkUntil(System.currentTimeMillis() + ((PbjGrpcClient) grpcClient).getConfig().readTimeout().plusSeconds(2));
```

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/grpc/specialized/SeparateJVMRunner.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    @Override
+    public void start() {
+        final String javaHome = System.getProperty("java.home");
+        final Path javaPath = Paths.get(javaHome, "bin", "java");
+
+        final ProcessBuilder pb = new ProcessBuilder(
+                javaPath.toString(),
+                "-cp",
```
jsync-swirlds, line 44: Might need to use module path here instead of classpath; with the latest Hiero Conventions plugin, unit tests should prefer the module path.

jsync-swirlds (approved):
A few questions, but nothing blocking.

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/grpc/GoogleProtobufGrpcServerGreeterHandle.java
anthony-swirldslabs: Will call `interrupt()`.

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/grpc/GrpcTestUtils.java
anthony-swirldslabs: Sounds good. Will use LockSupport here.

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/grpc/GrpcClientComprehensiveTest.java
anthony-swirldslabs: Ah, got it. Makes sense. Thanks for the pointer!
I've just added support for checking the suppressed exceptions here.

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/grpc/specialized/SeparateJVMRunner.java
jsync-swirlds: It _should_ be possible to fetch the module path (as far as what is set on the module path command line) from `System.getProperty('jdk.module.path')`.

jsync-swirlds (approved):

imalygin (approved):

Issue #556, opened 2025-08-03
Bytes.hashCode() is degenerate for short arrays


OlegMazurov:
### Description

`Bytes.hashCode()` uses multiplier 31 to mix consecutive bytes:
```
            for(int i = this.start + this.length - 1; i >= this.start; --i) {
                h = 31 * h + UnsafeUtils.getArrayByteNoChecks(this.buffer, i);
            }
```
While the mix may be considered good, the range of computed values for short arrays is unacceptable. F.e, four bytes representing 4 billion of `int` values produce only 7849921 distinct hash codes. That leads to lots of unjustified hash collisions and poor `HashMap` performance.
For good characteristics, the multiplier should be prime, contain many 1's  and be larger than 256. A good candidate is 503.

### Steps to reproduce

Run the program:
```
    void test() {
        byte[] bytes = new byte[4];
        long[] set = new long[1 << 26];
        long count = 0;
        long distinct = 0;
        for (int i0 = 0; i0 < 256; ++i0) {
            bytes[0] = (byte) i0;
            for (int i1 = 0; i1 < 256; ++i1) {
                bytes[1] = (byte) i1;
                for (int i2 = 0; i2 < 256; ++i2) {
                    bytes[2] = (byte) i2;
                    for (int i3 = 0; i3 < 256; ++i3) {
                        bytes[3] = (byte) i3;
                        long h = Bytes.wrap(bytes).hashCode() & 0xffffffffL;
                        if ((set[(int)(h/64)] & (1L << (h % 64))) == 0) {
                            set[(int)(h/64)] |= (1L << (h % 64));
                            distinct++;
                        }
                        count++;
                    }
                }
            }
        }
        System.out.printf("Number of values: %d, distinct hash codes: %d\n", count, distinct);
    }
```
Output:
```
Number of values: 4294967296, distinct hash codes: 7849921
```

### Version

v0.11.6

### Operating system

None

jasperpotts:
Would be helpful or too costly to add randomization that we do in generated model classes?
```
            // Shifts: 30, 27, 16, 20, 5, 18, 10, 24, 30
            hashCode += hashCode << 30;
            hashCode ^= hashCode >>> 27;
            hashCode += hashCode << 16;
            hashCode ^= hashCode >>> 20;
            hashCode += hashCode << 5;
            hashCode ^= hashCode >>> 18;
            hashCode += hashCode << 10;
            hashCode ^= hashCode >>> 24;
[...]

PR #558, opened 2025-08-04
fix: 556: Bytes.hashCode() is degenerate for short arrays

artemananiev:
Fixes: https://github.com/hashgraph/pbj/issues/556


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/PbjMap.java
```diff
@@ -129,7 +129,7 @@ public int hashCode() {
         // This is a convenience hashCode() implementation that delegates to Java hashCode,
         // and it's implemented here solely to support the above equals() method override.
         // Generated protobuf models compute map fields' hash codes differently and deterministically.
-        return 31 * map.hashCode() + sortedKeys.hashCode();
+        return 255 * map.hashCode() + sortedKeys.hashCode();
```
anthony-swirldslabs, line 132: I'm not a cryptography/hashing expert, but 255 isn't prime, while 31 is. I strongly believe that being a prime number is important when it comes to hashing multipliers.

If you must change the multiplier (the reason for this isn't obvious btw - would be nice to explain that), could you please find a prime multiplier?

The comment applies to this entire fix.

artemananiev: @OlegMazurov and I tested a few multipliers, mostly primes (like 311 or 503), but some non-primes, too. 255 showed the best results, that's why it's in the PR

anthony-swirldslabs: What is a "best result"? How much worse a result would, say, 311 produce?

anthony-swirldslabs: Thanks for providing the results. I see where the 255 is coming from, and the reason why it's considered "the best".

However, I'd like to hear from cryptography/hashing/math experts on the choice of a non-prime multiplier before I could approve the fix.

jsync-swirlds: For a non-secure hash, the primary value is dispersion (how many different hashes do we get and how evenly are they distributed) and (for hash map) avoiding having a common factor with the number of hash buckets.
Prime numbers are used because in the _general_ case we expect to get more even distribution with a prime number; but that may not hold in _all_ cases.
For Java hash table, as long as the number is _relatively prime_ to all powers of 2 from 1-32, it is effectively prime for this purpose.
Any _candidate_ Mersenne prime (they don't have to _actually_ be prime) should meet the relevant criteria.
In this case, 255 is a _candidate_ Mersenne prime that happens to not be prime (though the nearby Fermat number, 257, _is_ prime).

Did we test with 257?  I wonder if that would have produced a similarly high number of distinct hash values.

jsync-swirlds (approved):

pbj-core/pbj-runtime/src/test/java/com/hedera/pbj/runtime/io/buffer/BytesTest.java
```diff
@@ -1024,4 +1026,43 @@ void worksWithSubSlices() {
             assertTrue(slice.contains(needle));
         }
     }
+
+    public static void main(String[] args) {
```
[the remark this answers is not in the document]
artemananiev: Removed

jsync-swirlds (approved):

PR #590, opened 2025-08-27
test: cover PbjGrpcClient

### LINKED ISSUE #589: Cover PbjGrpcClient

anthony-swirldslabs:
**Description**:
Covering `PbjGrpcClient` with unit tests.

There's only two factory methods that are impossible to cover because they call deep into Helidon classes, including their package-private methods, which is impossible/too-very-difficult to mock (it would require using PowerMock, which doesn't support Java modules.)

This is fine because there's not any logic inside those factory methods. So we're not missing a lot.

**Related issue(s)**:

Fixes #589 

**Notes for reviewer**:
All tests should pass.

**Checklist**

### DIFF
--- a/pbj-core/pbj-grpc-client-helidon/src/test/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClientTest.java
+++ b/pbj-core/pbj-grpc-client-helidon/src/test/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClientTest.java
@@ -47,7 +47,6 @@ public class PbjGrpcClientTest {
 
         final PbjGrpcClient client = new PbjGrpcClient(webClient, config);
 
-        assertNotNull(client);
         assertEquals(webClient, client.getWebClient());
         assertEquals(http2Client, client.getHttp2Client());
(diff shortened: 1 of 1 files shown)


pbj-core/pbj-grpc-client-helidon/src/test/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClientTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Test
+    void testConstructorAndGetters() {
+        final PbjGrpcClientConfig config =
+                new PbjGrpcClientConfig(READ_TIMEOUT, tls, OPTIONS.authority(), OPTIONS.contentType());
+        doReturn(http2Client).when(webClient).client(Http2Client.PROTOCOL);
+
+        final PbjGrpcClient client = new PbjGrpcClient(webClient, config);
+
+        assertNotNull(client);
```
imalygin, line 50: I don't think that this assert is necessary

imalygin (approved):

imalygin (approved):

PR #618, opened 2025-09-18
perf: improve varint writing performance

### LINKED ISSUE #596: Revisit VarInt Performance
The performance of parsing and writing varInts has a decent size effect on Hiero Consensus Node performance as we do so much of it. We put some effort into this area in PBJ and there is `com.hedera.pbj.integration.jmh.VarIntBench` bench mark where we tried multiple approaches. I found some new articles and examples where great developers competed for best approach. We should test their findings and see what we can learn in PBJ and what performance boost we can see. 

We should also update the `VarIntBench` to test at different input lengths as some algorithms are better for small one byte encoded values while others are better at large input values. As one of the primary uses cases for varint encoding is protobuf tags and the majority of them are 1 byte we should make sure what ever solution is chosen is good for one byte while still being fast for larger values.

Reading List for Ideas:
- https://richardstartin.github.io/posts/dont-use-protobuf-for-telemetry especially the comments at bottom
- https://steinborn.me/posts/performance/how-fast-can-you-write-a-varint/
- https://github.com/astei/varint-writing-showdown/issues/1
- https://github.com/astei/varint-writing-showdown
- https://github.com/apache/kafka/pull/13312

anthony-swirldslabs:
**Description**:
Improving parts of varint writing code in PBJ in addition to the improvement already delivered at https://github.com/hashgraph/pbj/pull/614

Also, adding benchmarks for varints.

Most of the code is extracted from a draft PR at https://github.com/hashgraph/pbj/pull/612 with minor modifications to make it compile and to simplify running the benchmarks.

**Related issue(s)**:

Fixes #596

**Notes for reviewer**:
All tests should pass.

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
@@ -1010,12 +1010,7 @@ public final class ProtoWriterTools {
      * @return the number of bytes for encoded value
      */
     public static int sizeOfVarInt64(final long value) {
-        if (value >= 0) {
-            return sizeOfUnsignedVarInt64(value);
-        } else {
-            // Must sign-extend.
-            return MAX_VARINT_SIZE;
-        }
+        return sizeOfUnsignedVarInt64(value);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


abies:
For the sake of interest, can you please attach the results from the jhm runs?

anthony-swirldslabs:
@anies :
> For the sake of interest, can you please attach the results from the jhm runs?

Good idea. I've updated the Description of this PR with the results. Please have a look. An uber comment for the results is that we're trying to optimize the most frequent case of 1 byte varint - this is because most field tags in protobuf models only occupy 1 byte, so we want to be super efficient when processing such short varints.

PR #627, opened 2025-09-25
fix: add integ test for write(byte[]) and fix all bugs

### LINKED ISSUE #626: Integration tests for write(byte[])
The recently introduced `write(byte[])` feature should be tested in model integration tests. Any bugs found need to be fixed.

anthony-swirldslabs:
**Description**:
* Adding integration tests for the new `write(byte[])` feature to check the output against the existing `write(WritableSequentialData)` API
* Fixing all the bugs found along the way

**Related issue(s)**:

Fixes #626

**Notes for reviewer**:
All tests should pass.

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/TestGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/TestGenerator.java
@@ -284,11 +284,11 @@ public final class TestGenerator implements Generator {
             case STRING -> "STRING_TESTS_LIST";
             case BYTES -> "BYTES_TESTS_LIST";
             case ENUM -> "Arrays.asList(%s.values())".formatted(javaFieldType);
-            case ONE_OF -> throw new RuntimeException(
-                    "Should never happen, should have been caught in generateTestData()");
+            case ONE_OF ->
+                throw new RuntimeException("Should never happen, should have been caught in generateTestData()");
             case MESSAGE -> "%s%s.ARGUMENTS".formatted(javaFieldType, FileAndPackageNamesConfig.TEST_JAVA_FILE_SUFFIX);
-            case MAP -> throw new RuntimeException(
-                    "Should never happen, should have been caught in generateTestData()");
+            case MAP ->
+                throw new RuntimeException("Should never happen, should have been caught in generateTestData()");
         };
     }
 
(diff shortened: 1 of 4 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/TestGenerator.java
```diff
@@ -319,18 +319,47 @@ private static String generateTestMethod(final String modelClassName, final Stri
...
+                                ": in buffer got " + byteBuffer.array()[mismatchIndex] + ", but in array got " + bytesArray[mismatchIndex] +
+                                ".\\nBuffer:\\n" + dataBuffer.toString() +
+                                "\\nArray:\\n" + Arrays.toString(Arrays.copyOf(bytesArray, protoBufByteCount)) +
+                                "\\nModel:\\n" + modelObj +
+                                "\\nProtobuf size:\\n" + protoBufByteCount
+                        );
+                    }
+                    // This check is not precise. An IOOB exception would be best. But we don't want to create new arrays
+                    // for each test so as not to overwhelm the Java GC. So we do this instead:
+                    Arrays.parallelSort(bytesArray, protoBufByteCount, bytesArray.length);
```
jasperpotts, line 352: I don't understand the sort here. Does write(byte[]) not return number of bytes written?

anthony-swirldslabs: It does. However, the return value may be incorrect if there's any bugs in that generated code or in the write methods that can report incorrect bytesWritten if there's any bugs in them.

The sort here ensures that either the tail or the head of the remaining bytes will be non-zero if there's any non-zero bytes written by write(byte[]). This won't protect us from write(byte[]) writing zero bytes though, but this is still a lot faster than allocating a new array with a precise size for each test. I can do the latter, but then the tests will take half an hour+ instead of a couple of minutes.

anthony-swirldslabs: I see what you're saying. The parallelSort wasn't super slow - all tests took under a couple of minutes on my Mac (and about 4 minutes in GitHub: https://github.com/hashgraph/pbj/actions/runs/18016465096/job/51262602372?pr=627 ) which is somewhat comparable with what the tests took before. However, I agree that an O(n log n) sorting should be generally slower than an O(n) iteration. With a simple loop, the tests now take under a minute on my laptop. Not exactly super fast, but better than with sorting. Please see the latest commit.

PR #636, opened 2025-10-01
perf: improve strings handling

### LINKED ISSUE #620: Revisit String performance
UTF-8 encoding takes time. Many times, we don't even use strings that we parse (e.g. memo fields and similar). So PBJ may be wasting time encoding them when parsing objects.

The idea is to store strings as original `byte[]` objects in models and only perform the encoding if/when someone calls a getter on the model for the string. If the getter is never called, then we never need to waste cycles on the encoding.

A PoC of this change is available at https://github.com/hashgraph/pbj/pull/612

The task is to extract it, get it to the production level, and release the change.

anthony-swirldslabs:
**Description**:
This is the last part of changes extracted from @jasperpotts 's draft PoC https://github.com/hashgraph/pbj/pull/612 . We store string fields as UTF-8 byte arrays internally in models. We allow codecs to parse original UTF-8 and implant them directly into the model w/o performing the UTF-8 decoding (which internally would convert the text into UTF-16 (or Latin-1) because that's how Java stores strings.) The public API of the models is unchanged, and whenever anyone calls a getter for a string field, the model will encode it into a Java string on the fly.

The idea behind this optimization is that we rarely read or use strings in our business logic. So we could as well skip the decoding part when parsing models.

Three caveats come with the fix:
1. New public constructors are introduced that accept raw `byte[]` for strings and also have a fake unused argument so as to resolve a generic erasure clash. This looks a bit ugly. Also, this is not very safe because it allows a malicious code to create a mutable model instance (by retaining references to their `byte[]`) and then pass the object to a code that expects the models to be immutable. We could make the constructors `private`, but we'd have to move all the codecs into the same package with their models, which could be a breaking change.
2. If we end up reading the string multiple times, either directly or via log/toString(), either now or in the future, then this optimization will be defeated as we'll be encoding the bytes multiple times. We could work-around this by adding caching for the encoded strings, same as we do with hashCode/protobufSize already.
3. The fuzz test thresholds had to be relaxed because PBJ no longer decodes UTF-8 strings when parsing models, so it now fails less often than Protoc. Also, this means that encoding errors may now happen inside our business logic code rather than at deserialization points as before. There isn't a simple solution for this caveat.

**Related issue(s)**:

Fixes #620

**Notes for reviewer**:
All tests should pass.

**Checklist**

### DIFF
--- a/pbj-integration-tests/src/jmh/java/com/hedera/pbj/integration/jmh/utf8/Utf8ToolsV2.java
+++ b/pbj-integration-tests/src/jmh/java/com/hedera/pbj/integration/jmh/utf8/Utf8ToolsV2.java
@@ -1,0 +1,151 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.pbj.integration.jmh.utf8;
+
+import static java.lang.Character.MIN_SUPPLEMENTARY_CODE_POINT;
+
+import com.hedera.pbj.runtime.MalformedProtobufException;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.IOException;
+
+/**
+ * UTF8 tools based on protobuf standard library, so we are byte for byte identical
+ */
+public final class Utf8ToolsV2 {
+
+    /**
+     * Returns the number of bytes in the UTF-8-encoded form of {@code sequence}. For a string, this
+     * method is equivalent to {@code string.getBytes(UTF_8).length}, but is more efficient in both
+     * time and space.
+     *
+     * @throws IllegalArgumentException if {@code sequence} contains ill-formed UTF-16 (unpaired
+     *     surrogates)
+     */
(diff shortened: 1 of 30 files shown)


pbj-integration-tests/src/jmh/java/com/hedera/pbj/integration/jmh/utf8/Utf8ToolsV2.java
```diff
@@ -0,0 +1,151 @@
...
+                len += 2;
+            } else if (codePoint <= 0xFFFF) {
+                len += 3;
+            } else {
+                len += 4;
+            }
+            i += Character.charCount(codePoint);
+        }
+        return len;
+        //        if (in == null) {
```
imalygin, line 39: Do we still need this code?

imalygin (approved):

imalygin (approved):

PR #639, opened 2025-10-02
feat: Codec.parse(..., int maxSize)

### LINKED ISSUE #637: Make maxSize configurable
PBJ hard-codes a `maxSize` limit at

https://github.com/hashgraph/pbj/blob/4b876bc552a5a01e0900130153721680547d5b6a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Field.java#L26-L30

which serves to protect PBJ clients from "bad" payloads which could cause OOMs and other bad things.

However, there are use-cases that require parsing larger protobuf models.

So we want to add an `int maxSize` argument to the `Codec.parse()` method to allow client code to configure the limit on a case-by-case basis (and keep other default overloaded versions of this method use the default hard-coded constant.)

anthony-swirldslabs:
**Description**:
Introducing `Codec.parse(..., int maxSize)` that allows applications to override the default 2MB limit hard-coded in PBJ.

Note that this may be a breaking change because some code in the Consensus Node software may implement the Codec interface. So I'm bumping the version of PBJ and will be refactoring the CN code when I'm upgrading the PBJ dependency there (once this fix is released.)

Also, as a side effect, I'm also implementing the general support for `maxSize` in the JSON codecs as it was missing completely previously.

**Related issue(s)**:

Fixes #637 

**Notes for reviewer**:
All tests should pass. A new integ test is added to verify the feature.

**Checklist**


pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/MaxSizeTest.java
```diff
@@ -46,4 +47,26 @@ void testStringMaxSize() throws Exception {
...
+        MessageWithString.PROTOBUF.write(msg, data);
+
+        // First try the default maxSize:
+        data.resetPosition();
+        assertEquals(msg, MessageWithString.PROTOBUF.parse(data, false, false, Integer.MAX_VALUE));
+
+        // Then try a custom, large-enough maxSize:
+        data.resetPosition();
+        assertEquals(msg, MessageWithString.PROTOBUF.parse(data, false, false, Integer.MAX_VALUE, 666));
+
```
jsync-swirlds, line 66: Would it makes sense to add a case here for maxSize set to `Integer.MAX_VALUE`?

jsync-swirlds (approved):
Looks good, one possible improvement in unit test.

pbj-integration-tests/src/test/java/com/hedera/pbj/integration/test/MaxSizeTest.java
anthony-swirldslabs: Good idea! Done.

jsync-swirlds (approved):

imalygin (approved):

PR #640, opened 2025-10-03
fix: avoid unneeded byte flipping

### LINKED ISSUE #514: Fix unneeded byte flipping in bytes to int logic
Most of important platforms are little endian architectures. This means, that in PBJ UnsafeUtils.getInt(...) we are performing integer flipping after retrieving it from byte buffer. This is ok, as PBJ requires big endian logic. But then, in
```
    @Override
    public int getInt(final long offset, @NonNull final ByteOrder byteOrder) {
        return byteOrder == ByteOrder.BIG_ENDIAN ? getInt(offset) : Integer.reverseBytes(getInt(offset));
    }
```
we are performing second flip when we want to get big endian version of data.

My suggestion would be to optimize these places to do minimal amount of flips needed.

Additionally, we could write a benchmark for UnsafeUtils.getInt and make sure it is really faster than just retrieving 4 bytes from in array by hand and doing bitwise OR on them. This might be highly dependent on JVM, but it is not obvious that, especially for unaligned access, which one is faster.

lpetrovic05:
closes #514

### DIFF
--- a/pbj-core/pbj-runtime/src/jmh/java/com/hedera/pbj/runtime/io/BytesGetLong.java
+++ b/pbj-core/pbj-runtime/src/jmh/java/com/hedera/pbj/runtime/io/BytesGetLong.java
@@ -2,6 +2,7 @@
 package com.hedera.pbj.runtime.io;
 
 import com.hedera.pbj.runtime.io.buffer.Bytes;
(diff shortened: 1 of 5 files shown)


anthony-swirldslabs (approved):
LGTM. Thank you!

PR #645, opened 2025-10-07
feat: introduce WritableMessageDigest

### LINKED ISSUE #621: WritableSequentialData implementation to write to a message digest
`VirtualLeafBytes` has a method named `writeToForHashing()` to write leaf content during hashing:

```java
    public void writeToForHashing(final BufferedData out) { ... }
```

This `BufferedData` object is created in `VirtualHasher` from a thread local var, which is in turn a wrapper over another thread local, a byte array. Once a leaf is written, the byte array is then sent to a message digest.

It would be much more straightforward if there was a `WritableSequentialData` implementation to write directly to a message digest. The thread locals would go away and `writeToForHashing()` would be changed to use a `WritableSequentialData` instead of `BufferedData` almost as is.

anthony-swirldslabs:
**Description**:
Introducing `WritableMessageDigest` which is an implementation of `WritableSequentialData` that writes data to a `MessageDigest` object.

**Related issue(s)**:

Fixes #621 

**Notes for reviewer**:
New unit tests added. All should pass.

**Checklist**

### DIFF
--- a/pbj-integration-tests/src/jmh/java/com/hedera/pbj/integration/jmh/MessageDigestLargeModelBench.java
+++ b/pbj-integration-tests/src/jmh/java/com/hedera/pbj/integration/jmh/MessageDigestLargeModelBench.java
@@ -98,7 +98,7 @@ public class MessageDigestLargeModelBench {
 
     public static void main(String[] args) throws Exception {
         Options opt = new OptionsBuilder()
-                .include(MessageDigestBench.class.getSimpleName())
+                .include(MessageDigestLargeModelBench.class.getSimpleName())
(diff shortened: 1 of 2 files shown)


OlegMazurov (approved):

jasperpotts:
The implementation looks good but I wonder if it is the right solution. I assume the aim is better performance of hashing PBJ objects. It would be good to benchmark this vs the direct to byte[] writing work we did recently sent as a whole to the MessageDigest. I ponder if sending singular bytes to digest will be slow.

OlegMazurov (approved):

PR #662, opened 2025-10-24
fix: grpc client authority in headers

### LINKED ISSUE #656: Helidon gRPC client omits `:authority` pseudo-header
### Problem
`PbjGrpcCall` only adds the HTTP/2 `:authority` header when `requestOptions.authority()` is present. When the config leaves authority empty, that's the common case when relying on the base URI -> no `:authority` is sent. Upstream HTTP/2 servers (nginx, tonic) reject the request with 400 client sent invalid host header, so gRPC calls never reach the destination server.

### Expected
Client should always transmit authority info when there’s a host, e.g. default to the base URI host:port when the optional is empty, matching [RFC 9113 §8.3](https://datatracker.ietf.org/doc/html/rfc9113#name-http-control-data).

### Actual
Requests carry `:method`, `:scheme`, `:path`, but lack `:authority`, causing the proxy to log `host: ""` and terminate the stream before gRPC sees it.

### Steps to reproduce

1. Configure `PbjGrpcClientConfig` without an explicit `authority()` value (Helidon defaults).
2. Call a service through an HTTP/2 proxy such as nginx or tonic.
3. Observe repeated 400s: client sent invalid host header, `host: ""` and all gRPC traffic fails.

### Additional context

Previous client version sent a Host header and worked. New wrapper at P[bjGrpcCall.java](https://github.com/hashgraph/pbj/blob/main/pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcCall.java#L87) sets headers.add(Http2Headers.AUTHORITY_NAME, authority) only when present, leaving proxies no authority to forward. Suggest defaulting authority to the base URI host/port or always driving it from config.

### Hedera network

other

### Version

v0.12.1

### Operating system

Linux

georgi-l95:
**Description**:
This PR fixes the [issue](https://github.com/hashgraph/pbj/issues/656) reported in this issue, regarding missing authority attribute in headers.

**Related issue(s)**:

Fixes #656 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/pbj-integration-tests/src/main/java/com/hedera/pbj/integration/grpc/GrpcTestUtils.java
+++ b/pbj-integration-tests/src/main/java/com/hedera/pbj/integration/grpc/GrpcTestUtils.java
@@ -38,9 +38,8 @@ public class GrpcTestUtils {
                 WebClient.builder().baseUri("http://localhost:" + port).tls(tls).build();
 
         // If the requestOptions doesn't have an authority, provide one based on the port
-        final Optional<String> authority = requestOptions.authority().isPresent()
-                ? requestOptions.authority()
-                : Optional.of("localhost:" + port);
+        final Optional<String> authority =
+                requestOptions.authority().isPresent() ? requestOptions.authority() : Optional.of("localhost:" + port);
 
         final PbjGrpcClientConfig config =
                 new PbjGrpcClientConfig(READ_TIMEOUT, tls, authority, requestOptions.contentType());


pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClient.java
```diff
@@ -145,11 +154,7 @@ public PbjGrpcClientConfig getConfig() {
     /** Simple implementation of the {@link ServiceInterface.RequestOptions} interface. */
     private record Options(Optional<String> authority, String contentType) implements ServiceInterface.RequestOptions {}
 
-    private ClientConnection createClientConnection() {
-        final ClientUri clientUri = ((Http2ClientImpl) http2Client)
-                .prototype()
-                .baseUri()
-                .orElseThrow(() -> new IllegalStateException("No base URI provided in the WebClient."));
+    private ClientConnection createClientConnection(final ClientUri clientUri) {
```
anthony-swirldslabs: Minor: couldn't it use the `this.baseUri` to avoid adding an argument?

anthony-swirldslabs (commented):
LGTM overall. However, please fix `spotlessApply` per the failing PR check, and also, please address the version.txt comment. Thanks!

pbj-core/version.txt
```diff
@@ -1 +1 @@
-0.12.0-SNAPSHOT
+0.13.0
```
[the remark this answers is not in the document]
georgi-l95: Sorry, didn't meant to commit this change, was local for testing only. Reverted 👍

pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClient.java
georgi-l95: You're right, applied it.

PR #707, opened 2025-12-13
fix: use @Feature annotation instead of @Features

ivanguravel:
**Description**:

Fix `IllegalStateException` on Block Node startup caused by invalid Helidon feature metadata.

The `pbj-grpc-helidon-config` module uses `@Features.*` annotations which don't generate the required `m` (module name) field in `feature-metadata.properties`. This causes Helidon 4.3.2 to throw an exception during feature scanning.

* Replace `@Features.Preview` with `@Preview` annotation
* Replace `@Features.Name`, `@Features.Description`, `@Features.Flavor`, `@Features.Path` with consolidated `@Feature` annotation (which is old, but works)
* Match annotation pattern already used in `pbj-grpc-helidon` module

**Related issue(s)**:

Fixes https://github.com/hiero-ledger/hiero-block-node/issues/1964

**Notes for reviewer**:

Before fix - `feature-metadata.properties` was missing `m` field:
```
n=PBJConfig
d=WebServer gRPC-PBJ Config
```

After fix - `feature-metadata.properties` contains all required fields:
```
m=com.hedera.pbj.grpc.helidon.config
n=PBJConfig
d=WebServer gRPC-PBJ Config
in=SE
p=WebServer,PBJpr=true
```

Checklist
- [ ] Documented (Code comments, README, etc.) - probably don't need

Снимок экрана 2025-12-13 в 18 25 55

### DIFF
--- a/pbj-core/pbj-grpc-helidon-config/src/main/java/module-info.java
+++ b/pbj-core/pbj-grpc-helidon-config/src/main/java/module-info.java
@@ -1,12 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
-import io.helidon.common.features.api.Features;
+import io.helidon.common.features.api.Feature;
 import io.helidon.common.features.api.HelidonFlavor;
+import io.helidon.common.features.api.Preview;
 
-@Features.Preview
-@Features.Name("PBJConfig")
-@Features.Description("WebServer gRPC-PBJ Config")
-@Features.Flavor(HelidonFlavor.SE)
-@Features.Path({"WebServer", "PBJ"})
+@Preview
+@Feature(
+        value = "PBJConfig",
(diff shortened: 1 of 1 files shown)


ivanguravel:
@akugal  no, but I will try

ivanguravel:
Hello @anthony-swirldslabs 

In terms of the following info from the helidon side https://github.com/helidon-io/helidon/issues/10936#issuecomment-3675113912 I'm wondering if I should find another way how to resolve that issue with the config exception ?

anthony-swirldslabs:
@ivanguravel : Yes, that would be great. Per the reply from Helidon, it sounds like the best approach would be to switch back to the new annotations (that we used before this PR) **AND** replace the old processor with the new codegen module. It seems like we just were missing this second part in order to make the new annotations work.

Issue #520, opened 2025-06-13
GRPC: support compression


anthony-swirldslabs:
Currently, PBJ GRPC implementation doesn't support compression. When sending datagrams, it passes `0` as a compression flag, and when receiving messages, it only supports `identity` encoding (meaning no compression.) This applies to both the Helidon service plugin in PBJ as well as the upcoming PBJ GRPC client implementation.

We want to support compression and have a way to enable it when using PBJ APIs for GRPC so as to reduce the network traffic as well as be compatible with 3rd party GRPC clients and services.

anthony-swirldslabs:
Based on https://github.com/grpc/grpc/issues/26460 , GRPC doesn't officially support zstd or anything other than gzip for that matter. This is true for all languages including the core C GRPC implementation. However, separate Java and Go GRPC implementations (that don't rely on the core C implementation) provide experimental support for custom compressors/decompressors that applications can implement and install at runtime.

Also, a client can technically list several supported compression algorithms via its `grpc-accept-encoding` header, e.g. `"identity,gzip,zstd"`. A server can also support several compression algorithms at once. As such, we don't have to make a specific choice upfront but can instead add support for other compression algorithms as that becomes feasible.

Given that `zstd` would ultimately require adding native code, we could start with the JDK built-in GZip support first, which would also be the most compatible solution for 3rd-party clients. Once we migrate to Java 25 and really feel that we need a better compression algorithm, then we can consider adding support for zstd as well, with an understanding that it's likely to only be supported when using PBJ on both the client and server side. We can make this future task more or less simple by designing the initial support for gzip to be extensible and generic enough to support multiple algorithms.

PR #710, opened 2025-12-17
feat: support compression in PbjGrpcClient

### LINKED ISSUE #709: PbjGrpcClient: support compression
Support compression in the PbjGrpcClient.

anthony-swirldslabs:
**Description**:
Introducing support for compression in PBJ GRPC client.
1. The new `GrpcCompression` class implements a registry of known compressors/decompressors. With this fix, support for identity and gzip is added. This class will also be used by the PBJ GRPC Server compression in the future.
2. The `PbjGrpcClientConfig` is extended to allow specifying outgoing and incoming compression via the `encoding` and `acceptEncodings` parameters. A backward-compatibility ctor is added to avoid breaking existing code.
3. The `PbjGrpcCall` is extended to support compression for both outgoing and incoming traffic.
4. Finally, a new integration `GrpcCompressionTest` is added to verify the behavior.

As mentioned in the test, it was impossible to force the Google GRPC server to send replies with "gzip" encoding. However, a debugging println in our client implementation shows gzip'd request bytes when the test runs, and Google happily accepts our gzip'd request. Now that we have a client that is able to use "gzip", a future fix that will introduce compression in the PBJ GRPC Server will be able to fully test the compression for incoming messages in our client as well. For now we have what we have.

In the future we may explore adding additional compressors support, such as "zstd", or better gzip implementations. With the current structure of `GrpcCompression` this should be trivial. But that is outside of scope of this fix. See https://github.com/hashgraph/pbj/issues/520#issuecomment-3644314720 for details.

**Related issue(s)**:

Fixes #709

**Notes for reviewer**:
All tests, including the new integ test for the compression, should pass.

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/GrpcCompression.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/grpc/GrpcCompression.java
@@ -108,8 +108,9 @@ public class GrpcCompression {
     }
 
     /**
-     * A utility method to help process "grpc-accept-encoding" header.
-     * @param encodingList all values of the "grpc-accept-encoding" header as received from remote peer
+     * A utility method to help process "grpc-encoding" header. Note that it should normally contain just a single
+     * value. Multiple values result in an exception. No values default to the "identity" decompressor.
+     * @param encodingList all values of the "grpc-encoding" header as received from remote peer
      * @return a Decompressor that should be used to decompress bytes received from the peer
      * @throws IllegalStateException if the Decompressor cannot be determined
      */


pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcCall.java
```diff
@@ -173,13 +192,23 @@ private void receiveRepliesLoop() {
...
+                    PbjGrpcDatagramReader.Datagram datagram;
+                    while ((datagram = datagramReader.extractNextDatagram()) != null) {
+                        if (datagram.compressedFlag() != 0 && datagram.compressedFlag() != 1) {
+                            throw new IllegalStateException("GRPC datagram compressed flag " + datagram.compressedFlag()
+                                    + " is unsupported. Only 0 and 1 are valid.");
+                        }
+                        BufferData data = datagram.data();
                         final byte[] array = data.readBytes();
                         final Bytes bytes = Bytes.wrap(array);
+                        // If the compressedFlag is 0, then per the specification, the message isn't compressed
```
artemananiev: It makes sense to check the opposite: the flag is set, but decompressor is null (or IDENTITY). Today it will just throw an NPE at line 208 below, but it can be changed to a more useful exception

PR #716, opened 2025-12-20
fix: switch to helidon-common-features-codegen for @Features metadata

ivanguravel:
**Description**:
Use the new helidon-common-features-codegen annotation processor with @Features.* annotations as recommended by Helidon developers.

This generates feature-registry.json (new format) instead of feature-metadata.properties, which is the recommended approach for Helidon 4.x and will be required in Helidon 5.x.

Changes:
* Use @Features.* annotations instead of deprecated @Feature
* Replace helidon-common-features-processor with codegen
* Add helidon-common-features-codegen dependency
* Add module mapping in gradle/modules.properties

Ref: https://github.com/helidon-io/helidon/issues/10936
Fixes: https://github.com/hiero-ledger/hiero-block-node/issues/1964

**Related issue(s)**:  https://github.com/hiero-ledger/hiero-block-node/issues/1964 

**Notes for reviewer**:
Used the original approach which is about Helidon developer advice.

**Checklist**
Verified the build and verified https://github.com/hiero-ledger/hiero-block-node/issues/1964 
Снимок экрана 2025-12-20 в 23 57 30

- [ ] Documented (Code comments, README, etc.) - probably this step is not required for these changes

### DIFF
--- a/pbj-core/gradle/modules.properties
+++ b/pbj-core/gradle/modules.properties
@@ -1,0 +1,3 @@
+# SPDX-License-Identifier: Apache-2.0
+# Module to Maven artifact mappings for JPMS
+io.helidon.common.features.codegen=io.helidon.common.features:helidon-common-features-codegen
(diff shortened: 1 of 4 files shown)


anthony-swirldslabs (approved):
LGTM. Thank you!

PR #763, opened 2026-03-26
fix: improve sendPing() in GRPC client

### LINKED ISSUE #758: GRPC streaming client may kill itself
As described at https://github.com/hashgraph/pbj/issues/746#issuecomment-4078049881 , a streaming PBJ GRPC server may sometimes encounter a `Flow control update wait time-out` exception which effectively kills the server. This happens after it has served multiple requests and has streamed multiple replies.

To reproduce: in the TestBlockGrpcBench, change `Level.Invocation` to `Level.Trial` and run the streaming benchmark. To see the exception trace w/o enabling Helidon logging, add an `onError` implementation at TestBlockStreamerService.java .

anthony-swirldslabs:
**Description**:
Pings are an essential part of the GRPC (and HTTP2) protocol that are used to check if a connection is still alive. The existing `sendPing()` calls in the PBJ GRPC Client delegated to the Helidon's `Http2ClientStream.sendPing()` method, and its implementation apparently has issues. You can find more details in a code comment, but long story short, sending a second ping request is guaranteed to destroy the HTTP2 connection to the server. This could be reproduced with both PBJ and Google GRPC servers.

1. In this fix I'm implementing our own `sendPing()` method in the `PbjGrpcCall` on the client side.
2. I'm also adding a stress test that causes the PBJ GRPC client to start producing many pings, and verifies that the application runs smoothly and the connection to the server doesn't get destroyed.
3. I'm also adding support for Helidon logging in PBJ integration tests that was helpful in debugging this issue.

Previously, this issue prevented us from using the `Level.Trial` in GRPC benchmarks - we had to create new server/client instances on ever benchmark `Level.Invocation` instead. This makes the benchmarks take longer to run than desired. With this fix, we'll be able to construct server/client instances at the `Level.Trial` instead so that we use a single instance for all the iterations of the same benchmark. This will be done in a separate fix in the future.

**Related issue(s)**:

Fixes #758 

**Notes for reviewer**:
The new stress test can be run via its `main()` to verify the fix.

**Checklist**

### DIFF
--- a/pbj-integration-tests/build.gradle.kts
+++ b/pbj-integration-tests/build.gradle.kts
@@ -169,7 +169,7 @@ testing {
         targets.named("test") {
             testTask {
                 useJUnitPlatform { excludeTags("FUZZ_TEST") }
-                // dependsOn(tasks.named("fuzzTest"))
+                dependsOn(tasks.named("fuzzTest"))
             }
         }
         dependencies {


jsync-swirlds (approved):
First, thank you for finding and fixing this issue, it sounds like there is a bug in the upstream codebase.
Second, have we filed an issue with the upstream Helidon describing the cause and concerns?

PR #776, opened 2026-04-06
Updates and extends PBJ documentation

### LINKED ISSUE #773: Update PBJ Documentation
Fill out PBJ documentation for users and PBJ developers. Both for humans and for AI.

jasperpotts:
Fixes #773 updating all the PBJ documentation

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -318,7 +318,7 @@ ReadableSequentialData ───► Model object
 
 ### Immutable Model Objects
 
-Generated model classes are immutable — all fields are `final`, there are no setters, and construction goes through a `Builder`. This enables safe sharing across threads without synchronization, which is critical in a multi-threaded consensus node.
+Generated model classes are immutable — all fields are `final`, there are no setters, and construction goes through a `Builder`. This enables safe sharing across threads without synchronization, which is critical in multi-threaded code.
 
 ### Separation of Concerns
 
--- a/docs/usage-guide.md
+++ b/docs/usage-guide.md
@@ -229,14 +229,15 @@ Empty repeated fields return `Collections.emptyList()`, never `null`.
 
 ### Map Fields
 
-Map fields use `PbjMap<K, V>` with deterministic key ordering:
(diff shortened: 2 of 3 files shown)


docs/architecture.md
```diff
@@ -0,0 +1,355 @@
...
+ReadableSequentialData ───► Model object
+                            (unknown fields optionally preserved,
+                             missing fields → null)
+```
+
+## Key Design Decisions
+
+### Immutable Model Objects
+
+Generated model classes are immutable — all fields are `final`, there are no setters, and construction goes through a `Builder`. This enables safe sharing across threads without synchronization, which is critical in a multi-threaded consensus node.
```
anthony-swirldslabs: Minor: `consensus node` comes out of the blue. Might want to `s/consensus node/code/` to make the statement more generic.

docs/usage-guide.md
```diff
@@ -0,0 +1,784 @@
...
+}
+```
+
+Register it at application startup:
+
+```java
+import com.hedera.pbj.grpc.common.compression.ZstdGrpcTransformer;
+
+// Register with default compression level (3)
+new ZstdGrpcTransformer().register("zstd");
```
anthony-swirldslabs: Note that this is registered by default by both the client and server code.

docs/usage-guide.md
```diff
@@ -0,0 +1,784 @@
...
+            "package.ServiceName/MethodName",
+            SubscribeStreamRequest.PROTOBUF,    // request codec
+            SubscribeStreamResponse.PROTOBUF,   // response codec
+            responsePipeline,                    // handles responses
+            Map.of());                           // metadata headers
+
+// Send the request
+call.sendRequest(request, true);  // true = end of stream (unary)
+```
+
```
anthony-swirldslabs: This is a very low level code that I'd suggest not to advertise, other than for super rare cases where more control is needed. PBJ exposes client stubs in the generated service interfaces. The whole code above can basically be collapsed to

`new MyServiceInterface.MyServiceClient(grpcClient, options).methodName(request);`

Client code should rarely if ever deal with the GrpcCall objects directly.

PR #816, opened 2026-05-08
feat: cache frequently parsed objects

### LINKED ISSUE #814: Consider caching when parsing
There's a use-case for small same objects being parsed many times. For example, AccountID objects for accounts receiving fees are the same, and they participate in almost every transaction.

Currently, PBJ creates a new instance of such small objects every time it parses one. This increases the pressure on GC as it has to free those objects.

The idea is to cache frequently used objects and return a cached instance instead of creating a new object each time.

anthony-swirldslabs:
**Description**:
Caching frequently parsed objects. Goals of the new feature:
* No performance impact to any existing code.
* When enabled explicitly on a particular protobuf model, while a performance gain would be nice, but it isn't a goal. The goal is to reduce pressure on GC by avoiding creating new instances of the same object.

By default, everything remains as is today - w/o any caches. A protobuf model can be annotated with
```
// <<<pbj.cacheable = 16>>>
message ... {...
```
to enable a cache for this specific model. `16` is the size of the cache. The actual size will be the next (or equal) power of two. In this particular example, it will be 16. The cache size can also be overridden at runtime by defining a Java system property `pbj.cache.fully.qualified.model.name` with the desired size. The fqn refers to the protobuf model fqn (including the protobuf's `package` directive), rather than the model Java class fqn. Note that the model must be annotated at the source code in order to support caching. Specifying the runtime system property alone won't enable the cache if it wasn't enabled in the source code of the model in the first place (because we don't regenerate codecs at runtime.)

When enabled, the generated Protobuf codec will add a cache implementation in the form of an array with a size of the next (or equal) power of two. The array will be indexed using a mask of the PBJ's hashCode for that specific model. This array-based implementation proved to be a lot more performant than a Map-based cache, which in fact caused a severe performance degradation in the original implementation.

When the cache isn't enabled (which is the current default behavior), then the parser ends with:
```java
return new $modelClassName($fieldsList);
```

When the cache is enabled, then that line will be replaced with (pseudo-code):
```java
objectHashCode = { computePBJHashCode($fieldsList); }
obj = CACHE[objectHashCode & MASK];
if (obj != null) {
    if ({ obj.equals($fieldsList) }) {
        return obj;
    }
}
obj = new $modelClassName($fieldsList, objectHashCode);
CACHE[objectHashCode & MASK] = obj;
return obj;
```

Note that we preserve the computed hashCode so that we don't have to recompute it again in the future.

The hashCode and equals generators were modified to be able to use their implementations directly in the body of the parser and use the local `temp_` variables that contain the parsed data w/o having to call any extra methods or create any temporary objects.

**Performance testing**:
A `NotCacheableAccountID` and a `CacheableAccountID` models are added which are replicas of the existing `AccountID`. The cacheable one is annotated with `<<<pbj.cacheable = 16>>>`. A benchmark is added to test parsing for these two models with different number of repeated objects and different periods between the repeated objects. The benchmark was run on a real hardware and here's the results:

https://github.com/hashgraph/pbj/actions/runs/25697025584/job/75447727577
```
Benchmark             (numOfFrequentModels)  (period)                     (type)   Mode  Cnt    Score    Error   Units
CacheableBench.bench                      1         2  NotCacheableAccountIDType  thrpt   15  307.947 ± 11.512  ops/us
CacheableBench.bench                      1         2     CacheableAccountIDType  thrpt   15  334.049 ±  3.450  ops/us
CacheableBench.bench                      1         5  NotCacheableAccountIDType  thrpt   15  265.476 ± 48.826  ops/us
CacheableBench.bench                      1         5     CacheableAccountIDType  thrpt   15  323.867 ±  8.769  ops/us
CacheableBench.bench                      1        11  NotCacheableAccountIDType  thrpt   15  320.057 ±  7.005  ops/us
CacheableBench.bench                      1        11     CacheableAccountIDType  thrpt   15  338.901 ±  1.015  ops/us
CacheableBench.bench                      1        37  NotCacheableAccountIDType  thrpt   15  324.357 ±  2.640  ops/us
CacheableBench.bench                      1        37     CacheableAccountIDType  thrpt   15  292.483 ± 49.950  ops/us
CacheableBench.bench                      2         2  NotCacheableAccountIDType  thrpt   15  306.988 ± 32.137  ops/us
CacheableBench.bench                      2         2     CacheableAccountIDType  thrpt   15  321.458 ± 12.393  ops/us
CacheableBench.bench                      2         5  NotCacheableAccountIDType  thrpt   15  288.041 ± 24.641  ops/us
CacheableBench.bench                      2         5     CacheableAccountIDType  thrpt   15  330.236 ±  1.969  ops/us
CacheableBench.bench                      2        11  NotCacheableAccountIDType  thrpt   15  314.021 ±  5.800  ops/us
CacheableBench.bench                      2        11     CacheableAccountIDType  thrpt   15  333.272 ±  5.111  ops/us
CacheableBench.bench                      2        37  NotCacheableAccountIDType  thrpt   15  320.244 ±  4.494  ops/us
CacheableBench.bench                      2        37     CacheableAccountIDType  thrpt   15  305.337 ± 54.041  ops/us
CacheableBench.bench                      3         2  NotCacheableAccountIDType  thrpt   15  321.302 ±  5.017  ops/us
CacheableBench.bench                      3         2     CacheableAccountIDType  thrpt   15  342.591 ±  0.280  ops/us
CacheableBench.bench                      3         5  NotCacheableAccountIDType  thrpt   15  327.604 ±  2.536  ops/us
CacheableBench.bench                      3         5     CacheableAccountIDType  thrpt   15  342.869 ±  0.495  ops/us
CacheableBench.bench                      3        11  NotCacheableAccountIDType  thrpt   15  327.363 ±  2.103  ops/us
CacheableBench.bench                      3        11     CacheableAccountIDType  thrpt   15  342.895 ±  0.170  ops/us
CacheableBench.bench                      3        37  NotCacheableAccountIDType  thrpt   15  325.910 ±  1.367  ops/us
CacheableBench.bench                      3        37     CacheableAccountIDType  thrpt   15  342.712 ±  0.442  ops/us
CacheableBench.bench                      7         2  NotCacheableAccountIDType  thrpt   15  325.414 ±  2.817  ops/us
CacheableBench.bench                      7         2     CacheableAccountIDType  thrpt   15  342.875 ±  0.206  ops/us
CacheableBench.bench                      7         5  NotCacheableAccountIDType  thrpt   15  327.310 ±  2.734  ops/us
CacheableBench.bench                      7         5     CacheableAccountIDType  thrpt   15  330.587 ± 18.594  ops/us
CacheableBench.bench                      7        11  NotCacheableAccountIDType  thrpt   15  326.389 ±  2.393  ops/us
CacheableBench.bench                      7        11     CacheableAccountIDType  thrpt   15  330.704 ± 18.903  ops/us
CacheableBench.bench                      7        37  NotCacheableAccountIDType  thrpt   15  305.261 ± 33.142  ops/us
CacheableBench.bench                      7        37     CacheableAccountIDType  thrpt   15  342.904 ±  0.169  ops/us
CacheableBench.bench                     17         2  NotCacheableAccountIDType  thrpt   15  327.299 ±  1.057  ops/us
CacheableBench.bench                     17         2     CacheableAccountIDType  thrpt   15  342.767 ±  0.228  ops/us
CacheableBench.bench                     17         5  NotCacheableAccountIDType  thrpt   15  326.984 ±  2.431  ops/us
CacheableBench.bench                     17         5     CacheableAccountIDType  thrpt   15  340.367 ±  0.848  ops/us
CacheableBench.bench                     17        11  NotCacheableAccountIDType  thrpt   15  327.641 ±  1.187  ops/us
CacheableBench.bench                     17        11     CacheableAccountIDType  thrpt   15  342.710 ±  0.462  ops/us
CacheableBench.bench                     17        37  NotCacheableAccountIDType  thrpt   15  327.699 ±  0.885  ops/us
CacheableBench.bench                     17        37     CacheableAccountIDType  thrpt   15  342.426 ±  0.567  ops/us
```

It looks like the `CacheableAccountIDType` is parsed some ~6% faster than the `NotCacheableAccountIDType`, even though speeding up parsing wasn't the goal of the fix.

**Related issue(s)**:

Fixes #814 

**Notes for reviewer**:
All tests should pass.

**Checklist**


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/Common.java
```diff
@@ -216,97 +216,118 @@ public static String javaPrimitiveToObjectType(String primitiveFieldType) {
...
+                               result = 31 * result + Integer.hashCode(EnumWithProtoMetadata.protoOrdinal($prefixFieldName$fieldName));
                             }
                             """)
-                                    .replace("$fieldName", f.nameCamelFirstLower());
+                            .replace("$modelClassName", modelClassName)
+                            .replace("$prefixFieldName", fieldNamePrefix)
+                            .replace("$suffix", suffix)
+                            .replace("$fieldName", f.nameCamelFirstLower());
                 } else if (f.type() == Field.FieldType.MAP) {
                     generatedCodeSoFar += getMapHashCodeGeneration(generatedCodeSoFar, f);
```
imalygin: if I understand correctly, in case of cacheable objects `hashCode` will be generated with `temp_` prefix. But the template of `getMapHashCodeGeneration` still references `$fieldName` directly.

So, I suspect that for any cacheable message that contains a map field, the generated Codec would reference an undefined variable and fail to compile.

anthony-swirldslabs: Good catch! Thanks! I fixed it in the latest commit.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecGenerator.java
```diff
@@ -66,6 +67,45 @@ public void generate(
...
+            //   mask = 0b111 aka 7
+            cacheableSupport = """
+                    /** Cache for parsed objects to avoid creating new instances. Access is not synchronized for performance reasons by design. */
+                    private static final $modelClass[] CACHE;
+                    private static final int CACHE_KEY_MASK;
+                    static {
+                        final String property = System.getProperty("pbj.cache.$fqn");
+                        final int capacity = (property != null && !property.isBlank()) ? Integer.parseInt(property) : $size;
+                        // Next (or equal) power of 2:
+                        final int length = capacity <= 0 ? 1 : ((capacity & (capacity - 1)) == 0 ? capacity :  Integer.highestOneBit(capacity) << 1);
```
artemananiev, line 96: `capacity <= 1 ? capacity : Integer.highestOneBit(capacity - 1) * 2`

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
```diff
@@ -596,26 +606,46 @@ public boolean equals(Object that) {
...
-                    return false;
-                }
-            } else if (thatObj.$unknownFields != null && !thatObj.$unknownFields.isEmpty()) {
-                return false;
-            }
-            return true;
-        }""".indent(DEFAULT_INDENT);
+            $equalsBody
+        }"""
+                .replace("$equalsBody", generateEqualsBody(fields, javaRecordName, ""))
```
artemananiev, line 613: The prefix can be "this." to make it a little bit clearer

artemananiev (approved):

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecGenerator.java
anthony-swirldslabs: This wouldn't cover <= 0 values of capacity (which don't make sense, but you never trust user input) resulting in a zero-sized cache which the rest of the code is unable to handle. The current code ensures that the cache has at least one slot, so that I don't have to perform length checks anywhere else.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/ModelGenerator.java
anthony-swirldslabs: A few places have logic based on "prefix.isBlank()", so I do want to keep it empty here when generating a real equals() method rather than its inline implementation.

imalygin (approved):

PR #827, opened 2026-05-14
perf: default stmt to func

ldintr:
**Description**:

Moves default case into its own function so optimizer can choose what to inline

Prev bench numbers

Benchmark (type) Mode Cnt Score Error Units
GenericParserQuickBench.bench NotCacheableAccountIDType thrpt 15 292.432 ± 14.627 ops/us
https://github.com/hashgraph/pbj/actions/runs/25826669514/job/75881464290

New

Default Case Commit

Benchmark                                         (type)   Mode  Cnt    Score   Error   Units
GenericParserQuickBench.bench  NotCacheableAccountIDType  thrpt   15  446.331 ± 7.801  ops/us
https://github.com/hashgraph/pbj/actions/runs/25863069512/job/75997769734

Some additional cases as func from generateFieldCaseStatementPacked

Benchmark                                         (type)   Mode  Cnt    Score   Error   Units
GenericParserQuickBench.bench  NotCacheableAccountIDType  thrpt   15  450.417 ± 2.910  ops/us
https://github.com/hashgraph/pbj/actions/runs/25867304244/job/76012761470  
  
  
NOTE: Hash is different because I ran this before applying spotlight

Benchmark                                         (type)   Mode  Cnt    Score   Error   Units
GenericParserQuickBench.bench  NotCacheableAccountIDType  thrpt   15  451.509 ± 5.192  ops/us
https://github.com/hashgraph/pbj/actions/runs/25921616637/job/76191842411

**Checklist**

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
@@ -48,12 +48,14 @@ class CodecParseMethodGenerator {
     }
 
     static String generateParseMethod(
+            StringBuilder sbFuncMain,
             final String modelClassName,
             final String schemaClassName,
             final List<Field> fields,
             final boolean isCacheable) {
 
-        var parseLoopLs = generateParseLoop(generateCaseStatements(fields, schemaClassName), "", schemaClassName);
+        var parseLoopLs =
+                generateParseLoop(generateCaseStatements(sbFuncMain, fields, schemaClassName), "", schemaClassName);
         // spotless:off
         var retString = """
                 /**
@@ -263,20 +265,20 @@ if ($prefixf == null) {
      * @param fields list of all fields in record
      * @return string of case statement code
      */
-    private static String generateCaseStatements(final List<Field> fields, final String schemaClassName) {
+    private static String generateCaseStatements(StringBuilder sbFunc, List<Field> fields, String schemaClassName) {
         StringBuilder sb = new StringBuilder();
         for (Field field : fields) {
             if (field instanceof final OneOfField oneOfField) {
                 for (final Field subField : oneOfField.fields()) {
-                    generateFieldCaseStatement(sb, subField, schemaClassName);
+                    generateFieldCaseStatement(sb, sbFunc, subField, schemaClassName);
                 }
             } else if (field.repeated() && field.type().wireType() != Common.TYPE_LENGTH_DELIMITED) {
                 // for repeated fields that are not length encoded there are 2 forms they can be stored in file.
                 // "packed" and repeated primitive fields
-                generateFieldCaseStatement(sb, field, schemaClassName);
-                generateFieldCaseStatementPacked(sb, field);
+                generateFieldCaseStatement(sb, sbFunc, field, schemaClassName);
+                generateFieldCaseStatementPacked(sb, sbFunc, field);
             } else {
-                generateFieldCaseStatement(sb, field, schemaClassName);
+                generateFieldCaseStatement(sb, sbFunc, field, schemaClassName);
             }
         }
         return sb.toString().indent(DEFAULT_INDENT * 4);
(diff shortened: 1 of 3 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -47,12 +48,16 @@ static String generateUnsetOneOfConstants(final List<Field> fields) {
     }
 
     static String generateParseMethod(
+            StringBuilder sbFuncMain,
             final String modelClassName,
             final String schemaClassName,
             final List<Field> fields,
             final boolean isCacheable) {
+
+        var parseLoopLs =
```
anthony-swirldslabs: Minor: please replace all the new usages of `var` in this PR with actual concrete types. This makes the code easier to read, especially when browsing it on GitHub or in a non-IDE text editor.

Also, it's unclear what "Ls" stands for. It would be best to find a more descriptive name for this variable. Perhaps s/Ls/Parts/ or something similar?

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -105,6 +110,10 @@ static String generateParseMethod(
                         throw new ParseException(anyException);
                     }
                 }
+                List<UnknownField> defaultCase(int tag, int field, FieldDefinition f, boolean strictMode, boolean parseUnknownFields, List<UnknownField> $unknownFields, ReadableSequentialData input, int maxSize) throws ParseException, IOException {
```
anthony-swirldslabs: Please add a blank line between the two generated methods for readability of the generated code. Also, you probably want to make the new method private because there's no reason for it to be package-private.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -117,14 +126,16 @@ static String generateParseMethod(
...
+        .replace("$parseLoop", parseLoopLs.get(0))
+        .replace("$defaultCaseBody", parseLoopLs.get(1))
         .replace("$listFieldsWriteProtection", fields.stream()
                 .filter(Field::repeated)
                 .map(field -> "if (temp_" + field.name() + " instanceof UnmodifiableArrayList ual) ual.makeReadOnly();")
                 .collect(Collectors.joining("\n"))
                 .indent(DEFAULT_INDENT * 2))
         .indent(DEFAULT_INDENT);
         // spotless:on
+        return retString;
```
anthony-swirldslabs: Minor: it appears that you could keep the `return` statement at line 60 above instead of introducing a new variable.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -160,10 +171,12 @@ static String generateCacheableSupport(String modelClassName, final List<Field>
     }
 
     // prefix is pre-pended to variable names to support a nested parsing loop.
-    static String generateParseLoop(
+    // The list returned is [$parseLoop, $defaultCaseBody]
+    static List<String> generateParseLoop(
```
anthony-swirldslabs: Minor: instead of a generic list, you could introduce a `record ParseLoop(String loopBody, String defaultCaseBody) {}` and return it instead. This way we don't need to deal with the generic list and don't have to remember its indices for various parts of the parsing loop code, which will make the code easier to maintain.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -191,54 +204,58 @@ static String generateParseLoop(
...
                             }
                         }
-                """
-                .replace("$caseStatements",caseStatements)
-                .replace("$prefix",prefix)
-                .replace("$schemaClassName",schemaClassName)
+""");
+        list.add("""
+// The wire type is the bottom 3 bits of the byte. Read that off
+final int wireType = $prefixtag & TAG_WIRE_TYPE_MASK;
```
anthony-swirldslabs: Minor: I suggest to indent all the lines inside the `"""/"""` similar to how the rest of the code around here looks like. This will make it more readable.

anthony-swirldslabs (commented):
The performance numbers look great.

The code looks good overall, too. I added a few comments, mostly of stylistic nature about both the generator and the generated code.

Also, it'd be useful to attach a copy of a generated codec for reviewers to examine the changes (you can attach it to a GitHub comment in this PR.) A good candidate would be the `EverythingProtoCodec.java` file as it would show almost everything there's in PBJ.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -48,16 +48,16 @@ static String generateUnsetOneOfConstants(final List<Field> fields) {
...
-            StringBuilder sbFuncMain,
+            StringBuilder sbFunc,
             final String modelClassName,
             final String schemaClassName,
             final List<Field> fields,
             final boolean isCacheable) {
 
-        var parseLoopLs =
-                generateParseLoop(generateCaseStatements(sbFuncMain, fields, schemaClassName), "", schemaClassName);
+        var parseAndDefaultBodyPair =
```
anthony-swirldslabs: s/var/ParseAndDefaultBody/

Also, as per https://github.com/hashgraph/pbj/pull/827#pullrequestreview-4292173715 , please attach the generated `EverythingProtoCodec.java` to a comment in this PR. Thanks!

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -459,7 +454,7 @@ private static void generateFieldCaseStatement(
             // However(!), we read the key and value fields explicitly to avoid creating temporary entry objects.
             final MapField mapField = (MapField) field;
             final List<Field> mapEntryFields = List.of(mapField.keyField(), mapField.valueField());
-            var parseLoopLs = generateParseLoop(
+            var parseAndDefaultBodyPair = generateParseLoop(
```
anthony-swirldslabs: s/var/ParseAndDefaultBody/

ldintr:
Fixed. Here's the before and after of EverythingProtoCodec.java
[after.java](https://github.com/user-attachments/files/27812562/after.java)
[before.java](https://github.com/user-attachments/files/27812563/before.java)
[diff.patch](https://github.com/user-attachments/files/27812716/diff.patch)

anthony-swirldslabs (approved):

PR #832, opened 2026-05-19
fix: Allow JSON parsing of snake_case field names

### LINKED ISSUE #831: Allow JSON parsing of snake_case field names.
### Description

The PBJ compiler converts snake_case fields to camelCase fields in the protobuf object. When parsing JSON, it assumes that the JSON field name is camelCase. The snake_case field name does not match the camelCase field name, and the field value is always blank.

### Steps to reproduce

add this test case to JsonCodecTest.java. It will fail.

    @Test
    public void SnakeCaseStringTest() throws Exception {
        final String json = """
                {
                  "bytesField": "308201a2300d06092a86",
                  "string_field": "snake_case"
                }
                """;
        MessageWithBytesAndString bytesAndString = MessageWithBytesAndString.JSON.parse(Bytes.wrap(json));
        assertNotEquals(0, bytesAndString.bytesField().length());
        assertEquals("snake_case", bytesAndString.stringField());
    }

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.15.4

### Operating system

macOS

berryware:
____________________________________________________________________

**Description**:
- convert json field name to camelCase before matching to protobuf object field name

**Related issue(s)**:
Fixes #831

**Notes for reviewer**:

### DIFF
--- a/pbj-integration-tests/src/main/proto/bytesAndString.proto
+++ b/pbj-integration-tests/src/main/proto/bytesAndString.proto
@@ -13,5 +13,5 @@ option java_multiple_files = true;
  */
 message MessageWithBytesAndString {
   bytes bytesField = 1;
-  string stringField = 2;
+  string string_field = 2;
 }
--- a/pbj-integration-tests/src/main/proto/bytesAndString.proto
+++ b/pbj-integration-tests/src/main/proto/bytesAndString.proto
@@ -13,5 +13,5 @@ option java_multiple_files = true;
  */
 message MessageWithBytesAndString {
   bytes bytesField = 1;
-  string string_field = 2;
+  string stringField = 2;
 }
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/json/JsonCodecParseMethodGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/json/JsonCodecParseMethodGenerator.java
@@ -77,7 +77,7 @@ class JsonCodecParseMethodGenerator {
                         // -- EXTRACT VALUES FROM PARSE TREE ---------------------------------------------
 
                         for (JSONParser.PairContext kvPair : root.pair()) {
-                            switch (kvPair.STRING().getText()) {
+                            switch (toJsonFieldName(kvPair.STRING().getText())) {
                                 $caseStatements
                                 default: {
                                     if (strictMode) {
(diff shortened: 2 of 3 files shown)


pbj-integration-tests/src/main/proto/bytesAndString.proto
```diff
@@ -13,5 +13,5 @@ option java_multiple_files = true;
  */
 message MessageWithBytesAndString {
   bytes bytesField = 1;
-  string stringField = 2;
+  string string_field = 2;
```
anthony-swirldslabs: Is this change strictly necessary? Isn't the JSON parser supposed to silently adapt snake_case names encountered in a JSON representation into proper camelCase names as used in PBJ protobuf models?

berryware: The compiler  detects snake_case and generates protobuf objects with JSON field names.

```
message NodeEntry {
    int64 node_id = 1;
    string public_key = 2;
    TimestampRange timestamp = 3;
}
```
becomes
```
public final class NodeEntry {
    /** Protobuf codec for reading and writing in protobuf format */
    public static final Codec<NodeEntry> PROTOBUF = new org.hiero.block.node.roster.bootstrap.rsa.codec.NodeEntryProtoCodec();
    /** JSON codec for reading and writing in JSON format */
    public static final JsonCodec<NodeEntry> JSON = new org.hiero.block.node.roster.bootstrap.rsa.codec.NodeEntryJsonCodec();
    /** Default instance with all fields set to default values */
    public static final NodeEntry DEFAULT = newBuilder().build();

    /** Field <b>(1)</b>  */
    private final long nodeId;
    /** Field <b>(2)</b>  */
    private final @NonNull String publicKey;
    /** Field <b>(3)</b>  */
    private final @Nullable TimestampRange timestamp;
    /** Computed hash code, manual input ignored. */
```

The JSON codec generated for this class then uses "nodeId", and "publicKey" for matching the JSON fields, which will not match.

```
            for (JSONParser.PairContext kvPair : root.pair()) {
                switch (kvPair.STRING().getText()) {
                    case "nodeId" /* [1] */ : temp_node_id = parseLong(kvPair.value()); break;
    case "publicKey" /* [2] */ : temp_public_key = unescape(checkSize("public_key", kvPair.value().STRING()==null?null:kvPair.value().STRING().getText(), maxSize)); break;
    case "timestamp" /* [3] */ : temp_timestamp = TimestampRange.JSON.parse(kvPair.value().getChild(JSONParser.ObjContext.class, 0), false, maxDepth - 1, maxSize); break;
    
                    default: {
                        if (strictMode) {
                            // Since we are parsing is strict mode, this is an exceptional condition.
                            throw new UnknownFieldException(kvPair.STRING().getText());
                        }
                    }
                }
            }
```

kvPair.STRING().getText() needs to be converted to a JSONFieldName to make sure it matches the converted name stored in the protobuf object.

anthony-swirldslabs (approved):
Looks good. Thanks!

anthony-swirldslabs:
@berryware : PBJ 0.15.5 with this fix is being released now.

PR #836, opened 2026-05-20
perf: Remove interface

ldintr:
**Description**:

The first commit is easier to go through, touches a lot of signature. The rest can be squashed

Bench:

Current:

Benchmark                                         (type)   Mode  Cnt    Score   Error   Units
GenericParserQuickBench.bench  NotCacheableAccountIDType  thrpt   15  458.108 ± 2.490  ops/us
https://github.com/hashgraph/pbj/actions/runs/26190721867/job/77058132411

main:

Benchmark                                         (type)   Mode  Cnt    Score   Error   Units
GenericParserQuickBench.bench  NotCacheableAccountIDType  thrpt   15  458.108 ± 2.490  ops/us
https://github.com/hashgraph/pbj/actions/runs/26180014678/job/77020208498

**Checklist**

- [x ] Documented (Code comments, README, etc.)
- [x ] Tested (unit, integration, etc.)

### DIFF
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
@@ -59,7 +59,7 @@ class CodecParseMethodGenerator {
         // spotless:off
         return """
                 /**
-                 * Parses a $modelClassName object from ProtoBuf bytes in a {@link ReadableSequentialData}. Throws if in strict mode ONLY.
+                 * Parses a $modelClassName object from ProtoBuf bytes in a {@link SlimBuffer}. Throws if in strict mode ONLY.
                  * <p>
                  * The {@code maxSize} specifies a custom value for the default `Codec.DEFAULT_MAX_SIZE` limit. IMPORTANT:
                  * specifying a value larger than the default one can put the application at risk because a maliciously-crafted
@@ -83,7 +83,7 @@ class CodecParseMethodGenerator {
                  * @throws ParseException If parsing fails
                  */
                 public @NonNull $modelClassName parse(
-                        @NonNull final ReadableSequentialData input,
+                        @NonNull final SlimBuffer input,
                         final boolean strictMode,
                         final boolean parseUnknownFields,
                         final int maxDepth,
@@ -111,7 +111,7 @@ class CodecParseMethodGenerator {
                     }
                 }
                 
-                private List<UnknownField> defaultCase(int tag, int field, FieldDefinition f, boolean strictMode, boolean parseUnknownFields, List<UnknownField> $unknownFields, ReadableSequentialData input, int maxSize) throws ParseException, IOException {
+                private List<UnknownField> defaultCase(int tag, int field, FieldDefinition f, boolean strictMode, boolean parseUnknownFields, List<UnknownField> $unknownFields, SlimBuffer input, int maxSize) throws ParseException, IOException {
                     $defaultCaseBody
                     return $unknownFields;
                 }
@@ -300,7 +300,7 @@ class CodecParseMethodGenerator {
                 .formatted(tag, wireType, field.type(), fieldNum, field.name()));
         sbCase.append("%s = case%d(input, maxSize, %s);%n".formatted(tempFieldName, tag, tempFieldName));
         sbFunc.append("""
-%s case%d(ReadableSequentialData input, int maxSize, %s %s) throws ParseException, IOException {""".formatted(fieldType, tag, fieldType, tempFieldName));
+            %s case%d(SlimBuffer input, int maxSize, %s %s) throws ParseException, IOException {""".formatted(fieldType, tag, fieldType, tempFieldName));
         final String preRead;
         if (field.type() == Field.FieldType.ENUM) {
             // spotless:off
--- a/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
+++ b/pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
@@ -181,8 +181,8 @@ class CodecParseMethodGenerator {
         list.add("""
                         // -- PARSE LOOP ---------------------------------------------
                         // Continue to parse bytes out of the input stream until we get to the end.
-                        while (input.hasRemaining()) {
-                            // Note: ReadableStreamingData.hasRemaining() won't flip to false
+                        while (input.hasMore()) {
+                            // Note: ReadableStreamingData.hasMore() won't flip to false
                             // until the end of stream is actually hit with a read operation.
                             // So we catch this exception here and **only** here, because an EOFException
                             // anywhere else suggests that we're processing malformed data and so
@@ -321,17 +321,15 @@ class CodecParseMethodGenerator {
         }
         sbFunc.append("""
                 // Read the length of packed repeated field data
-                final long length = input.readVarInt(false);
+                final int length = input.readVarInt(false);
                 if (length > $maxSize) {
                     throw new ParseException("$fieldName size " + length + " is greater than max " + $maxSize);
                 }
-                if (input.remaining() < length) {
-                    throw new BufferUnderflowException();
-                }
+                input.ensure(length);
                 final var beforeLimit = input.limit();
                 final long beforePosition = input.position();
                 input.limit(input.position() + length);
-                while (input.hasRemaining()) {
+                while (input.hasMore()) {
                     $preRead$tempFieldName = addToList($tempFieldName,$readMethod);
                 }
                 input.limit(beforeLimit);
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
@@ -268,7 +268,11 @@ public interface Codec<T> {
      * @return The length of the data item in the input
      * @throws ParseException If parsing fails
      */
-    int measure(@NonNull SlimBuffer input) throws ParseException;
+    default int measure(@NonNull SlimBuffer input) throws ParseException {
+        final long startPosition = input.position();
+        parse(input);
+        return (int) (input.position() - startPosition);
+    }
 
     default int measure(@NonNull ReadableSequentialData input) throws ParseException {
         return measure(new SlimBuffer(input));
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
@@ -2,6 +2,7 @@
 package com.hedera.pbj.runtime;
 
 import com.hedera.pbj.runtime.io.ReadableSequentialData;
+import com.hedera.pbj.runtime.io.SlimBuffer;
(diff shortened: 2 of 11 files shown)


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -181,8 +181,8 @@ static ParseAndDefaultBody generateParseLoop(
         list.add("""
                         // -- PARSE LOOP ---------------------------------------------
                         // Continue to parse bytes out of the input stream until we get to the end.
-                        while (input.hasRemaining()) {
-                            // Note: ReadableStreamingData.hasRemaining() won't flip to false
+                        while (input.hasMore()) {
+                            // Note: ReadableStreamingData.hasMore() won't flip to false
```
anthony-swirldslabs, line 185: Minor: the type name in the comment text should probably be updated.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+
+public class SlimBuffer {
+    byte[] buf = new byte[16 << 10];
+    int pos, end;
+    int relLimit;
+    long absoluteLimit, offset;
+    boolean seenEOF;
+    private ReadableSequentialData input;
```
anthony-swirldslabs, line 17: All the fields should be `private`.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+
+    public SlimBuffer(ReadableSequentialData input) {
+        this.input = input;
+    }
+
+    public SlimBuffer(byte[] buf) {
+        this.buf = buf;
```
anthony-swirldslabs, line 24: You want to remove the default initializer from line 12 and move it into a constructor that needs that array initialized. Otherwise, when using this specific constructor at line 23, a new object would allocate a 16K array and then immediately throw it away at this line 24, which is bad for GC and performance in general.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+package com.hedera.pbj.runtime.io;
+
+import com.hedera.pbj.runtime.io.buffer.Bytes;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+
+public class SlimBuffer {
```
anthony-swirldslabs, line 11: Please add some javadoc for every public class/method in here. You can use the short syntax with `///`, or the long one with `/** ... */`.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+    public SlimBuffer(byte[] buf) {
+        this.buf = buf;
+        end = buf.length;
+        relLimit = buf.length;
+        absoluteLimit = buf.length;
+        seenEOF = true;
+    }
+
+    private void grow(int minCap) {
+        int newCap = Math.max(buf.length * 2, 1 << 63 - Long.numberOfLeadingZeros((minCap << 1) - 1));
```
anthony-swirldslabs, line 32: Please add parentheses around the `63 - Long...` expression. I agree that Java/javac don't need them. But humans who read this code do.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+    private void bufferMore(int relAmount) {
+        if (relAmount <= 0) {
+            relAmount = 1;
+        }
+        offset += pos;
+        relLimit = (int) (absoluteLimit - offset);
+        if (pos < end && pos != 0) {
+            int moveLen = end - pos;
+            System.arraycopy(buf, pos, buf, 0, end - pos);
+            end = moveLen;
```
anthony-swirldslabs, line 51: Is this some sort of a circular buffer or something? If so, please add a code comment and/or a javadoc for the `buf` to indicate so. If not, please add a short code comment to describe what's going on here.

In general, this particular `bufferMore` method could benefit from a javadoc and/or extra code comments.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+            end += rdlen;
+            absoluteLimit = offset + end;
+            relLimit = (int) (absoluteLimit - offset);
+            if (rdlen == 0) {
+                seenEOF = true;
+                return;
+            }
+            if (rdlen <= 0) {
+                int q = 1;
+            }
```
anthony-swirldslabs, line 72: Please remove this.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+        } catch (UncheckedIOException e) {
+            throw e;
+        }
+    }
+
+    public boolean hasMore() {
+        if (!seenEOF && pos == end) {
+            bufferMore();
```
anthony-swirldslabs, line 80: A javadoc for the `input` constructor argument could mention that the SlimBuffer relies on the caller to set a proper `limit(...)` so that this buffer doesn't buffer more than it should.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+        }
+        while (true) {
+            if (pos < relLimit) return true;
+            if (seenEOF || relLimit < end) return false;
```
anthony-swirldslabs, line 84: The `SlimBuffer(byte[])` ctor always sets seenEOF to true, so the `hasMore()` will always return `false`. Is this by design?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+            bufferMore();
```
anthony-swirldslabs, line 85: Why does this need to happen in a `while(true)` loop? Or to rephrase, why do we want to be able to call the `bufferMore()` more than once? Can we rewrite this code to make it clear that this method calls `bufferMore()` **at most** once (aka, can we nuke the while(true))?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+
+    public void ensure(int amount) throws BufferUnderflowException {
+        if (amount < 0) {
+            throw new IllegalArgumentException();
+        }
+        if (pos + amount <= relLimit) return;
+        if (seenEOF) throw new BufferUnderflowException();
```
anthony-swirldslabs, line 94: Again, this will always throw for an object constructed via SlimBuffer(byte[]). How is this constructor useful?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+        throw new BufferUnderflowException();
+    }
+
+    public long limit() {
+        if (!seenEOF) {
+            for (int i = 0; !seenEOF && i < 32; i++) {
+                bufferMore();
+            }
+            if (seenEOF == false) {
+                throw new RuntimeException(); // logic error
```
anthony-swirldslabs, line 106: I'm not sure I understand why this is an error.

I don't understand why the bufferMore() called 32 times is assumed to guarantee to read the entire input stream?

Also, I'm unsure why the limit() getter method is supposed to perform I/O.

Also, I'm unsure why it's okay to read the entire input in memory when it may be very huge. For example, some block objects may occupy over 32MB. We don't want to read them all into memory when trying to parse them.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/SlimBuffer.java
```diff
@@ -0,0 +1,235 @@
...
+        long value = 0;
+        for (int i = 0; i < 10; i++) {
+            final byte b = readByte();
+            value |= (long) (b & 0x7F) << (i * 7);
+            if (b >= 0) {
+                return zigZag ? (value >>> 1) ^ -(value & 1) : value;
+            }
+        }
+        throw new DataEncodingException("Malformed var int");
+    }
```
anthony-swirldslabs, line 140: This is an old algorithm. We replaced it recently at https://github.com/hashgraph/pbj/pull/826

You want to implement the new algorithm here.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -65,13 +66,18 @@ public interface Codec<T> {
...
-    @NonNull
-    T parse(
+    default @NonNull T parse(
             @NonNull ReadableSequentialData input,
             boolean strictMode,
             boolean parseUnknownFields,
             int maxDepth,
             int maxSize)
+            throws ParseException {
+        return parse(new SlimBuffer(input), strictMode, parseUnknownFields, maxDepth, maxSize);
```
anthony-swirldslabs, line 76: This will create a new SlimBuffer object for every parse() call in all the existing code in Consensus and Block node. I'm afraid that this may degrade the performance. And I see that you've modified the GenericParserQuickBench to use SlimBuffers natively, which hides this potential performance degradation.

I suggest building a benchmark (or reverting the changes in the aforementioned one) that calls the `parse(ReadableSequentialData...` at all times, and then run it on the `main` branch and on the branch of this PR. The numbers would give us an idea of how much the performance of the CN and BN would degrade.

If the degradation looks reasonable in the test (say, at most 5% - but I'll let @OlegMazurov and @alex-kuzmin-hg provide their opinion), then we could proceed with releasing this change and then benchmarking the CN and BN code using our performance team tools before actually upgrading PBJ in those systems.

An alternative approach, and/or if the performance degradation is significant,  would be to modify all the code in CN and BN to switch to using the SlimBuffer before upgrading PBJ.

Yet another alternative is to keep existing parse(ReadableSequentialData) implementations as they are and introduce the new parse(SlimBuffer) as an extra API. This latter approach would simplify the migration of the CN and BN code as it could be executed step by step migrating one piece of code at a time.

In general, I'm extremely concerned about this particular change here at this line (and all lines below that create new instances of SlimBuffer silently), and I'd suggest to ask our performance engineers to take a look at this.

ldintr:
Cleaned up code can be found ldintr-wip https://github.com/hashgraph/pbj/tree/ldintr-wip
I'll post updates next week as I remove and optimize more

PR #837, opened 2026-05-20
feat: RandomAccessData.put methods

### LINKED ISSUE #790: BufferedData to provide random write access API
`ByteBuffer` provide two sets of methods to write data. First, write to the current position, e.g. `put(byte)`. Second, write to a specified position, e.g. `put(offset, byte)`. This isn't limited to bytes, there are similar methods for byte arrays, byte buffers, etc.

`BufferedData` is a PBJ wrapper on top of `ByteBuffer`. It provides methods to read data at the current position (e.g. `readByte()`) and at a given offset (e.g. `getByte(offset)`), but only methods to write data to the current position (e.g. `writeByte(byte)`). There are no methods to write data to a specified offset. I'd expect them to be named `putByte(offset, byte)` to be consistent with `getByte()`.

As a result, there are no methods in high level write APIs like `ProtoWriterTools`. This class has methods to write data to the current position of a given buffer (typically, `WritableSequentialData`), but no methods to write to a given offset. Such missing methods wouldn't work with `WritableSequentialData`, since, well, it's sequential, but they can be implemented for `BufferedData` or similar. It may be reasonable to introduce a new interface like `WritableRandomAccessData` for this purpose, and let `BufferedData` implement it, but I'm not 100% sure about this idea.

anthony-swirldslabs:
**Description**:
Introducing new `RandomAccessData.put*()` methods for absolute writes into RandomAccessData objects. By default the methods throw an exception, and this works perfectly for the immutable `Bytes` which implements this interface. Other implementations, such as `BufferedData` and `MemoryData` have overrides that implement actual put operations.

The methods are: `putByte(long, byte)`, `putBytes(long, byte[])`, `putBytes(long, byte[], int, int)`, `putLong(long, long)`, and `putVarLong(long, long)` which is the minimum set of methods that could be immediately useful in the MerkleDB implementation in the Consensus Node. If necessary, we can add methods for other types (float, String, and whatnot) in the future.

Unit tests are updated to verify all the new methods.

**Related issue(s)**:

Fixes #790 

**Notes for reviewer**:
All tests should pass.

**Checklist**

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
@@ -355,10 +355,7 @@ public interface RandomAccessData {
     /// Puts a given long `value` as a big-endian 8 bytes at a given `offset`.
     /// @return the number of bytes written
     default int putLong(final long offset, final long value) {
-        checkOffset(offset, length());
-        if ((length() - offset) < Long.BYTES) {
-            throw new BufferOverflowException();
-        }
+        checkOffsetToWrite(offset, length(), Long.BYTES);
         putByte(offset, (byte) ((value >>> 56) & 0xFF));
         putByte(offset + 1, (byte) ((value >>> 48) & 0xFF));
         putByte(offset + 2, (byte) ((value >>> 40) & 0xFF));
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
@@ -17,6 +17,11 @@ import java.security.MessageDigest;
  * this interface is only backed by a buffer of some kind: an array, a {@link ByteBuffer}, a memory-mapped file, etc.
  * Unlike {@link BufferedSequentialData}, it does not define any kind of "position" cursor, just a "length" representing
  * the valid range of indexes and methods for reading data at any of those indexes.
+ * <p>
+ * In addition to "get" methods for reading data, this interface also defines a few "put" methods for writing data.
+ * By default, the "put" methods throw the `UnsupportedOperationException` and must be overridden in concrete classes
+ * to implement the write operations. If a backing datastore such as a `MemorySegment` in the `MemoryData` class
+ * is readonly, then calling a "put" method would still throw an exception because the store is immutable.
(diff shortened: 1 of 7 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
```diff
@@ -505,6 +562,21 @@ default long getVarLong(long offset, final boolean zigZag) {
         throw new DataEncodingException("Malformed var int");
     }
 
+    /// Puts a given long `value` as a no-zigZag varint at a given `offset`.
+    /// @return the number of bytes written
+    default int putVarLong(final long offset, long value) {
+        long pos = offset;
```
imalygin, line 565: Other `put` methods have `checkOffsetWrite`. Shouldn't you have it here as well?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
```diff
@@ -313,6 +352,24 @@ default long getLong(final long offset) {
                 + (b8 & 255));
     }
 
+    /// Puts a given long `value` as a big-endian 8 bytes at a given `offset`.
+    /// @return the number of bytes written
+    default int putLong(final long offset, final long value) {
+        checkOffset(offset, length());
```
imalygin: shouldn't it be `checkOffsetWrite`?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
anthony-swirldslabs: No, we shouldn't. The default implementations of put*() methods other than the core putByte() delegate to the latter, and it has proper bounds checks implemented. We don't want to run the checks twice as this is bad for performance.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/RandomAccessData.java
anthony-swirldslabs: Please see above.

anthony-swirldslabs: Oh, my bad! Good catch, actually! I fixed this one. Thanks!

imalygin (approved):

PR #864, opened 2026-06-22
chore: Bump helidon version to 4.5.0

### LINKED ISSUE #863: pbj-grpc-client-helidon binary incompatible with Helidon 4.5.0 — update Http2ClientStream constructor call
Problem

  pbj-grpc-client-helidon fails at runtime when the consuming project uses Helidon 4.5.0. The failure manifests as a NoSuchMethodError the first time a
  gRPC call is made, silently breaking all client connectivity.

  Root cause

  PbjGrpcClientStream extends Helidon's internal Http2ClientStream and calls its protected constructor:

  // Compiled against Helidon 4.4.1
  super(connection, serverSettings, ctx, http2StreamConfig, http2ClientConfig, streamIdSeq); // 6 params

  Helidon 4.5.0 removed that 6-parameter constructor and replaced it with two new ones:

  // Helidon 4.5.0
  protected Http2ClientStream(..., Http2ClientImpl http2Client)                               // 7 params
  protected Http2ClientStream(..., Http2FrameListener sendListener, Http2FrameListener recv)  // 8 params

  Because this is a protected constructor on a non-public class (Http2ClientStream is marked "not for application use"), the mismatch is not caught at
  compile time — the library compiles cleanly against 4.4.1 artifacts. The NoSuchMethodError only surfaces at runtime when a gRPC call triggers
  PbjGrpcClientStream instantiation.

  Impact

  Any project that:
  1. depends on pbj-grpc-client-helidon, and
  2. upgrades Helidon to 4.5.0

  will have all gRPC client calls fail with NoSuchMethodError. The error is not caught by catch (Exception e) guards (since Error is not an Exception),
  so callers see unexpected propagation rather than a clean failure.

  ---
  Solution

  Use the new 7-parameter constructor, passing the Http2ClientImpl that PbjGrpcClient already holds. Http2ClientImpl exposes sendListener() and
  recvListener() which the 7-param constructor uses internally to delegate to the 8-param form — this is the correct Helidon-idiomatic path.

  Changes:

  PbjGrpcClientStream.java — add Http2ClientImpl parameter, update super():
  // Before
  super(connection, serverSettings, ctx, http2StreamConfig, http2ClientConfig, streamIdSeq);

  // After
  super(connection, serverSettings, ctx, http2StreamConfig, http2ClientConfig, streamIdSeq, http2Client);

  PbjGrpcClient.createPbjGrpcClientStream() — pass the client impl:
  return new PbjGrpcClientStream(
          connection, Http2Settings.create(), clientConnection.helidonSocket(),
          new Http2StreamConfig() { ... },
          null, connection.streamIdSequence(),
          (Http2ClientImpl) http2Client);  // added

  hiero-dependency-versions/build.gradle.kts — bump Helidon:
  val helidon = "4.5.0"

  One test assertion also needed updating: Http2ClientStream.close() in 4.5.0 calls the package-private releaseReservedStream() rather than
  removeStream(int), so the verify(connection, times(1)).removeStream(0) check in PbjGrpcClientTest was removed (the method is not accessible outside
  io.helidon.webclient.http2, so it cannot be verified via Mockito from our test package).

derektriley:
**Description**:
This pull request updates the Helidon dependency and makes necessary adjustments to the gRPC client stream implementation to support the new version. The changes ensure compatibility with Helidon 4.5.0 by updating constructor signatures and usages to include the new required `Http2ClientImpl` parameter.

**Dependency upgrade:**

* Updated the Helidon dependency version from `4.4.1` to `4.5.0` in `build.gradle.kts`, preparing the project for the latest features and fixes.

**Code changes for Helidon 4.5.0 compatibility:**

* Modified the `PbjGrpcClientStream` class constructor to accept an additional `Http2ClientImpl` parameter and updated its call to the superclass constructor accordingly.
* Updated the instantiation of `PbjGrpcClientStream` in `PbjGrpcClient.java` to pass the new `Http2ClientImpl` argument as required by the updated Helidon API.
* Added the necessary import for `Http2ClientImpl` in `PbjGrpcClientStream.java` to support the new constructor parameter.

Fixes:
* BidiStreamingBuilderImpl.onComplete() calls incoming.onComplete() followed by super.onComplete(). The first call cascades
  through the user's handler → response converter → SendToClientSubscriber.onComplete(), which sends a HEADERS+END_STREAM
  trailer frame to the client. The super.onComplete() call then invokes SendToClientSubscriber.onComplete() a second time,
  sending a duplicate trailer frame.

  On the client side (Helidon 4.5.0), pushTrailers() detects the duplicate via inboundEndQueued and throws Http2Exception.
  This propagates up to the connection read loop's exception handler, which calls buffer.fail() on the stream's
  StreamBuffer. Because StreamBuffer.fail() causes all subsequent and currently-waiting poll() calls to throw immediately —
  even if unread DATA frames are already queued — the client's reply-reading loop dies after consuming only 1 of 3
  expected replies.

  The fix makes SendToClientSubscriber.onComplete() idempotent with an AtomicBoolean guard. The framework's safety-net
  behavior (calling replies.onComplete() via super.onComplete() in case the user's handler doesn't) is preserved for
  handlers that don't call it themselves; duplicate calls from handlers that do are now no-ops.

**Test code cleanup:**

* Removed a redundant verification in the `testCreatePbjGrpcClientStream` test, as the stream ID behavior has changed or is no longer relevant.

**Related issue(s)**:

Fixes #863 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
+++ b/pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
@@ -57,6 +57,7 @@ import java.util.concurrent.Flow;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -703,6 +704,8 @@ final class PbjProtocolHandler implements Http2SubProtocolSelector.SubProtocolHa
      * receives bytes from the handlers to send to the client.
      */
     private final class SendToClientSubscriber implements Pipeline<Bytes> {
+        private final AtomicBoolean completedOnce = new AtomicBoolean(false);
+
         @Override
         public void onSubscribe(@NonNull final Flow.Subscription subscription) {
             // FUTURE: Add support for flow control
@@ -754,6 +757,9 @@ final class PbjProtocolHandler implements Http2SubProtocolSelector.SubProtocolHa
 
         @Override
         public void onComplete() {
+            if (!completedOnce.compareAndSet(false, true)) {
+                return;
+            }
             new TrailerBuilder().send();
 
             deadlineFuture.cancel(false);
--- a/pbj-core/hiero-dependency-versions/build.gradle.kts
+++ b/pbj-core/hiero-dependency-versions/build.gradle.kts
@@ -3,7 +3,7 @@ group = "com.hedera.hashgraph"
 
 val antlr = "4.13.2"
 val grpc = "1.82.0"
-val helidon = "4.4.1"
+val helidon = "4.5.0"
 val protobuf = "4.35.0"
 
 val junit5 = "6.1.0"
--- a/pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClient.java
+++ b/pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClient.java
@@ -157,7 +157,8 @@ public final class PbjGrpcClient implements GrpcClient, AutoCloseable {
                     }
                 },
                 null,
-                connection.streamIdSequence());
+                connection.streamIdSequence(),
+                (Http2ClientImpl) http2Client);
     }
 
     /**
--- a/pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClient.java
+++ b/pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClient.java
@@ -156,7 +156,7 @@ public final class PbjGrpcClient implements GrpcClient, AutoCloseable {
                         return getConfig().readTimeout();
                     }
                 },
-                null,
+                ((Http2ClientImpl) http2Client).prototype(),
                 connection.streamIdSequence(),
                 (Http2ClientImpl) http2Client);
     }
--- a/pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClientStream.java
+++ b/pbj-core/pbj-grpc-client-helidon/src/main/java/com/hedera/pbj/grpc/client/helidon/PbjGrpcClientStream.java
@@ -5,13 +5,14 @@ import io.helidon.common.socket.SocketContext;
 import io.helidon.http.http2.Http2Settings;
 import io.helidon.webclient.http2.Http2ClientConfig;
 import io.helidon.webclient.http2.Http2ClientConnection;
+import io.helidon.webclient.http2.Http2ClientImpl;
 import io.helidon.webclient.http2.Http2ClientStream;
 import io.helidon.webclient.http2.Http2StreamConfig;
 import io.helidon.webclient.http2.LockingStreamIdSequence;
 
 /**
  * A package-private class that extends a Helidon client stream class only for the purpose
- * of accessing its only protected constructor. While the Http2ClientStream is marked as "not for applications use"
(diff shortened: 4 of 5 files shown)


anthony-swirldslabs (dismissed):

alex-kuzmin-hg (dismissed):
Looks OK. I'll run full SDPT after new BN tag

AlfredoG87 (approved):
LGTM

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
```diff
@@ -754,6 +757,9 @@ public void onError(@NonNull final Throwable throwable) {
 
         @Override
         public void onComplete() {
+            if (!completedOnce.compareAndSet(false, true)) {
+                return;
+            }
```
anthony-swirldslabs, line 762: Minor: I understand that this works-around the issue and seems to be correct as far as the HTTP2/GRPC protocols are concerned. However, this also means that either user service implementations and/or PBJ GRPC server implementation calls onComplete() multiple times.

If it's the former, I think the fix is great. If it's the latter, then I'd rather suggest fixing the root cause on PBJ side.

It would be nice to see some stack traces to understand who and when (and why) calls onComplete() multiple times to ensure that this fix is good.

anthony-swirldslabs (approved):

pbj-core/pbj-grpc-helidon/src/main/java/com/hedera/pbj/grpc/helidon/PbjProtocolHandler.java
derektriley: Thanks, this appears to be a PBJ framework completion-path issue (not just user misuse).

**What is happening**
In streaming end-of-stream flow, PBJ currently executes two completion paths in Pipelines:

incoming.onComplete()
super.onComplete()
For bidi/client-streaming handlers that already complete replies during incoming.onComplete(), the second call becomes a duplicate completion attempt.

**Supporting call stacks**
When the handler completes replies, completion reaches transport via this path:

com.hedera.pbj.grpc.helidon.PbjProtocolHandler$SendToClientSubscriber.onComplete
com.hedera.pbj.runtime.grpc.Pipelines$MapSubscriber.onComplete
<com.hedera...handler>.onComplete
com.hedera.pbj.runtime.grpc.Pipelines$BidiStreamingBuilderImpl.onComplete
com.hedera.pbj.runtime.grpc.Pipelines$BidiStreamingBuilderImpl.clientEndStreamReceived
com.hedera.pbj.grpc.helidon.PbjProtocolHandler.data
PBJ then also invokes fallback completion:


com.hedera.pbj.grpc.helidon.PbjProtocolHandler$SendToClientSubscriber.onComplete
com.hedera.pbj.runtime.grpc.Pipelines$PipelineBuilderImpl.onComplete
com.hedera.pbj.runtime.grpc.Pipelines$BidiStreamingBuilderImpl.onComplete
com.hedera.pbj.runtime.grpc.Pipelines$BidiStreamingBuilderImpl.clientEndStreamReceived
com.hedera.pbj.grpc.helidon.PbjProtocolHandler.data
So the framework is effectively creating a duplicate terminal signal when handlers already complete normally.

Maybe this is the best fix?

Keep fallback behavior for compatibility, but make it conditional:
call incoming.onComplete()
call super.onComplete() only if replies were not already completed by incoming path

Keep SendToClientSubscriber.onComplete() idempotent guard as transport hardening.

What do you think?

anthony-swirldslabs: Thanks for looking into it. The `incoming.onComplete()` is implemented in a custom service implementation, and it may in fact call `replies.onComplete()` (and it in fact does this in multiple tests) to indicate that no more replies are coming.

`super.onComplete()` does this almost unconditionally, too. Not sure if this is the correct behavior, but it seems to make sense - e.g. it helps service implementations that "forget" to call onComplete() in their custom implementations.

So the issue seems to be on the boundary between PBJ and service implementation, and hence isn't in full control of PBJ.

From this perspective, I think that your current fix makes total sense as it enforces the HTTP2/GRPC protocol regardless of the custom service implementation.

anthony-swirldslabs:
@derektriley @AlfredoG87 @alex-kuzmin-hg : I'm releasing PBJ 0.15.10 with this fix. It should be in Maven in about 20 minutes.

PR #883, opened 2026-08-07
perf: PBJReader/PbjWriter optimization branch

ldintr:
**Description**:

My branch for PBJ, it switches to PbjReader and PbjWriter instead of using an interface. `parse` and `write` are significantly faster. Here are some notes

- parse and write have overloads that use a thread local storage object, prefer that to creating new `PbjReader`/`PbjWriter`
- `PbjWriter` has a `toPbjReader` method which reuses the buffer. It can be used as long as you don't need to write to the writer object before the reader is done. Otherwise use toByteArray to clone the buffer
- Bytes now has `array()` and `arrayOffset()` so `PbjReader` doesn't need to copy data
- RandomAccessSequenceAdapter has been changed for the same reason (to skip a copy)
- Many exceptions were deleted. `PbjReader`/`PbjWriter` doesn't throw checked exceptions. If you need them you can use throwOnError(), or use `error()` to get an error code 
- Some exceptions look at the error message, prefer using error codes in the future
- Many functions that have `ReadableSequentialData` have either the signature change to `PbjReader` or had the function copied with the parameter changed to `PbjReader`. Same with `WritableSequentialData` and `PbjWriter`
- pbj.ReaderWriter.useStackTrace defaults to true, use false to use premade exceptions/stacktraces which will run faster

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

parse and write were changed to realParse and realWrite and changed back in the cleanup. That was to help me develop easier. hasMore changed back to hasRemaining

**Checklist**


pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -319,29 +314,37 @@ private static void generateFieldCaseStatementPacked(
...
                 if (input.position() != beforePosition + length) {
-                    throw new BufferUnderflowException();
+                    input.setError(PbjReader.BufferUnderflow);
                 }""".replace("$tempFieldName", tempFieldName)
                 .replace("$preRead", preRead)
+                .replace("$fieldType", fieldType)
                 .replace("$readMethod", field.type() == Field.FieldType.ENUM ? "value" : readMethod(field))
                 .replace("$maxSize", field.maxSize() >= 0 ? String.valueOf(field.maxSize()) : "maxSize")
                 .replace("$fieldName", field.name())
+                .replace("$divideString", divideAmount == 1 ? "" : "/%s".formatted(divideAmount))
```
anthony-swirldslabs: `%s` is designed to handle `Formattable` objects, or it falls back to calling `toString()`. This means that it accepts an object as an argument. Which in turn means that the primitive integer variable `divideAmount` gets boxed into an `Integer` at runtime. This may be inefficient. Is there a reason not to use `%d`?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -759,6 +769,26 @@ public byte[] toByteArray(final int offset, final int len) {
...
     }
 
+    /**
+     * Returns the raw backing byte array. The logical content starts at {@link #arrayOffset()} and spans
+     * {@link #length()} bytes. Mutating the returned array breaks the immutability contract of this class.
+     *
+     * @return the internal backing byte array
+     */
+    @NonNull
+    public byte[] array() {
```
anthony-swirldslabs: It's nice to mention that the API is unsafe in the javadoc, but the method name still looks too very inviting. An engineer may not think it through when using an auto-complete in their favorite IDE and choose this unsafe version vs. the safe toByteArray. I suggest to rename this method to `arrayUnsafe()` instead.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ByteArraySequentialData.java
```diff
@@ -0,0 +1,18 @@
...
+package com.hedera.pbj.runtime.io;
+
+/**
+ * Implemented by {@link ReadableSequentialData} types that are backed directly by a heap byte array.
+ * Callers (e.g. {@link PbjReader}) can detect this and use the array directly instead of copying
+ * data through an intermediate buffer.
+ */
+public interface ByteArraySequentialData {
+    /** The raw backing byte array. */
+    byte[] byteArray();
```
anthony-swirldslabs: Similar to above, this method is public and it may expose the underlying array which may be considered immutable by e.g. Bytes. Both the name of this method and its javadoc need to change to make it less likely to misuse it in applications. E.g. `byteArrayUnsafe` might be a better name.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+    private long absoluteLimit = Long.MAX_VALUE, offset;
+    private ReadableSequentialData rsd;
+    private InputStream stream;
+    private Exception cause;
+    private boolean seenEOF, includeCause;
+    private byte[] ownedBuf;
+
+    private char[] charArray;
+    private final boolean useStacktrace =
+            !"false".equalsIgnoreCase(System.getProperty("pbj.ReaderWriter.useStackTrace"));
```
anthony-swirldslabs: System properties cannot change at runtime. So reading it every time a new instance of the PbjReader is created is inefficient. You want to make this field static to read it just once during a JVM run.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+
+    public static final int EOF = -1,
+            DataEncoding = 1,
```
anthony-swirldslabs, line 45: Minor: per Java Coding Conventions, static fields should use UPPER_CASE_SNAKE_NAMES. The current naming looks unnatural to a Java programmer and causes a confusion because the names can be seen as inner class names instead. It would be nice to rename all the error codes accordingly.

The comment applies to any and all static fields, in this class and elsewhere in this fix.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.BufferOverflowException;
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+
+/**
+ * A buffer-backed reader for decoding data.
```
anthony-swirldslabs: This class has a mutable state. You want to add a paragraph to the class javadoc to explain its threading contract - specifically to mention that the application is responsible for synchronizing access to the class, and ideally should be using it on a single thread.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+    public void resetWith(byte[] buffer, int position, int endPosition) {
+        err = 0;
+        cause = null;
+        includeCause = false;
+        rsd = null;
+        stream = null;
+        construct(buffer, position, endPosition);
+        offset = 0;
+        relLimit = end;
+        absoluteLimit = end;
```
anthony-swirldslabs: I see that construct() already initializes these two fields.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+            construct(ba.byteArray(), ba.byteArrayOffset(), ba.byteArrayEnd());
+            return;
+        } else if (seq != null) {
+            rsd = seq;
+            absoluteLimit = seq.limit();
+            offset = (int) seq.position();
+        } else if (inputStream != null) {
+            stream = inputStream;
+        }
+        ownedBuf = buf = new byte[16 << 10]; // 16k is friendly to x86-64 L1 cache
```
anthony-swirldslabs, line 119: I'm curious to see TPS numbers in a test from Oleg's suite (CryptoTransfer? Or whatever Oleg would suggest) w/ and w/o the fix to see how this 16K buffer affects anything.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+     * @param offset the inclusive start index
+     * @param end    the exclusive end index
+     */
+    public PbjReader(byte[] data, int offset, int end) {
+        construct(data, offset, end);
+    }
+    /** Creates a reader backed by the given {@link ByteBuffer}. */
+    public PbjReader(ByteBuffer bb) {
+        int position = bb.arrayOffset() + bb.position();
+        construct(bb.array(), position, position + bb.remaining());
```
anthony-swirldslabs, line 181: Please note that this won't work with a direct ByteBuffer that uses off-heap memory. You want to mention this in the javadoc.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+    /** Resets this reader to read from the given {@link InputStream}. */
+    public void resetWith(InputStream inputStream) {
+        resetWith(null, inputStream);
+    }
+    /** Resets this reader to read from the given byte array. */
+    public void resetWith(byte[] data) {
+        resetWith(data, 0, data.length);
+    }
+    /** Resets this reader to read from the given {@link ByteBuffer}. */
+    public void resetWith(ByteBuffer bb) {
```
anthony-swirldslabs, line 202: Same as above, you want to mention this doesn't support direct buffers.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+    /** Resets this reader to read from the given {@link Bytes}. */
+    public void resetWith(Bytes bytes) {
+        resetWith(bytes.array(), bytes.arrayOffset(), bytes.arrayOffset() + (int) bytes.length());
+    }
+
+    /**
+     * Reads all remaining input into the internal buffer, growing it as needed, until EOF is reached.
+     */
+    public void bufferToEOF() {
+        while (err == 0) {
```
anthony-swirldslabs: Minor: the class javadoc says the error is sticky. This method reads until EOF and then sets the `err` to EOF. So the javadocs imply that all methods will return 0/null after that. You may want to clarify the javadocs to avoid the confusion.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+            buf = newBuf;
+            pos = 0;
+            int remLen = buf.length - end;
+            int rdlen = readFromInput(buf, end, remLen);
+            end += rdlen;
+            relLimit = (int) Math.min(absoluteLimit - offset, end);
+            if (rdlen == 0) {
+                seenEOF = true;
+                err = EOF;
+            }
```
anthony-swirldslabs: The logic in this method is hard to understand. It would help to add comments to explain what's going on.

Also, I see that it allocates a new buffer despite the fact that the currently allocated buffer may be sufficient for reading all the bytes. This may be inefficient, but depends on how/when the bufferToEOF() is actually used.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+     */
+    public long readVarLong(boolean zigZag) {
+        long value = 0;
+        for (int i = 0; i < 10; i++) {
+            byte b = readByte();
+            value |= (long) (b & 0x7F) << (i * 7);
+            if (b >= 0) {
+                return zigZag ? (value >>> 1) ^ -(value & 1) : value;
+            }
+        }
```
anthony-swirldslabs, line 431: We've put a lot of effort in optimizing the varint reading (and writing for that matter) operation. We want to preserve the currently used algorithm(s) unless new thorough benchmarking of this operation run on an x86 Linux system shows different results. You can find latest varint fixes with references to specific benchmarks in the commit history from spring this year.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+                case IllegalArgument:
+                    throw new IllegalArgumentException("throwOnError", cause);
+                case Unsupported:
+                    throw new RuntimeException("Hit an unsupported feature", cause);
+                case MaxDepthReached:
+                    throw new ParseException("Reached maximum allowed depth");
+                case Parse:
+                default:
+                    if (includeCause) throw new ParseException(cause);
+                    var ex = new ParseException("");
```
anthony-swirldslabs: Minor: we want to avoid using `var` in PBJ code.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+        if (useStacktrace) {
+            if (errorKind == UnknownField) {
+                cause = new UnknownFieldException(message);
+            } else if (errorKind == BufferUnderflow) {
+                cause = new BufferUnderflowException();
+            } else if (errorKind == BufferOverflow) {
+                cause = new BufferOverflowException();
+            } else if (errorKind == Parse) {
+                cause = new RuntimeException(message);
+                includeCause = false;
```
anthony-swirldslabs, line 501: Why is that? The `message` could contain very useful diagnostic information, and I see that below it would be dropped when this flag is reset. This could make debugging a lot harder than it needs to be.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+    }
+
+    /**
+     * Returns the raw internal buffer array. Intended for low-level introspection;
+     * the valid data occupies an internal window and should not be accessed without
+     * knowledge of the internals.
+     *
+     * @return the internal byte buffer
+     */
+    public byte[] array() {
```
anthony-swirldslabs: Similar to comments above, the current name of the method looks too inviting, while it can expose an array that should be considered immutable, e.g. when the reader is backed by a Bytes.
You want to rename this method to help prevent misuse and update the javadoc accordingly.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
```diff
@@ -0,0 +1,960 @@
...
+     * Reads bytes into the given {@link ByteBuffer}, advancing its position by the number of
+     * bytes read. If fewer bytes are available than the buffer's remaining capacity, the buffer
+     * is partially filled, and the number of bytes copied is returned.
+     * Returns {@code -1} on an error
+     *
+     * @param dst the destination buffer; bytes are written starting at its current position
+     * @return the number of bytes actually read, or {@code -1} if a pre-existing error is set
+     */
+    public long readBytes(@NonNull ByteBuffer dst) {
+        int len = readBytesInternalCopy(dst.array(), dst.arrayOffset() + dst.position(), dst.remaining());
```
anthony-swirldslabs, line 699: Again, you want to mention in the javadoc that this method doesn't support direct buffers.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjWriter.java
```diff
@@ -0,0 +1,1053 @@
...
+
+    /**
+     * Overwrites a single byte at the given absolute position without advancing the write cursor.
+     * Used to patch in placeholder bytes reserved earlier via {@link #placehold}.
+     *
+     * @param pos   the absolute write position to patch
+     * @param value the byte value to write
+     */
+    public void writeAt(int pos, byte value) {
+        buf[pos - offset] = value;
```
anthony-swirldslabs, line 209: What if the writer may not grow, and/or the buffer has been flushed already, would this write operation hit the correct index then?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjWriter.java
```diff
@@ -0,0 +1,1053 @@
...
+     *
+     * @param position the absolute position of the 1-byte placeholder
+     */
+    public void reinsertVarInt(int position) {
+        int relPos = position - offset;
+        int len = this.pos - relPos - 1;
+        System.arraycopy(buf, relPos + 1, buf, relPos + 2, len);
+        buf[relPos] = (byte) ((len & 0x7F) | 0x80);
+        buf[relPos + 1] = (byte) (len >>> 7);
+        this.pos++;
```
anthony-swirldslabs, line 225: This is a weird and very specialized operation. It doesn't seem to reinsert a generic varint, but a specific varint with a specific value, and it expects the value to be strictly > 127 and strictly <= 16383, or otherwise that wouldn't be a valid 2-byte varint.

I'm unsure if such a specialized operation belongs to PBJ code. This logic seems to belong to an application.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjWriter.java
```diff
@@ -0,0 +1,1053 @@
...
+            System.arraycopy(src, offset, buf, pos, length);
+            pos += length;
+            return;
+        }
+        writeBytesInternal(src, offset, length);
+    }
+
+    private void writeBytesInternal(byte[] src, int srcOffset, int length) {
+        flushOrGrow(length);
+        if (output != null && length >= 2048) {
```
anthony-swirldslabs, line 406: Why length >= 2048 ?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -62,13 +76,44 @@ public interface Codec<T> {
...
+
+    default @NonNull T wrapParse(
             @NonNull ReadableSequentialData input,
             boolean strictMode,
             boolean parseUnknownFields,
             int maxDepth,
             int maxSize)
+            throws ParseException {
+        ReadCache cache = tlsReader.get();
+        if (cache.inUse) {
```
anthony-swirldslabs, line 99: It took me a while to realize that this logic exists to support recursive parsing. You may want to add a one-liner comment above the `if` to mention this. Otherwise one might just delete it thinking that ThreadLocal guarantees this won't be reused.

The comment applies to all such patterns. Alternatively, you could document the inUse flags properly above where they're defined.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -119,54 +175,151 @@ default T parse(@NonNull ReadableSequentialData input, boolean strictMode, boole
...
-     * <p>
-     * If {@code strictMode} is {@code true}, then throws an exception if fields
-     * have been defined on the encoded object that are not supported by the parser. This
-     * breaks forwards compatibility (an older parser cannot parse a newer encoded object),
-     * which is sometimes requires to avoid parsing an object that is newer than the code
-     * parsing it is prepared to handle.
-     * <p>
-     * The {@code maxDepth} specifies the maximum allowed depth of nested messages. The parsing
-     * will fail with a ParseException if the maximum depth is reached.
+     * Parses an object from the {@link PbjReader} and returns it.
```
anthony-swirldslabs: Minor: this javadoc seems inaccurate w.r.t. the argument type.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -262,26 +447,83 @@ default int write(@NonNull T item, @NonNull byte[] output, final int startOffset
...
+        return writer.toByteArrayWrapped();
+    }
+
+    default Bytes toBytesGetInternalWrapped(@NonNull T item) {
+        WriteCache cache = tlsWriter.get();
+        if (cache.inUse) {
+            int len = measureRecord(item);
+            PbjWriter writer = new PbjWriter(len, false);
+            write(item, writer);
+            return writer.internalArrayWrapped();
```
anthony-swirldslabs, line 545: Internal array in a writer is generally mutable, so wrapping it in Bytes might violate its mutability contract. You want to carefully describe the intended usage in the javadoc to avoid misuse of this method.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -994,6 +1677,13 @@ public static <T extends WritableSequentialData> void writeDelimited(
         writer.accept(out);
     }
 
+    public static void writeDelimited(
```
anthony-swirldslabs, line 1680: Minor: for arrays, we've introduced a separate `ProtoArrayWriterTools`. The reason for that was because this `ProtoWriterTools.java` was already too long a file. I suggest to move all the PbjWriter-backed methods into a new class for the same reason because it's difficult to work with source files that are thousands lines long.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Utf8Tools.java
```diff
@@ -12,6 +13,31 @@
  */
 public final class Utf8Tools {
 
+    // return length or 0 on error
+    static int encodedLength(String sz) {
```
anthony-swirldslabs, line 17: Minor: this API could be error-prone if `sz.length() == 0`. Please consider a different value to indicate an error.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Utf8Tools.java
```diff
@@ -115,6 +141,10 @@ static void encodeUtf8(final CharSequence in, final WritableSequentialData out)
         }
     }
 
+    static void WriteUTF8(String str, PbjWriter out) {
```
anthony-swirldslabs: In Java, method names should start with a lower-case letter. Apart from being a part of Java Coding Conventions, in practice this helps avoid confusing them with class names and constructor names.

pbj-core/version.txt
```diff
@@ -1 +1 @@
-0.15.0-SNAPSHOT
+0.pbj.1
```
anthony-swirldslabs: This may be a part of your current testing/CN-repo-interoperability setup. However, we don't want to merge this change in `version.txt` into `main`.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
ldintr: Good catch, I updated both places

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
ldintr: I'll probably change this to all caps. Does anyone prefer the way it is? If no one does Ill change

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
ldintr: I know its a good number, but I always tested in C(++) code. After seeing how little pbj is used in other codebases, I doubt it'll show up outside of pbj test

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
ldintr: Hmm... I rather write a test and fix this. I ran this on pbj, cn, and entente (test and outside of test) and havent had an issue AFAIK. I would prefer more to delete this outright and the resetWith variant. I guess fix this for the next patch

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
ldintr: I fixed by adding the below to the end of the PbjReader javadoc. If anyone wants something better let me know

> * EOF is not an error and will return 0 when error() is called.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
ldintr: bufferToEOF isn't used anywhere (except in my test which I can delete). It was old code that I kept in case anyone wanted it. I think I should delete this since its unused and may grow stale

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
ldintr: I agree. I was hoping I can measure a before and after to see if it still works on my version of the code. I'd like to wait for numbers before I make the change. Hopefully I'll get some soon. I have added this to my todo list

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
ldintr: ✅ I removed var from Pbj reader, writer, and I don't see any in codec, jsonCodec or Utf8Tools.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
ldintr: BufferOverflowException didn't have a message constructor or an obvious way to set a message. Claude says only solution is to wrap this in another exception, which is no good bc it'd change what code is expecting

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
ldintr: I updated the class comment to "ByteBuffer using a direct buffer is not supported" until I can make a fix (or remove the overloads)

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjWriter.java
ldintr: I was meaning to change code to avoid that. I renamed to writeAtUnsafe, maybe writeAtInternal is better? I'll see if I can move logic into PbjWriter so that function isnt exposed

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjWriter.java
ldintr: I didn't measure. I was thinking it'd be more efficient to write it directly into the output stream instead of copying then flush. I choose it because 2k is 1/8th of the size. However, maybe 4 or 8k would be better. In practice, I have no idea if any messages are 2k/4k/8k

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: I added it to inUse and I'll add it to the next PR notes. "For recursive situations if the caller ever requires it". PR note will be different

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
ldintr: I want them next to eachother for the moment so its easy to see theyre both the same implementation. I imagine we can't delete the WritableSequentialData version anytime soon? If we can I can move things elsewhere

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Utf8Tools.java
ldintr: I was worried about someone assuming positive only values and getting a negative count losing data if they're summing things up. I believe elsewhere I also return empty string on error so its consistent. I'll see if I can move this and others into pbj writer (or reader) so the error can be found in one place

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Utf8Tools.java
ldintr: Done. I missed that one :( most places had writeUTF8

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
anthony-swirldslabs: Consensus Node uses PBJ for both disk (aka the state/merkleDB) and network (e.g. for parsing API requests) I/O. So it's going to be using lots of PbjReader objects. Oleg can suggest a specific test that would maximize the visibility of the usage to measure the performance effects.

Please note that I'm not particularly concerned about the exact size of a buffer but rather about the buffer existence in the first place because this is an extra object in the Java heap which GC would have to deal with eventually. Adding more work to GC (and generally, consuming more RAM) generally slows things down.

That's the reason why I'm curious to see how a real, Consensus Node-based benchmark performs w.r.t. the TPS as well as RAM usage under load.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjReader.java
anthony-swirldslabs: Please let me clarify: I see that you create a `cause` object that preserves the `message` which is great. But you immediately set the `includeCause` flag to false. And there's a code down there that uses this flag to skip adding the `cause` Throwable to an exception being thrown. This effectively drops the `message` as far as the PBJ client code is concerned.

So my question really is about the line 428 - why do we set the flag to false?

Could you please revisit the logic around `cause` and the `includeCause` to ensure we preserve the most information possible about any failures reported from the internal code.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjWriter.java
anthony-swirldslabs: `Unsafe` is better as it makes an engineer stop and think. This is the whole point of having such a name.

But yes, ideally unsafe logic would reside in some private methods inaccessible by client code. It never hurts to have a scary name for such methods, anyway, regardless of whether they are private or public.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/PbjWriter.java
anthony-swirldslabs: I see. But even 2048 may not fit into the buffer if the buffer is currently almost full. So from this perspective you could choose the full buffer size, or, say 1/2. Also, in the future even the buffer size might change, so perhaps the threshold should be computed instead?

But regardless, the main point of my comment here is that these 2048 look like a magic number. And even if the value was computed dynamically, it would still be hard to understand why this threshold is present in the logic. You want to add a comment in code directly above this line to explain this so that the code is maintainable.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -485,6 +490,11 @@ public ReadableSequentialData toReadableSequentialData() {
         return new RandomAccessSequenceAdapter(this);
     }
 
+    @NonNull
+    public PbjReader toPbjReader() {
```
artemananiev, line 494: Is this one really needed? There's clear way to create a PbjReader from Bytes, just use PbjReader constructor, no need to expose any Bytes internals to that class or whatever

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
artemananiev: Could you provide more info when users may need the backing array, please? It would be great to avoid exposing the array in API in any way (as Bytes does)

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -759,6 +769,26 @@ public byte[] toByteArray(final int offset, final int len) {
...
+    public byte[] array() {
+        return buffer;
+    }
+
+    /**
+     * Returns the offset within {@link #array()} where the logical content of this {@link Bytes} begins.
+     *
+     * @return start offset into the backing array
+     */
+    public int arrayOffset() {
```
artemananiev: This looks even more internal (and therefore subject to removal) than array(). I've always found ByteBuffer.arrayOffset() confusing, and this one looks pretty similar to that

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
artemananiev: Oh, wait, this is `Bytes`, not `PbjReader` or `PbjWriter`. So exposing the backing array is not acceptable

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ByteArraySequentialData.java
artemananiev: I am not sure this interface is the right way to provide access to a byte array to `PbjReader`. What is `Bytes.toPbjReader()` hands over its backing array to the created `PbjReader` without exposing any methods like `array()` or `arrayOffset()`? Also, need to make sure the array cannot be accessed in `PbjReader` in any legal way

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -16,6 +18,18 @@
...
+        PbjWriter writer = new PbjWriter();
+        boolean inUse = false;
+    }
+
+    class ReadCache {
+        PbjReader reader = new PbjReader(Bytes.EMPTY);
+        boolean inUse = false;
+    }
+
+    ThreadLocal<WriteCache> tlsWriter = ThreadLocal.withInitial(WriteCache::new);
```
artemananiev, line 33: This means tons of thread locals, one per generated Codec class (there are many!)

artemananiev: Actually, two per class

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -16,6 +18,18 @@
  * @param <T> The type of object to serialize and deserialize
  */
 public interface Codec<T> {
+    class WriteCache {
+        PbjWriter writer = new PbjWriter();
```
artemananiev, line 22: This looks like an implementation optimization rather than a part of the contract. It would be better to place it to a subclass, e.g. introduce `AbstractCodec implements Codec` and change codec generation code to use `AbstractCodec`. This would not be perfect, either, but still slightly better than extra classes and instance variables in an interface, from my perspective

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -62,13 +76,44 @@ public interface Codec<T> {
...
+    default @NonNull T wrapParse(
             @NonNull ReadableSequentialData input,
             boolean strictMode,
             boolean parseUnknownFields,
             int maxDepth,
             int maxSize)
+            throws ParseException {
+        ReadCache cache = tlsReader.get();
+        if (cache.inUse) {
+            PbjReader reader = new PbjReader(input);
```
artemananiev, line 100: New PbjReader per recursive call. Is it possible to wrap a `ReadableSequentialData` input into a `PbjReader` just once and use it for all recursive calls?

In general, this parse/wrapParse approach looks somewhat confusing to me, especially at the interface level

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -119,54 +175,151 @@ default T parse(@NonNull ReadableSequentialData input, boolean strictMode, boole
...
+    }
+
+    @NonNull
+    default T parse(@NonNull InputStream in, boolean strictMode, boolean parseUnknownFields, int maxDepth)
+            throws ParseException {
+        return parse(in, strictMode, false, maxDepth, DEFAULT_MAX_SIZE);
+    }
+
+    @NonNull
+    default T parse(@NonNull InputStream in, boolean strictMode, boolean parseUnknownFields, int maxDepth, int maxSize)
```
artemananiev, line 236: Are all these new `parse()` methods actually needed?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -267,6 +330,112 @@ public static String readString(final ReadableSequentialData input, final long m
...
+            if (codepoint > 0x10FFFF) return -1; // Illegal range
+            int v = codepoint - 0x10000;
+            dst[di + 0] = (char) (0xD800 + ((v >> 10) & 0x3FF));
+            dst[di + 1] = (char) (0xDC00 + (v & 0x3FF));
+            di += 2;
+        }
+        return di;
+    }
+
+    public static int fromUTF8(char[] dst, byte[] src, int offset, int pos, int length) {
```
artemananiev, line 409: JavaDoc, please

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -307,6 +476,15 @@ public static Bytes readBytes(final ReadableSequentialData input, final long max
         return bytes;
     }
 
+    public static Bytes readBytes(PbjReader input, final long maxSize) {
```
artemananiev, line 497: JavaDoc, please.

Why isn't it consistent with the method above (doesn't throw ParseException, etc.)?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
ldintr: I did it to keep it consistent with how people use bytes with toReadableSequentialData. I think I should delete it because the parse have Bytes overload, but I dont know if deleting this will make it harder to switch to my new code.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: Thats true in C++, I asked claude and it said one per thread and mentioned type erasers. I just double checked now by writing code and it seems like test spawn more workers than threads I have on my machine. But I can see it's not making more than one per thread

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: There's code in CN that overloads just a few parameters, which is why I changed the generated code to realParse briefly, so I can have it run the expected code. I think having all these overloads would be better than rearchitecture the code

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Utf8Tools.java
anthony-swirldslabs: This specific method is package-private, so it's only usable in PBJ and by PBJ directly. There should be very few usages, I think, so it should be easy to make sure they all have the correct assumptions. Also, the method could still have a javadoc to explain what it returns in more details.

But my main point is that `encodedLength("")` would return 0, but this is actually not an error - it's real encoded length of an empty string. I think this result is a lot more confusing and error-prone than using a negative value to represent an error.

PR #892, opened 2026-08-20
Codec Interface to Abstract class, made all paths call parseImpl/writeImpl

ldintr:
**Description**:

I didn't like how classes could have a 5 signature parse call a 3 signature overload ignoring some parameters (idr where I saw that, it was outside of obj). I would rather a consistent entry (which I'll use to call throwOnError in the future), and a consistent impl.

Here's a quick look. I may change (or rebase in the future) better javadoc.

**Checklist**

**Comments**

I noticed I couldn't make this final because sometimes a class overloads it and other times there's no impl

```java
    protected final int writeImpl(@NonNull T item, @NonNull byte[] output, final int startOffset) {
        final BufferedData bufferedData = BufferedData.wrap(output, startOffset, output.length - startOffset);
        try {
            write(item, bufferedData);
        } catch (IOException e) {
            throw new UncheckedIOException(e);
        }
        return (int) bufferedData.position();
    }
```

It seems like there's a chance they cane become inconsistent

```java
    protected int writeImpl(@NonNull MessageInProtoPackage data, @NonNull byte[] output, final int startOffset) {
        int offset = startOffset;
        // [1] - bytesField
        offset += ProtoArrayWriterTools.writeBytes(output, offset, MessageInProtoPackageSchema.BYTES_FIELD,  data.bytesField(), true);

        // Write unknown fields if there are any
        for (final UnknownField uf : data.getUnknownFields()) {
            final int tag = (uf.field() << TAG_FIELD_OFFSET) | uf.wireType().ordinal();
            offset += ProtoArrayWriterTools.writeUnsignedVarInt(output, offset, tag);
            offset += uf.bytes().writeTo(output, offset);
        }
        return offset - startOffset;
    }

    protected void writeImpl(@NonNull MessageInProtoPackage data, @NonNull final WritableSequentialData out) throws IOException {
            // [1] - bytesField
        writeBytes(out, MessageInProtoPackageSchema.BYTES_FIELD, data.bytesField(), true);

        // Check if not-empty to avoid creating a lambda if there's nothing to write.
        if (!data.getUnknownFields().isEmpty()) {
            data.getUnknownFields().forEach(uf -> {
                final int tag = (uf.field() << TAG_FIELD_OFFSET) | uf.wireType().ordinal();
                out.writeVarInt(tag, false);
                uf.bytes().writeTo(out);
            });
        }
    }
```

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
@@ -3,12 +3,13 @@ package com.hedera.pbj.runtime;
 
 import com.hedera.pbj.runtime.io.ReadableSequentialData;
 import com.hedera.pbj.runtime.io.WritableSequentialData;
-import com.hedera.pbj.runtime.io.buffer.BufferedData;
+import com.hedera.pbj.runtime.io.buffer.BufferedSequentialData;
 import com.hedera.pbj.runtime.io.buffer.Bytes;
+import com.hedera.pbj.runtime.io.buffer.PbjReader;
+import com.hedera.pbj.runtime.io.buffer.PbjWriter;
 import com.hedera.pbj.runtime.io.stream.WritableStreamingData;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
-import java.io.UncheckedIOException;
 
 /**
  * Encapsulates Serialization, Deserialization and other IO operations.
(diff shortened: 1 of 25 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -94,7 +105,8 @@ T parse(
      * @throws ParseException If parsing fails
      */
     @NonNull
-    default T parse(@NonNull ReadableSequentialData input, boolean strictMode, boolean parseUnknownFields, int maxDepth)
+    public final T parse(
+            @NonNull ReadableSequentialData input, boolean strictMode, boolean parseUnknownFields, int maxDepth)
```
anthony-swirldslabs, line 206: Minor: I'd go as far as to actually nuke all the overloads that supply default argument values. In the current fix where these methods have simply been made `final`, the benefit of the code simplification is a bit small to my liking.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -221,7 +237,12 @@ default T parseStrict(@NonNull Bytes bytes) throws ParseException {
      * @throws UncheckedIOException If the there is a problem writing to the output array.
      * @throws IndexOutOfBoundsException If the output array is not large enough to hold the entire item.
      */
-    default int write(@NonNull T item, @NonNull byte[] output, final int startOffset) {
+    public final int write(@NonNull T item, @NonNull byte[] output, final int startOffset) {
+        return writeImpl(item, output, startOffset);
+    }
+
+    // some classes generate this overload, but not all
```
anthony-swirldslabs: How come? Our codec generator here in PBJ is supposed to generate them (or not) unconditionally.

Also, you want to add javadocs to any public/protected methods that you introduce here.

anthony-swirldslabs (commented):
You want to add javadocs to any new protected/public methods. I also left a minor comment. Other than that, the changes look great. Thanks!

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: I looked into it and changed the comment to

```
CodecWriteByteArrayMethodGenerator.java generates this method for Codecs but not JsonCodecs
We provide a default implementation here so hand written codecs don't need to.
```

anthony-swirldslabs: Thanks! If that comment in code says exactly this verbatim, then that part would look good.

PR #899, opened 2026-08-25
feat: UTF8 to string (fromUTF8)

ldintr:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
@@ -267,6 +267,126 @@ public final class ProtoParserTools {
         }
     }
 
+    private static int fromUTF8Tail(char[] dst, int di, byte[] src, int i, int endPos) {
+        while (i < endPos) {
+            int a = src[i];
+            if ((a & 0x80) == 0) {
+                dst[di++] = (char) a;
+                i++;
+                continue;
+            }
+            if (i + 1 >= endPos) return -1;
+
+            int b = src[i + 1];
+            if ((a & 0xE0) == 0xC0) {
+                if ((b & 0xC0) == 0x80) {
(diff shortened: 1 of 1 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -267,6 +267,126 @@ public static String readString(final ReadableSequentialData input, final long m
...
+     * @param dst    the destination char array, written starting at index {@code pos}
+     * @param src    the source byte array containing UTF-8 data
+     * @param offset the base index within {@code src} where the UTF-8 region begins
+     * @param pos    bytes already decoded by the ASCII fast path; doubles as the read offset
+     *               (relative to {@code offset}) and the initial write index in {@code dst}
+     * @param length the total byte length of the UTF-8 region in {@code src} starting at {@code offset}
+     * @return the total number of {@code char}s written to {@code dst}, or {@code -1} if the
+     *         input contains an illegal byte sequence (surrogate range or out-of-range codepoint)
+     */
+    public static int fromUTF8(char[] dst, byte[] src, int offset, int pos, int length) {
```
anthony-swirldslabs, line 346: Minor: I'm not an expert in UTF-8, so I cannot meaningfully review all the conditionals and math/bitwise logic. The code looks good overall, but I cannot tell if there's no any off-by-one errors or any incorrect logic. For this reason, it would be nice to add a small unit test for this new method as a part of this PR, to ensure it does what we expect it to do.

anthony-swirldslabs (approved):

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
ldintr: Test will be in a later PR. I also had AI look it over as well. I'll try to get that committed soon

PR #900, opened 2026-08-25
feat: Adding PbjReader

ldintr:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ByteArraySequentialData.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ByteArraySequentialData.java
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.pbj.runtime.io;
+
+/**
+ * Implemented by {@link ReadableSequentialData} types that are backed directly by a heap byte array.
+ * Callers (e.g. {@link com.hedera.pbj.runtime.io.buffer.PbjReader}) can detect this and use the array directly
+ * instead of copying data through an intermediate buffer.
+ */
+public interface ByteArraySequentialData {
+    /** The raw backing byte array. */
+    byte[] byteArrayUnsafe();
+
+    /** Absolute index within {@link #byteArrayUnsafe()} where currently readable data begins. */
+    int byteArrayUnsafeOffset();
+
+    /** Absolute exclusive index within {@link #byteArrayUnsafe()} where currently readable data ends. */
+    int byteArrayUnsafeEnd();
+}
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
@@ -485,6 +485,11 @@ public final class Bytes implements RandomAccessData, Comparable<Bytes> {
         return new RandomAccessSequenceAdapter(this);
     }
 
+    @NonNull
+    public void setPbjReader(@NonNull PbjReader reader) {
+        reader.resetWith(buffer, start, start + length);
+    }
+
     /**
      * Exposes this {@link Bytes} as an {@link InputStream}. This is a zero-copy operation.
      *
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
@@ -486,7 +486,7 @@ public final class Bytes implements RandomAccessData, Comparable<Bytes> {
     }
 
     @NonNull
-    public void setPbjReader(@NonNull PbjReader reader) {
+    public void resetPbjReader(@NonNull PbjReader reader) {
         reader.resetWith(buffer, start, start + length);
     }
 
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
@@ -1,0 +1,1032 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.pbj.runtime.io.buffer;
+
+import com.hedera.pbj.runtime.ParseException;
+import com.hedera.pbj.runtime.ProtoParserTools;
+import com.hedera.pbj.runtime.UnknownFieldException;
+import com.hedera.pbj.runtime.io.ByteArraySequentialData;
+import com.hedera.pbj.runtime.io.DataEncodingException;
+import com.hedera.pbj.runtime.io.ReadableSequentialData;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.BufferOverflowException;
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+
+/**
+ * A buffer-backed reader for decoding data. Not thread safe.
+ *
+ * <p>PbjReader maintains an internal byte buffer and reads from a {@link ReadableSequentialData},
+ * {@link InputStream}, or a byte array. When backed by a stream the internal buffer is refilled
+ * on demand as data is consumed. A stream will never be read past its limit.
+ *
+ * <p>Errors are tracked internally rather than thrown immediately. Call {@link #error()} to check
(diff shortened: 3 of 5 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/Bytes.java
```diff
@@ -485,6 +485,11 @@ public ReadableSequentialData toReadableSequentialData() {
         return new RandomAccessSequenceAdapter(this);
     }
 
+    @NonNull
+    public void setPbjReader(@NonNull PbjReader reader) {
+        reader.resetWith(buffer, start, start + length);
```
anthony-swirldslabs: `set` may not be the perfect verb for this method because it's not a setter, i.e. it doesn't update the internal state of the Bytes object. `resetPbjReader` might sound better.

Also, you want to add a javadoc for this new public method.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1032 @@
...
+            MaxDepthReached = 12,
+            MalformString = 13;
+
+    private static final UnknownFieldException premadeUnknown;
+    private static final BufferUnderflowException premadeUnderflow;
+    private static final BufferOverflowException premadeOverflow;
+    private static final RuntimeException premadeRuntime, premadeUnsupported;
+    private static final DataEncodingException premadeDataEncoding;
+    private static final IllegalArgumentException premadeIllegal;
+    private static final ParseException premadeParseEmpty, premadeParseUnknown, premadeMaxDepth;
```
anthony-swirldslabs, line 66: `static` members in Java should all be UPPER_SNAKE_CASE. Otherwise, it's too easy to confuse them with instance members or local variables.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1032 @@
...
+     * On a malformed varint (more than 10 bytes), sets {@link #DataEncoding} and returns
+     * {@code -1}; however, {@code -1} is also a valid decoded value, so callers must check
+     * {@link #error()} to distinguish an error from a legitimate result.
+     *
+     * @return the decoded long value
+     */
+    public long readVarLongNoZZ() {
+        if (pos + 10 <= relLimit) {
+            long value = 0;
+            for (int i = 0; i < 10; i++) {
```
anthony-swirldslabs, line 407: For reading varints, we want to use the algorithm that we've determined is the fastest and which is implemented elsewhere in PBJ.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1032 @@
...
+        if (useStacktrace) {
+            if (errorKind == UnknownField) {
+                cause = new UnknownFieldException(message);
+            } else if (errorKind == BufferUnderflow) {
+                cause = new BufferUnderflowException();
+            } else if (errorKind == BufferOverflow) {
+                cause = new BufferOverflowException();
+            } else if (errorKind == Parse) {
+                cause = new RuntimeException(message);
+                includeCause = false;
```
anthony-swirldslabs, line 498: I'm still concerned about setting `includeCause` to `false` in this branch. To me it seems as though we're going to lose some context-related information.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1032 @@
...
+        int utf16Len = ProtoParserTools.fromUTF8(charArray, data, bufPos, i, length);
+        if (utf16Len >= 0) {
+            return new String(charArray, 0, utf16Len);
+        }
+        setError(PbjReader.Parse);
+        return "";
+    }
+
+    @Override
+    public void close() {
```
anthony-swirldslabs, line 1021: I've already seen this code a couple of times in previous iterations, so I mostly just skimmed it through this time. However, I'd like @artemananiev to review it thoroughly this time.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ByteArraySequentialData.java
```diff
@@ -0,0 +1,18 @@
...
+package com.hedera.pbj.runtime.io;
+
+/**
+ * Implemented by {@link ReadableSequentialData} types that are backed directly by a heap byte array.
+ * Callers (e.g. {@link com.hedera.pbj.runtime.io.buffer.PbjReader}) can detect this and use the array directly
+ * instead of copying data through an intermediate buffer.
+ */
+public interface ByteArraySequentialData {
+    /** The raw backing byte array. */
+    byte[] byteArrayUnsafe();
```
anthony-swirldslabs, line 11: As discussed previously, these operations look unsafe and we'd prefer not to introduce such APIs. Can we not use an approach similar to what's done with Bytes in this PR?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: I still would like to remeasure which implementation is the fastest since so much has changed. I rather leave it to this simple implementation until we measure

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: IIRC CN depends on this behavior. Will fix in future commit when we decide on removing or keeping error codes. I added this to my note

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ByteArraySequentialData.java
ldintr: Oops, this was meant for a later commit. IIRC there was some issue with RandomAccessSequenceAdapter being private but I think I can solve it. I'll remove this

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: I'll have to do this in a later PR, or force push so the PRs after this commit doesn't break from the rename

anthony-swirldslabs: A force-push shouldn't be necessary for this. You should be able to add a commit to the current branch, and then merge changes from upstream branches to upper PRs.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
anthony-swirldslabs: Could you please file an issue assigned to yourself so that we can track this?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
anthony-swirldslabs: I suggest to modify the CN code once the CN is being integrated with the PbjReader. As of now, this current implementation seems to lose an important piece of information about an exception, and I'd prefer we don't lose it.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/ByteArraySequentialData.java
anthony-swirldslabs: Please remove it in this PR, so that we don't merge that interface accidentally.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: Sure. https://github.com/hashgraph/pbj/issues/904

PR #906, opened 2026-08-31
feat: UTF8 to string (fromUTF8)

### LINKED ISSUE #922: feat: UTF8 to string (fromUTF8)
Adds `fromUTF8`/`fromUTF8Tail` to `ProtoParserTools`, decoding raw UTF-8 bytes directly into a caller-supplied `char[]` (handling 1–4 byte sequences and surrogate pairs), returning `-1` on malformed input instead of throwing.

**Benefit:** groundwork for `PbjReader.readString` to avoid allocating an intermediate `String`/`CharBuffer` per read.

PR: hashgraph/pbj#906

ldintr:
Adds fromUTF8/fromUTF8Tail to ProtoParserTools, decoding raw UTF-8 bytes directly into a caller-supplied char[] (handling 1–4 byte sequences and surrogate pairs), returning -1 on malformed input instead of throwing. Groundwork for PbjReader.readString to avoid allocating an intermediate String/CharBuffer.

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
@@ -267,6 +267,126 @@ public final class ProtoParserTools {
         }
     }
 
+    private static int fromUTF8Tail(char[] dst, int di, byte[] src, int i, int endPos) {
+        while (i < endPos) {
+            int a = src[i];
+            if ((a & 0x80) == 0) {
+                dst[di++] = (char) a;
+                i++;
+                continue;
+            }
+            if (i + 1 >= endPos) return -1;
+
+            int b = src[i + 1];
+            if ((a & 0xE0) == 0xC0) {
+                if ((b & 0xC0) == 0x80) {
+                    dst[di++] = (char) (((a & 0x1F) << 6) | (b & 0x3F));
+                    i += 2;
+                    continue;
+                } else {
+                    return -1; // Bad encoding
+                }
+            }
+
+            if (i + 2 >= endPos) return -1;
+
+            int c = src[i + 2];
+            int codepoint = -1;
+            if ((a & 0xF0) == 0xE0) {
+                if ((b & 0xC0) == 0x80 && (c & 0xC0) == 0x80) {
+                    codepoint = ((a & 0xF) << 12) | ((b & 0x3F) << 6) | (c & 0x3F);
+                    i += 3;
+                } else {
+                    return -1; // Bad encoding
+                }
+            } else {
+                if (i + 3 >= endPos) return -1;
+                int d = src[i + 3];
+                if ((a & 0xF8) == 0xF0 && (b & 0xC0) == 0x80 && (c & 0xC0) == 0x80 && (d & 0xC0) == 0x80) {
+                    codepoint = ((a & 7) << 18) | ((b & 0x3F) << 12) | ((c & 0x3F) << 6) | (d & 0x3F);
+                    i += 4;
+                } else {
+                    return -1; // Bad encoding
+                }
+            }
+
+            if (codepoint <= 0xFFFF) {
+                if (codepoint < 0 || (codepoint >= 0xD800 && codepoint < 0xE000)) return -1; // [D800, E000) is illegal
+                dst[di++] = (char) codepoint;
+                continue;
+            }
+
+            if (codepoint > 0x10FFFF) return -1; // Illegal range
+            int v = codepoint - 0x10000;
+            dst[di + 0] = (char) (0xD800 + ((v >> 10) & 0x3FF));
+            dst[di + 1] = (char) (0xDC00 + (v & 0x3FF));
+            di += 2;
+        }
+        return di;
+    }
+
+    /**
+     * Decodes UTF-8 bytes from {@code src} into the {@code char[]} destination, supporting all
+     * four UTF-8 byte widths (1–4 bytes). Codepoints above U+FFFF are written as surrogate pairs.
+     * Prefer using the simpler {@link #readString} function.
+     *
+     * <p>Decoding starts at {@code src[offset + pos]} and continues until {@code src[offset + length]}.
+     * The main loop keeps a 4-byte lookahead (exits when fewer than 5 bytes remain), delegating
+     * the final bytes to {@link #fromUTF8Tail}. Any unrecognised byte sequence returns {@code -1}.
+     *
+     * @param dst    the destination char array, written starting at index {@code pos}
+     * @param src    the source byte array containing UTF-8 data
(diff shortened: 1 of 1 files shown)


imalygin (changes requested):
Please, add unit tests.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -267,6 +267,126 @@ public static String readString(final ReadableSequentialData input, final long m
...
 
+    private static int fromUTF8Tail(char[] dst, int di, byte[] src, int i, int endPos) {
+        while (i < endPos) {
+            int a = src[i];
+            if ((a & 0x80) == 0) {
+                dst[di++] = (char) a;
+                i++;
+                continue;
+            }
+            if (i + 1 >= endPos) return -1;
```
artemananiev, line 278: I wonder how spotless reacts to these one-line "if' statements without { }, are they allowed?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -267,6 +267,126 @@ public static String readString(final ReadableSequentialData input, final long m
...
+     * @param dst    the destination char array, written starting at index {@code pos}
+     * @param src    the source byte array containing UTF-8 data
+     * @param offset the base index within {@code src} where the UTF-8 region begins
+     * @param pos    bytes already decoded by the ASCII fast path; doubles as the read offset
+     *               (relative to {@code offset}) and the initial write index in {@code dst}
+     * @param length the total byte length of the UTF-8 region in {@code src} starting at {@code offset}
+     * @return the total number of {@code char}s written to {@code dst}, or {@code -1} if the
+     *         input contains an illegal byte sequence (surrogate range or out-of-range codepoint)
+     */
+    public static int fromUTF8(char[] dst, byte[] src, int offset, int pos, int length) {
```
artemananiev, line 346: How is this method supposed to be used? This "pos" argument looks like a caller has already partially parsed a string and needs to parse the remainder. That's expected for methods like the above "fromUTF8Tail()", which is private and only called from "fromUTF8()", but this method is different. Some examples (benchmarks, unit tests) would be nice to have

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
ldintr: I ran spotless on it before committing. I don't know if there's an exception for return statements but it didn't complain

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
ldintr: It's used by PbjReader.readString, they use to be in the same commit but I was asked to split them

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -267,6 +267,126 @@ public static String readString(final ReadableSequentialData input, final long m
...
+            int a = src[i];
+            if ((a & 0x80) == 0) {
+                dst[di++] = (char) a;
+                i++;
+                continue;
+            }
+            if (i + 1 >= endPos) return -1;
+
+            int b = src[i + 1];
+            if ((a & 0xE0) == 0xC0) {
```
imalygin, line 281: This decoder validates the surrogate range and the > U+10FFFF ceiling, but it never rejects overlong (non-shortest) forms. 

All three widths leak:

- `C0 AF` → decodes to `U`+002F` /. The check `(a & 0xE0) == 0xC0` admits lead bytes C0 and C1, which are always overlong.
- `E0 80 AF → U+002F`. Any `E0` lead with a second byte `< A0` yields a codepoint < `0x800`.
- `F0 80 80 AF → U+002F`, and worse, it's collapsed into a single BMP char via the codepoint `<= 0xFFFF` branch (4 bytes in, 1 char out). Any `F0` lead with second byte < 90 is overlong.

I assume this method is replacing `readString`. JDK decoder rejects all three cases above, and `Utf8Tools` is documented as "byte for byte identical" to protoc (whose decoder also rejects overlong forms). So this isn't a stylistic strictness choice - accepting overlong forms is a behavioral regression against the exact contract the existing path was written to uphold, and it's the classic overlong-/ filter-bypass hazard.

ldintr: I finished the fix

imalygin: Great! Please add the fix to this branch.

Jeffrey-morgan34:
is there a reason we're not catching 2/3/4-byte overlongs here? the old ProtoParserTools.readString ran a strict CharsetDecoder with onMalformedInput(REPORT)  there's a comment on it saying that's deliberate so C0 80, E0 80 80, F0 80 80 80 all threw MalformedProtobufException and now they decode to U+0000. surrogates and >10FFFF are both still caught? 

It's only the min-value checks that are missing: fromUTF8 at 358 / 365 / 370, and the same three in fromUTF8Tail at 283 / 297 / 306. protobuf-java rejects overlongs too, so this puts us out of step with anything parsing the same bytes downstream, and it means two different byte sequences decode to the same string.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
artemananiev: +1

artemananiev:
At least one issue has been found in the PR (and fixed, but the commit with the fix is missing in the PR). How were these changes tested? How to make sure there are no more issues? Unit tests would be nice to have

Jeffrey-morgan34:
> is there a reason we're not catching 2/3/4-byte overlongs here? the old ProtoParserTools.readString ran a strict CharsetDecoder with onMalformedInput(REPORT) there's a comment on it saying that's deliberate so C0 80, E0 80 80, F0 80 80 80 all threw MalformedProtobufException and now they decode to U+0000. surrogates and >10FFFF are both still caught?
> 
> It's only the min-value checks that are missing: fromUTF8 at 358 / 365 / 370, and the same three in fromUTF8Tail at 283 / 297 / 306. protobuf-java rejects overlongs too, so this puts us out of step with anything parsing the same bytes downstream, and it means two different byte sequences decode to the same string.

it looks like this is a fix for this but its only on the ldintr-addPbjReadWriteTest branch

PR #907, opened 2026-08-31
feat: Adding PbjReader

ldintr:
New PbjReader class: varint/fixed-width/byte/boolean/string/bytes reads over a reusable ~16KB buffer, constructible from byte[]/ByteBuffer/Bytes/InputStream/ReadableSequentialData. Uses a sticky error code instead of throwing on every read; a system property lets callers trade real stack traces for shared exception instances. Adds supporting ParseException constructors and Bytes.resetPbjReader.

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
@@ -109,7 +109,7 @@ public class PbjReader implements AutoCloseable {
         if (seq != null) {
             rsd = seq;
             absoluteLimit = seq.limit();
-            offset = (int) seq.position();
+            offset = seq.position();
         } else if (inputStream != null) {
             stream = inputStream;
         }
@@ -129,7 +129,7 @@ public class PbjReader implements AutoCloseable {
             rsd = seq;
             stream = null;
             absoluteLimit = seq.limit();
-            offset = (int) seq.position();
+            offset = seq.position();
         } else if (inputStream != null) {
             rsd = null;
             stream = inputStream;
@@ -197,7 +197,7 @@ public class PbjReader implements AutoCloseable {
         resetWith(bb.array(), position, position + bb.remaining());
     }
 
-    private void bufferMore(int relAmount) {
+    private void bufferMore() {
         if (err != 0) return;
         offset += pos;
         if (pos < end && pos != 0) {
@@ -232,7 +232,7 @@ public class PbjReader implements AutoCloseable {
     // still small, but less likely to hit this case in steaming, and only once when not streaming
     private boolean hasRemainingInternal() {
         if (seenEOF) return false;
-        bufferMore(1);
+        bufferMore();
         return pos < relLimit;
     }
 
@@ -294,8 +294,8 @@ public class PbjReader implements AutoCloseable {
         count -= skippedInBuffer;
         pos = relLimit;
         offset += count;
-        if (stream != null) {
-            try {
+        try {
+            if (stream != null) {
                 long remaining = count;
                 while (remaining > 0) {
                     long skipped = stream.skip(remaining);
@@ -311,11 +311,11 @@ public class PbjReader implements AutoCloseable {
                 if (remaining != 0) {
                     setError(BUFFER_UNDERFLOW);
                 }
-            } catch (IOException e) {
-                setError(IO_ERROR);
+            } else {
+                rsd.skip(count);
             }
-        } else {
-            rsd.skip(count); // may throw
+        } catch (IOException e) {
+            setError(IO_ERROR);
         }
     }
 
@@ -432,7 +432,7 @@ public class PbjReader implements AutoCloseable {
         byte[] bytes = new byte[10];
         if (pos + 10 <= relLimit) {
             for (int i = 0; i < 10; i++) {
-                bytes[i] = readByte();
+                bytes[i] = buf[pos++];
                 if (bytes[i] >= 0) {
                     return Bytes.wrap(bytes, 0, i + 1);
                 }
@@ -592,7 +592,7 @@ public class PbjReader implements AutoCloseable {
 
     private int bufferedInternal(int count) {
         if (count <= buf.length) {
-            bufferMore(count);
+            bufferMore();
             if (pos + count <= relLimit) {
                 int origPos = pos;
                 pos += count;
@@ -772,7 +772,7 @@ public class PbjReader implements AutoCloseable {
     }
 
     private int readIntBEInternal() {
-        bufferMore(4);
+        bufferMore();
         if (pos + 4 > relLimit) {
             setError(BUFFER_UNDERFLOW);
             return 0;
@@ -786,7 +786,7 @@ public class PbjReader implements AutoCloseable {
     }
 
     private int readIntLEInternal() {
-        bufferMore(4);
+        bufferMore();
         if (pos + 4 > relLimit) {
             setError(BUFFER_UNDERFLOW);
             return 0;
@@ -827,7 +827,7 @@ public class PbjReader implements AutoCloseable {
     }
 
     private long readLongBEInternal() {
-        bufferMore(8);
+        bufferMore();
         if (pos + 8 > relLimit) {
             setError(BUFFER_UNDERFLOW);
             return 0;
@@ -859,7 +859,7 @@ public class PbjReader implements AutoCloseable {
     }
 
     private long readLongLEInternal() {
-        bufferMore(8);
+        bufferMore();
         if (pos + 8 > relLimit) {
             setError(BUFFER_UNDERFLOW);
             return 0;
@@ -946,7 +946,7 @@ public class PbjReader implements AutoCloseable {
 
     private byte readByteInternal() {
         if (pos + 1 > relLimit) {
-            bufferMore(1);
+            bufferMore();
             if (pos + 1 > relLimit) {
                 setError(BUFFER_UNDERFLOW);
                 return 0;
@@ -975,6 +975,7 @@ public class PbjReader implements AutoCloseable {
         if (bufPos >= 0) {
             data = buf;
         } else {
+            // larger than internal buffer
             data = new byte[length];
             int copiedLen = readBytesInternalCopy(data, 0, length);
             if (copiedLen < length) {
@@ -985,8 +986,12 @@ public class PbjReader implements AutoCloseable {
         }
 
         if (charArray == null || length > charArray.length) {
-            int power2Capacity = 2 << (63 - Long.numberOfLeadingZeros(Math.max(2048, length)));
-            charArray = new char[power2Capacity];
+            long power2Capacity = 2 << (63 - Long.numberOfLeadingZeros(Math.max(2048, length)));
+            if (power2Capacity < 0 || power2Capacity > Integer.MAX_VALUE) {
+                setError(ILLEGAL_ARGUMENT, "length larger than int max");
+                return "";
+            }
+            charArray = new char[(int) power2Capacity];
         }
 
         int i = 0;
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
@@ -1,0 +1,1024 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.pbj.runtime.io.buffer;
+
+import com.hedera.pbj.runtime.ParseException;
+import com.hedera.pbj.runtime.ProtoParserTools;
+import com.hedera.pbj.runtime.UnknownFieldException;
+import com.hedera.pbj.runtime.io.DataEncodingException;
+import com.hedera.pbj.runtime.io.ReadableSequentialData;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.BufferOverflowException;
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+
+/**
+ * A buffer-backed reader for decoding data. Not thread safe.
+ *
+ * <p>PbjReader maintains an internal byte buffer and reads from a {@link ReadableSequentialData},
+ * {@link InputStream}, or a byte array. When backed by a stream the internal buffer is refilled
+ * on demand as data is consumed. A stream will never be read past its limit.
+ *
+ * <p>Errors are tracked internally rather than thrown immediately. Call {@link #error()} to check
+ * for a pending error code, or {@link #throwOnError()} to surface it as a {@link ParseException}.
+ * Once an error is set it is sticky — subsequent reads return default values (zero or empty).
+ * EOF is not an error and will return 0 when error() is called.
+ * ByteBuffer using a direct buffer is not supported
+ */
+public class PbjReader implements AutoCloseable {
+    private byte[] buf;
+    private int pos, end;
+    private int relLimit, err;
+    private long absoluteLimit = Long.MAX_VALUE, offset;
+    private ReadableSequentialData rsd;
+    private InputStream stream;
+    private Exception cause;
+    private boolean seenEOF, includeCause;
+    private byte[] ownedBuf;
+
+    private char[] charArray;
+    private static final boolean useStacktrace =
+            !"false".equalsIgnoreCase(System.getProperty("pbj.ReaderWriter.useStackTrace"));
+
+    public static final int EOF = -1,
+            CLOSED = -2,
+            DATA_ENCODING = 1,
+            BUFFER_UNDERFLOW = 2,
+            PARSE = 3,
+            ILLEGAL_ARGUMENT = 4,
+            IO_ERROR = 5,
+            UNSUPPORTED = 6, // used with WIRE_TYPE_GROUP_START, WIRE_TYPE_GROUP_END
+            USAGE_ERROR = 9,
+            UNKNOWN_FIELD = 10,
+            BUFFER_OVERFLOW = 11,
+            MAX_DEPTH_REACHED = 12,
+            MALFORM_STRING = 13;
+
+    private static final UnknownFieldException premadeUnknown;
+    private static final BufferUnderflowException premadeUnderflow;
+    private static final BufferOverflowException premadeOverflow;
+    private static final RuntimeException premadeRuntime, premadeUnsupported;
+    private static final DataEncodingException premadeDataEncoding;
+    private static final IllegalArgumentException premadeIllegal;
+    private static final ParseException premadeParseEmpty, premadeParseUnknown, premadeMaxDepth;
+
+    static {
+        premadeUnknown = new UnknownFieldException("");
+        premadeUnderflow = new BufferUnderflowException();
+        premadeOverflow = new BufferOverflowException();
+        premadeRuntime = new RuntimeException();
+        premadeUnsupported = new RuntimeException("Hit an unsupported feature");
+        premadeDataEncoding = new DataEncodingException("");
+        premadeIllegal = new IllegalArgumentException("");
+
+        premadeParseEmpty = new ParseException("parse error");
+        premadeParseUnknown = new ParseException("parse error", premadeUnknown);
+        premadeMaxDepth = new ParseException("Reached maximum allowed depth");
+    }
+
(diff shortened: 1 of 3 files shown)


imalygin (changes requested):
Please, add meaningful PR description which includes motivation for this change and add unit tests for `PbjReader`.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1024 @@
...
+            int copiedLen = readBytesInternalCopy(data, 0, length);
+            if (copiedLen < length) {
+                setError(BUFFER_UNDERFLOW);
+                return "";
+            }
+            bufPos = 0;
+        }
+
+        if (charArray == null || length > charArray.length) {
+            int power2Capacity = 2 << (63 - Long.numberOfLeadingZeros(Math.max(2048, length)));
```
imalygin: For length `>= 2^30 (~1 GB)`, the highest set bit is at position 30, so this evaluates to `2 << 30 == Integer.MIN_VALUE`, and `new char[Integer.MIN_VALUE]` throws `NegativeArraySizeException`. 
Add validation for `maxSize`.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1024 @@
...
+    /**
+     * Returns an {@link InputStream} view of the remaining data in this reader.
+     * For streaming readers, returns the underlying stream directly. For byte-array readers,
+     * returns a {@link ByteArrayInputStream} over the buffered data.
+     *
+     * @return an {@code InputStream} over the unread bytes
+     * @throws UnsupportedOperationException if the reader is in a partially-buffered streaming state
+     */
+    public InputStream asInputStream() {
+        if (end == 0) return stream != null ? stream : rsd.asInputStream();
```
imalygin, line 920: For new `PbjReader(new byte[0])` (or `resetWith(new byte[0])`), `end == 0`, `stream == null`, `rsd == null` → `rsd.asInputStream()` throws NPE.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1024 @@
...
+        rsd = null;
+        stream = null;
+        construct(buffer, position, endPosition);
+    }
+
+    private void construct(ReadableSequentialData seq, InputStream inputStream) {
+        if (seq != null) {
+            rsd = seq;
+            absoluteLimit = seq.limit();
+            offset = (int) seq.position();
```
imalygin: `offset` is `long`. Is there any reason why cast ` seq.position()` to int?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: Writing this in a test shows that it seems to be fine. No exceptions in this
```
        PbjReader reader = new PbjReader(new byte[0]);
        assertEquals(false, reader.hasRemaining());
```

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1024 @@
...
+                    }
+                }
+                if (remaining != 0) {
+                    setError(BUFFER_UNDERFLOW);
+                }
+            } catch (IOException e) {
+                setError(IO_ERROR);
+            }
+        } else {
+            rsd.skip(count); // may throw
```
imalygin: The stream branch catches `IOException` and converts a short skip into `BUFFER_UNDERFLOW`; this branch is just `rsd.skip(count); // may throw`. Per the `ReadableSequentialData` contract, a short skip surfaces as an exception, which here propagates uncaught - violating the internal-error-tracking model that the rest of the class maintains. It also never sets `BUFFER_UNDERFLOW` on a partial `rsd` skip.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1024 @@
...
+        int utf16Len = ProtoParserTools.fromUTF8(charArray, data, bufPos, i, length);
+        if (utf16Len >= 0) {
+            return new String(charArray, 0, utf16Len);
+        }
+        setError(PbjReader.PARSE);
+        return "";
+    }
+
+    @Override
+    public void close() {
```
imalygin, line 1018: Only `stream` is closed; an `rsd` holding resources leaks. `err = CLOSED` maps to `error() == 0`, so post-close reads silently degrade to `BUFFER_UNDERFLOW` rather than signaling closed. Is it intentional?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1024 @@
...
+    public void resetWith(byte[] data) {
+        resetWith(data, 0, data.length);
+    }
+    /** Resets this reader to read from the given {@link ByteBuffer}. */
+    public void resetWith(ByteBuffer bb) {
+        int position = bb.arrayOffset() + bb.position();
+        resetWith(bb.array(), position, position + bb.remaining());
+    }
+
+    private void bufferMore(int relAmount) {
```
imalygin: `relAmount` is not used

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1024 @@
...
+     *
+     * @return the byte value
+     */
+    public byte readByte() {
+        if (pos + 1 <= relLimit) return buf[pos++];
+        return readByteInternal();
+    }
+
+    private byte readByteInternal() {
+        if (pos + 1 > relLimit) {
```
imalygin, line 948: This check is redundant, and probably the method as well, as it's used only once. Consider inlining it.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1024 @@
...
+     * Reads a base-128 varint and returns the raw encoded bytes without decoding them.
+     * Sets {@link #DATA_ENCODING} if the varint is malformed.
+     *
+     * @return the raw varint bytes, or {@link Bytes#EMPTY} on error
+     */
+    public Bytes readVarLongBytes() {
+        byte[] bytes = new byte[10];
+        if (pos + 10 <= relLimit) {
+            for (int i = 0; i < 10; i++) {
+                bytes[i] = readByte();
```
imalygin: If we assume that it's a hot path, using `readByte` is wasteful. Consider  usage of `buf[pos++]` instead, just as you did it in `readVarLongNoZZ`

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1024 @@
...
+        relLimit = (int) Math.min(absoluteLimit - offset, end);
+    }
+
+    /**
+     * Returns the current absolute read position, accounting for any bytes already consumed
+     * from the internal buffer plus any previously buffered data.
+     *
+     * @return the current read position
+     */
+    public long position() {
```
imalygin, line 269: The semantic of this method may be a bit unexpected for the client code, as `position()` means different things across reader kinds. For a stream/rsd reader it's the logical source position; for `new PbjReader(data, 5, 10)` or a `ByteBuffer` with nonzero `arrayOffset/position`, it returns the array-relative index (5, not 0).

I suppose you can define it as "absolute index into the backing array," but confirm that's intended, since a caller expecting a 0-based slice position will be surprised.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: Fixed with the below

```java
long power2Capacity = 2 << (63 - Long.numberOfLeadingZeros(Math.max(2048, length)));
if (power2Capacity < 0 || power2Capacity > Integer.MAX_VALUE) {
    setError(ILLEGAL_ARGUMENT, "length larger than int max");
    return "";
}

```

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: Fixed by removing the cast

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: Intentional to not inline, so optimizer is likely to optimize the fast path. I could skip the if check but its shouldn't be in the fast path and won't matter I don't think I'll be changing the implementation, but I also think it's fine to leave the extra if in

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: Fixed by making that change

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: Fixed.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: ReadableStreamingData doesn't expose any close or a way to dispose resource. I'd like to remove ReadableStreamingData from PbjReader in the future

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1029 @@
...
+        err = 0;
+        offset = 0;
+        cause = null;
+        includeCause = false;
+        rsd = null;
+        stream = null;
+        construct(buffer, position, endPosition);
+    }
+
+    private void construct(ReadableSequentialData seq, InputStream inputStream) {
```
artemananiev, line 108: Why is there a separate `construct()` for byte arrays, but one `construct()` for RSD and InputStream? Should there be three, one for each type? Should there be one with all possible input types?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1029 @@
...
+        construct(bb.array(), position, position + bb.remaining());
+    }
+
+    /** Creates a reader backed by the given {@link Bytes}. */
+    public PbjReader(Bytes bytes) {
+        bytes.resetPbjReader(this);
+    }
+
+    /** Resets this reader to read from the given {@link ReadableSequentialData}. */
+    public void resetWith(ReadableSequentialData seq) {
```
artemananiev, line 183: Do I get it right that the whole purpose of all these `resetWith()` methods is to avoid creating new PbjReader instances, but reusing existing objects instead?

As far as I can see, `resetWith()` calls `construct()`, which throws away all the state of the current PbjReader object and assigns all fields new values. This may be slower than to create a new object (it will be short-living, so the changes are it will be in the young GC generation). The only field not reassigned seems to be `ownedBuf`, is it what really drives the reuses?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1029 @@
...
+        resetWith(data, 0, data.length);
+    }
+    /** Resets this reader to read from the given {@link ByteBuffer}. */
+    public void resetWith(ByteBuffer bb) {
+        int position = bb.arrayOffset() + bb.position();
+        resetWith(bb.array(), position, position + bb.remaining());
+    }
+
+    private void bufferMore() {
+        if (err != 0) return;
```
artemananiev, line 201: This is (one of) the reason I proposed previously to get rid of error codes. If `err` is not zero, this means there was a problem in past. If exceptions were used instead of error codes, the stack would be immediately unwinded to the caller. No way the caller could forget to check the error code, they would be forced to catch (and handle) the exceptions.

For now, this just pollutes the code. And this is error prone.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1029 @@
...
+    /**
+     * Reads a base-128 varint and returns its zigzag-decoded value as an {@code int}.
+     * Zigzag decoding maps the raw unsigned value {@code n} to {@code (n >>> 1) ^ -(n & 1)}.
+     * On a malformed varint, sets the error flag and returns {@code -1}; however,
+     * {@code -1} is also a valid decoded value, so callers must check {@link #error()} to
+     * distinguish an error from a legitimate result.
+     *
+     * @return the decoded integer value
+     */
+    public int readVarIntZZ() {
```
artemananiev, line 356: Is it different from `readVarInt(true)`? If so, this method is completely redundant, just bloats the API

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1029 @@
...
+        return Bytes.EMPTY;
+    }
+
+    /**
+     * Records an error on this reader with no message. Once an error is recorded
+     * subsequent reads return default values (zero or empty).
+     *
+     * @param errorKind one of the error-code constants ({@link #DATA_ENCODING}, {@link #BUFFER_UNDERFLOW}, etc.)
+     */
+    public void setError(int errorKind) {
```
artemananiev, line 463: Should it really be public?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
imalygin: >Intentional to not inline, so optimizer is likely to optimize the fast path. 

Can you elaborate? I still don't see how having this method is beneficial.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: It does absolute positioning correctly

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: I use to have streams and buffers not mixed but realized it's better for both lib and users not to do that. I'll implement a fix today

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: The project I'm on is running into allocation issues. There's too many and the gc is doing a lot of work. Reusing helps them and if someone finds a situation where it's faster to use `new` they still can do it

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: I'm not sure if that's ever hit in practice. I don't mind the check so I left it in, but if you want it to be removed I can double check if it's possible to reach it (and ask AI). Personally I don't think an extra if statement, especially one on a rare path is bad (rare because it happens on the first read, and 16k after that)

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: It's the same but I was able to measure a 1% improvement by having a version that doesn't use an if statement. I made many changes at once in the 13th item and measured it. (idk why gh shows 12 for some of these) https://github.com/hashgraph/pbj/pull/919 if you really want to we can rerun the measurements but I believe we ran them 3 times already

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: Yes. parseNoEx (that gets introduce in codec in "Switched to PbjReader/PbjWriter" hash 71e5a9d4f) returns null on error and error code is needed to understand what the error is, especially if trying to avoid exceptions (the throwOnError alternative).

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: I looked at it again, the ones with bytes are completely different than the stream that there'd be no overlap. The two streams are very similar but I have no problem separating them

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
artemananiev: Here are my benchmarks:
```java
@Benchmark
    public void testReadVarLongBoolean(final Blackhole blackhole) {
        long sum = 0;
        try (final PbjReader in = new PbjReader(array)) {
            for (int i = 0; i < size; i++) {
                if (i % 2 == 0) {
                    sum += in.readVarLong(true);
                } else {
                    sum += in.readVarLong(false);
                }
            }
        }
        blackhole.consume(sum);
    }

    @Benchmark
    public void testReadVarLongZZ(final Blackhole blackhole) {
        long sum = 0;
        try (final PbjReader in = new PbjReader(array)) {
            for (int i = 0; i < size; i++) {
                if (i % 2 == 0) {
                    sum += in.readVarLongZZ();
                } else {
                    sum += in.readVarLongNoZZ();
                }
            }
        }
        blackhole.consume(sum);
    }
```

And here is the result:

```
Benchmark                                      (size)   Mode  Cnt      Score      Error  Units
PbjReaderReadVarLongZZ.testReadVarLongBoolean   10000  thrpt    5  28512.090 ± 2278.525  ops/s
PbjReaderReadVarLongZZ.testReadVarLongZZ        10000  thrpt    5  28345.451 ± 3432.496  ops/s
```

ldintr: Thats a big error margin.

The way I've been measuring was using EverythingBench.parsePbjReader which is like the other parse benchmark except I used PbjReader directly. I measured before and after applying the commit. It was about a 1% difference and I suspect the reason was because some functions were already large enough that the optimizer didn't want to check that readVarLong was tiny and could remove the if statement with little work.

Alternatively, maybe it didn't inline it because it noticed it was called alot and didn't check how small it was for some reason or another

PR #909, opened 2026-08-31
feat: Adding PbjWriter

ldintr:
Write-side counterpart to PbjReader: fixed-width/varint(+zigzag) writes, a placeholder/patch mechanism (placehold/writeAtUnsafe/reinsertVarInt) for filling in length prefixes after writing content (avoiding a two-pass encode for short strings), constructible from OutputStream/ByteBuffer/growable buffers/WritableSequentialData, with toByteArray()/toPbjReader() round-tripping. Makes sizeOfStringNoTag public for the large-string fallback.

### DIFF
--- a/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjWriter.java
+++ b/pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjWriter.java
@@ -1,0 +1,1067 @@
+// SPDX-License-Identifier: Apache-2.0
+package com.hedera.pbj.runtime.io.buffer;
+
+import static java.lang.Character.MAX_SURROGATE;
+import static java.lang.Character.MIN_SURROGATE;
+import static java.lang.Character.isSurrogatePair;
+import static java.lang.Character.toCodePoint;
+
+import com.hedera.pbj.runtime.ProtoWriterTools;
+import com.hedera.pbj.runtime.io.WritableSequentialData;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.UncheckedIOException;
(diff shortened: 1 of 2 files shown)


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjWriter.java
```diff
@@ -0,0 +1,1067 @@
...
+    }
+
+    private void writeIntLEInternal(int value) {
+        flushOrGrow(4);
+        buf[pos] = (byte) value;
+        buf[pos + 1] = (byte) (value >>> 8);
+        buf[pos + 2] = (byte) (value >>> 16);
+        buf[pos + 3] = (byte) (value >>> 24);
+        pos += 4;
+    }
```
jasperpotts, line 530: I don't see where `writeIntLEInternal()` is used other than in  `writeIntLE()` so why not just have:
```
    public void writeIntLE(int value) {
        if (pos + 4 > cap) flushOrGrow(4);
        buf[pos] = (byte) value;
        buf[pos + 1] = (byte) (value >>> 8);
        buf[pos + 2] = (byte) (value >>> 16);
        buf[pos + 3] = (byte) (value >>> 24);
        pos += 4;
    }
```
?

ldintr: There's a lot of functions I written like this, especially the ones used in a hotloop.

Essentially the optimizer will notice that flushOrGrow is rarely called (not hot) and be very likely to inline this function even if it isn't inlining flushOrGrow. Now that I'm looking at this again, the patterns elsewhere I used were `if (veryLikely) { ... return result; } return funcNameInternal();`. I don't actually remember if there's a difference between the two patterns and if I should change this to the other variant. I suspect I should for consistency reasons

PR #946, opened 2026-09-21
feat: Switches from interface to the new PBJ reader/writer code

ldintr:
**Description**:

Switches from interface to the new PBJ reader/writer code

**Notes for reviewer**:

ProtoParserTools.java and ProtoWriter.java has a lot of copy/paste functions. I put them next to eachother so you can see they're the same

**Checklist**


pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ParseException.java
```diff
@@ -16,4 +16,12 @@ public ParseException(Throwable cause) {
...
         super(message);
     }
+
+    public ParseException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    public ParseException(Throwable cause, boolean skipStackTag) {
+        super(null, cause, true, false);
+    }
```
anthony-swirldslabs, line 26: The `skipStackTag` argument is unused.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1110 @@
...
+ * {@link InputStream}, or a byte array. When backed by a stream the internal buffer is refilled
+ * on demand as data is consumed. A stream will never be read past its limit.
+ *
+ * <p>Errors are tracked internally rather than thrown immediately. Call {@link #error()} to check
+ * for a pending error code, or {@link #throwOnError()} to surface it as a {@link ParseException}.
+ * Once an error is set it is sticky — subsequent reads return default values (zero or empty).
+ * EOF is not an error and will return 0 when error() is called.
+ * ByteBuffer using a direct buffer is not supported
+ */
+public class PbjReader implements AutoCloseable {
```
anthony-swirldslabs, line 30: Since you ended up implementing the AutoCloseable interface here, couldn't the PbjReader class also implement the `ReadableSequentialData` interface, and thus avoid any API changes in the Codec interface and related changes in the codec generator?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1110 @@
...
+            MAX_DEPTH_REACHED = 12,
+            MALFORM_STRING = 13;
+
+    private static final UnknownFieldException premadeUnknown;
+    private static final BufferUnderflowException premadeUnderflow;
+    private static final BufferOverflowException premadeOverflow;
+    private static final RuntimeException premadeRuntime, premadeUnsupported;
+    private static final DataEncodingException premadeDataEncoding;
+    private static final IllegalArgumentException premadeIllegal;
+    private static final ParseException premadeParseEmpty, premadeParseUnknown, premadeMaxDepth;
```
anthony-swirldslabs: Minor: all static fields should be UPPER_SNAKE_CASE.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1110 @@
...
+     */
+    private void construct(ReadableSequentialData seq, InputStream inputStream) {
+        if (seq != null) {
+            rsd = seq;
+            absoluteLimit = seq.limit();
+            offset = (int) seq.position();
+        } else if (inputStream != null) {
+            stream = inputStream;
+        }
+        ownedBuf = buf = new byte[16 << 10]; // 16k is friendly to x86-64 L1 cache
```
anthony-swirldslabs: Minor: the number may be friendly to the hardware. But it may or may not be friendly to the typical sizes of objects being parsed. E.g. if the vast majority of models occupy 256 - 4096 bytes, then we're wasting quite a bit of memory on this buffer. You may want to run a performance testing in a real Consensus Node environment to find the best size.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1110 @@
...
+            absoluteLimit = seq.limit();
+            offset = seq.position();
+        } else if (inputStream != null) {
+            rsd = null;
+            stream = inputStream;
+            absoluteLimit = Long.MAX_VALUE;
+            offset = 0;
+        }
+        if (ownedBuf == null) {
+            ownedBuf = new byte[16 << 10]; // 16k is friendly to x86-64 L1 cache
```
anthony-swirldslabs: Minor: might want to extract the buffer size magic number into a private static final constant.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1110 @@
...
+        int position = bb.arrayOffset() + bb.position();
+        resetWith(bb.array(), position, position + bb.remaining());
+    }
+    /** Resets this reader to read from the given {@link Bytes}. */
+    public void resetWith(Bytes bytes) {
+        bytes.resetPbjReader(this);
+    }
+
+    /*
+     * fills the buffer from the stream repecting the set limit
```
anthony-swirldslabs: nit: "respecting"

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1110 @@
...
+        if (useStacktrace) {
+            if (errorKind == UNKNOWN_FIELD) {
+                cause = new UnknownFieldException(message);
+            } else if (errorKind == BUFFER_UNDERFLOW) {
+                cause = new BufferUnderflowException();
+            } else if (errorKind == BUFFER_OVERFLOW) {
+                cause = new BufferOverflowException();
+            } else if (errorKind == PARSE) {
+                cause = new RuntimeException(message);
+                includeCause = false;
```
anthony-swirldslabs, line 548: I'm still concerned about `includeCause = false` at line 546 here. I believe this leads to hiding the message and the stack trace that we're attempting to preserve at line 545 because the cause will never be reported to the client because of the flag set to false.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
```diff
@@ -0,0 +1,1110 @@
...
+     */
+    private ParseException throwOnErrorImpl() throws ParseException {
+        if (useStacktrace) {
+            switch (err) {
+                case DATA_ENCODING:
+                    throw new DataEncodingException("throwOnError", cause);
+                case ILLEGAL_ARGUMENT:
+                    throw new IllegalArgumentException("throwOnError", cause);
+                case UNSUPPORTED:
+                    throw new RuntimeException("Hit an unsupported feature", cause);
```
anthony-swirldslabs: There's an UnsupportedOperationException in JDK. I think it would be more appropriate here.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjWriter.java
```diff
@@ -0,0 +1,1093 @@
...
+ * buffer is flushed automatically when full. When not backed by a stream the buffer grows as
+ * needed (unless constructed with {@code mayGrow = false}).
+ *
+ * <p>Errors are tracked internally rather than thrown immediately. Call {@link #error()} to check
+ * for a pending error code, or {@link #throwOnError()} to surface it as an exception. Once an
+ * error is set it is sticky — subsequent write calls become no-ops.
+ *
+ * <p>Implements {@link AutoCloseable}: closing flushes pending bytes to the underlying stream.
+ */
+public class PbjWriter implements AutoCloseable {
```
anthony-swirldslabs, line 31: Similar to the PbjReader above, since we're implementing an interface here, couldn't we also implement the `WritableSequentialData` and avoid any API changes in the Codec?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjWriter.java
```diff
@@ -0,0 +1,1093 @@
...
+    private int offset, err;
+    private RuntimeException cause;
+    private OutputStream output;
+    private boolean mayGrow = true;
+
+    /*
+     * If a project doesn't care about stacktraces, setting pbj.ReaderWriter.useStackTrace to false
+     * will throw a premade exception that doesn't have the correct stacktrace. It's fast and good against DOS attacks
+     */
+    private static final boolean useStacktrace =
```
anthony-swirldslabs: Minor: static fields should be UPPER_SNAKE_CASE. This comment applies to line 60 below as well.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjWriter.java
```diff
@@ -0,0 +1,1093 @@
...
+        });
+    }
+
+    /**
+     * Returns the raw internal byte array. The valid data occupies indices {@code [0, position())}.
+     * Intended for low-level inspection; prefer {@link #toByteArray()} for a correctly sized copy
+     *
+     * @return the internal buffer array
+     */
+    public byte[] internalArray() {
```
anthony-swirldslabs: Minor: `arrayUnsafe()` ?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -86,10 +169,17 @@ public final T parse(
             int maxDepth,
             int maxSize)
             throws ParseException {
-        return parseImpl(input, strictMode, parseUnknownFields, maxDepth, maxSize);
+        final PbjReader reader = new PbjReader(input);
+        try {
+            return parse(reader, strictMode, parseUnknownFields, maxDepth, maxSize);
+        } finally {
+            syncPosition(input, reader);
+        }
```
anthony-swirldslabs, line 175: This is a tricky piece of logic.
1. What is someone implements a buffered SequentialData that doesn't inherit from the BufferedSequentialData ? How would they know to come here and modify the logic?
2. Why wouldn't the PbjReader maintain the positions in sync at all times?
3. Why would someone want to buffer from an already buffered SequentialData?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -161,10 +253,12 @@ public final T parse(@NonNull ReadableSequentialData input, final boolean strict
      */
     @NonNull
     public final T parse(@NonNull Bytes bytes, final boolean strictMode, final int maxDepth) throws ParseException {
-        return parse(bytes.toReadableSequentialData(), strictMode, maxDepth);
+        return parse(new PbjReader(bytes), strictMode, false, maxDepth, DEFAULT_MAX_SIZE);
     }
 
     /**
+     * Temporary for test compatibility
```
anthony-swirldslabs, line 258: Minor: if production code in CN/BN never uses it, then might as well nuke the method and refactor the tests to use another overloaded version of the method.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -81,6 +82,16 @@ public static int readInt32(final ReadableSequentialData input) {
...
     }
 
+    /**
+     * Read a protobuf int32 from input
+     *
+     * @param input The input data to read from
+     * @return the read int
+     */
+    public static int readInt32(PbjReader input) {
+        return input.readVarInt(false);
```
anthony-swirldslabs, line 92: Minor: couldn't callers call input.read*() directly? Could we avoid adding all these call-inverting utility wrappers into the ProtoParserTools class?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
```diff
@@ -126,6 +167,20 @@ public static boolean readBool(final ReadableSequentialData input) throws IOExce
...
+     *
+     * @param input The input data to read from
+     * @return the read boolean
+     */
+    public static boolean readBool(PbjReader input) {
+        final var i = input.readVarInt(false);
+        if (i != 1 && i != 0) {
+            input.setError(PbjReader.DATA_ENCODING);
+        }
+        return i == 1;
```
anthony-swirldslabs, line 181: Especially here, callers calling into the PbjReader directly would be a lot more appropriate than having this helper method here. This error-code-setting-logic truly belongs inside the PbjReader itself and not here. I'd go as far as making setError private.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -132,6 +165,47 @@ assert switch (field.type()) {
...
+
+        if (skipDefault && !field.oneOf() && value == 0) {
+            return;
+        }
+        switch (field.type()) {
+            case INT32 -> {
+                writeTag(out, field, WIRE_TYPE_VARINT_OR_ZIGZAG);
+                out.writeVarInt(value, false);
+            }
+            case UINT32 -> {
```
anthony-swirldslabs, line 192: It's odd to use a switch expression that doesn't yield any value. It's especially confusing because it's unclear if we're having a fall-through at lines 191-192 or not. Could you please rewrite this using the regular switch/case: with proper breaks instead?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -180,6 +265,43 @@ assert switch (field.type()) {
...
+        assert !field.repeated() : "Use writeLongList with repeated types";
+        if (skipDefault && !field.oneOf() && value == 0) {
+            return;
+        }
+        switch (field.type()) {
+            case INT64, UINT64 -> {
+                writeTag(out, field, WIRE_TYPE_VARINT_OR_ZIGZAG);
+                out.writeVarLong(value, false);
+            }
+            case SINT64 -> {
```
anthony-swirldslabs, line 291: Same as above, please consider using a regular switch/case: with breaks here, for the same reasons as above.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
```diff
@@ -1340,7 +2205,7 @@ public static int sizeOfString(FieldDefinition field, String value, boolean skip
      * @param value string value to get encoded size for
      * @return the number of bytes for encoded value
      */
-    static int sizeOfStringNoTag(String value) {
+    public static int sizeOfStringNoTag(String value) {
```
anthony-swirldslabs, line 2208: This class/file is already very long. I'd recommend creating a new ProtoPbjWriterTools class and adding PbjWriter-specific helpers in there.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ParseException.java
ldintr: I can delete it if you want. Its only used when "pbj.ReaderWriter.useStackTrace" is set to false. Allegedly it skips building the stack but I don't believe AI, below is what it claims. I named the parameter 'Tag' because people use that to say the parameter is only to give the function a unique signature. It's intentionally unused. Since I don't believe AI I should delete this but I may measure instead unless you want me to delete it regardless
 
> I tested it directly rather than going from memory. The claim holds up:
> 
> length=0
> T$1: msg
> ---
> after explicit fillInStackTrace length=0
> 
> - getStackTrace() returns a zero-length array when writableStackTrace=false.
> - printStackTrace() prints only the message line — no at ... frames.
> - Even an explicit call to fillInStackTrace() afterward is a no-op — once writableStackTrace=false, the instance can never get a stack trace, by construction or later mutation.
> 
> This matches the actual OpenJDK Throwable constructor logic:
> public Throwable(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace) {
>     fillInStackTrace = writableStackTrace;   // gate
>     detailMessage = message;
>     this.cause = cause;
>     if (!enableSuppression) suppressedExceptions = null;
>     if (fillInStackTrace) {
>         fillInStackTrace();   // skipped when writableStackTrace == false
>     }
> }
> 
> So for ParseException's (Throwable, boolean) constructor — super(null, cause, true, false) — the stack trace genuinely is never captured, and that's the cost-saving mechanism the hot parse paths in PbjReader rely on.
>

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: No, the speed comes from changing the signature to the class. The fact the class implements an interface doesn't really matter.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
ldintr: Should I find the code that requires this in consensus node?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjWriter.java
ldintr: Same answer

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjWriter.java
ldintr: idr if a later commit uses this, but I'll delete this now bc none might

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: This was mostly for test compatibility before deleting RSD from codec. What should I do in the meantime?

1. I can't imagine anyone implementing more RSD when pbj is merged in
2. It'll slow down and make pbj reader uglier, only to be deleted eventually. Also it doesn't make sense to not use codec to parse the object
3. You all had me remove logic that peeks into objects to access their internal buffer. The public API forces a copy. The signature change forces PbjReader/Writer to be used

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: I agree, but when? Should this be after this merge be accepted? Next time I work on CN? It seems like CN takes a while to accept changes, I haven't worked on BN yet

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
ldintr: I decided to be consistent, which should make reviewing easier. I like the idea of deleting these. When should I delete these?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
ldintr: That might be do-able if we can delete functions from ProtoParserTools.java

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
ldintr: This is a copy-paste of the function on line 134. I could update both but I rather not touch them

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
ldintr: Same answer

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
ldintr: I rather delete code. In this case, all I did was add public

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ParseException.java
anthony-swirldslabs: People add dummy params to differentiate between different overloaded versions of a single method and/or constructor. ParseException currently does not have a single `public ParseException(Throwable cause)` defined. That would be the very first constructor with this signature. Since there's no other constructors accepting only a cause, there's no need to add dummy parameters to the method. Please remove it.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
```diff
@@ -77,41 +77,35 @@ static String generateParseMethod(
...
-                 * @param maxDepth a ParseException will be thrown if the depth of nested messages exceeds the maxDepth value.
-                 * @param maxSize a ParseException will be thrown if the size of a delimited field exceeds the limit
-                 * @return Parsed $modelClassName model object or null if data input was null or empty
-                 * @throws ParseException If parsing fails
+                 * @param maxDepth sets the {@link PbjReader#MAX_DEPTH_REACHED} error if the depth of nested messages exceeds it.
+                 * @param maxSize sets the {@link PbjReader#PARSE} error if the size of a delimited field exceeds the limit
+                 * @return Parsed $modelClassName model object, or {@code null} if an error was set on {@code input}
                  */
-                protected final @NonNull $modelClassName parseImpl(
+                protected final $modelClassName parseImpl(
```
anthony-swirldslabs, line 84: That wasn't a part of the fix that I reviewed yesterday. Why is this change pushed into this PR? Was the PR incomplete yesterday? If so, could you please make it Draft, complete all the work, and then request reviews? If it was complete, could you please avoid adding new logic/API changes and postpone these additions to a future, separate PR?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ParseException.java
ldintr: Actually... thats the first overload. This is why I never use gh to review code. It makes things harder
```java
public class ParseException extends Exception {
    public ParseException(Throwable cause) {
        super(cause);
    }

    public ParseException(String message) {
        super(message);
    }

    public ParseException(String message, Throwable cause) {
        super(message, cause);
    }

    public ParseException(Throwable cause, boolean skipStackTag) {
        super(null, cause, true, false);
    } 
}
```

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/io/buffer/PbjReader.java
anthony-swirldslabs: "Requires" may be too strong a word. Nothing in the business logic in there should require an absence of a useful exception message. If there's a test in CN that checks it, then it should be a one-liner very quick PR to fix it. Please do fix it in CN rather than swallow the message in PBJ.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
anthony-swirldslabs: Why not make a quick small change in CN to use another overload first, and then proceed with this PBJ PR? This way we avoid committing "dead" code or introduce "features" to support dummy use-cases. We never want to do any of this.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
anthony-swirldslabs: I'd prefer to do this in this PR.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
anthony-swirldslabs: I'd prefer to rewrite at least this new one, but ideally both at once.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ParseException.java
anthony-swirldslabs: Oh, it was just me being lazy a bit. I should've expanded the code above. :) Thanks for confirming. That looks good then.

pbj-core/pbj-compiler/src/main/java/com/hedera/pbj/compiler/impl/generators/protobuf/CodecParseMethodGenerator.java
artemananiev: +1

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -37,18 +38,98 @@ public abstract class Codec<T> {
      * consistent entry point. Subclasses implement this method rather than {@code parse} directly, since
      * {@code parse} may perform additional work before and after delegating to this implementation.
      *
+     * The only difference is it's allowed to return null, the parse overloads call throwOnError.
```
artemananiev, line 41: Please, update it, so `parseImpl()` can only return a null, if an error is occurred. Without errors, it should not return nulls

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -37,18 +38,98 @@ public abstract class Codec<T> {
...
+     *
      * @see #parse(ReadableSequentialData, boolean, boolean, int, int) for a description
```
artemananiev, line 43: This link needs to be updated

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -37,18 +38,98 @@ public abstract class Codec<T> {
...
+     * @param maxSize a ParseException will be thrown if the size of a delimited field exceeds the limit
+     * @return The parsed object. It must not return null.
+     * @throws ParseException If parsing fails
+     */
+    @NonNull
+    public final T parse(
+            @NonNull PbjReader input, boolean strictMode, boolean parseUnknownFields, int maxDepth, int maxSize)
+            throws ParseException {
+        T res = parseImpl(input, strictMode, parseUnknownFields, maxDepth, maxSize);
+        input.throwOnError();
```
artemananiev, line 83: A null check for `res` is needed here

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -37,18 +38,98 @@ public abstract class Codec<T> {
...
+        return res;
+    }
+
+    /**
+     * Same as parse, except doesn't throw. Check for error by using input.error() != 0 (or > 0), or input.ok()
+     * Return value may be null
+     */
+    public final T parseNoEx(
```
artemananiev, line 91: Is this method really needed? What are use cases, when the `parse()` method above would not work / would be slow, so the caller needs to check for the errors manually?

To me, this method bloats the API and makes it error prone (the caller doesn't check for errors)

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
artemananiev: This issue is not specific to codecs, it's about how PbjReader reads data from ReadableSequentialData.

From RSD perspective, the position is already forwarded to how many bytes have been buffered by PbjReader. This is right and should not be changed. The PbjReader object is who owns the buffered bytes now. These bytes must be completely processed (read using PbjReader methods) before the PbjReader instance can be reused for a different input. Otherwise some bytes will be lost.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -225,19 +319,45 @@ public final T parseStrict(@NonNull Bytes bytes) throws ParseException {
...
+    }
+
+    /**
+     * Writes an item to the given {@link PbjWriter}.
+     *
+     * @param item The item to write. Must not be null.
+     * @param output The {@link PbjWriter} to write to.
      */
-    protected abstract void writeImpl(@NonNull T item, @NonNull WritableSequentialData output) throws IOException;
+    public final void writeNoEx(@NonNull T item, @NonNull PbjWriter output) {
```
artemananiev, line 343: Same question as about `parseNoEx()`

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -225,19 +319,45 @@ public final T parseStrict(@NonNull Bytes bytes) throws ParseException {
...
      * @param item The item to write. Must not be null.
      * @param output The {@link WritableSequentialData} to write to.
      * @throws IOException If the {@link WritableSequentialData} cannot be written to.
      */
-    public final void write(@NonNull T item, @NonNull WritableSequentialData output) throws IOException {
-        writeImpl(item, output);
+    public void write(@NonNull T item, @NonNull WritableSequentialData output) throws IOException {
+        final PbjWriter writer = new PbjWriter(output);
+        writeImpl(item, writer);
+        writer.flush();
```
artemananiev, line 359: When `write(T, PbjWriter)` is called, whose responsibility is to call PbjWriter.flush()? At the very least, it must be in the JavaDoc

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
artemananiev: I second to Anthony, let's have CN (Hedera and other known apps) code cleaned up first to remove usages of all Codec.write/parse overloads, then drop these methods from Codec, then migrate Codec to PbjReader/writer

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -262,18 +382,18 @@ public final int write(@NonNull T item, @NonNull byte[] output, final int startO
      * The actual byte-array writing logic for a specific codec, invoked by the {@link #write(Object, byte[], int)}
      * methods through a single, consistent entry point. Generated codecs override this method with a
      * performance-focused implementation; this default implementation delegates to
-     * {@link #writeImpl(Object, WritableSequentialData)} so that hand-written codecs don't need to provide one.
+     * {@link #writeImpl(Object, PbjWriter)} so that hand-written codecs don't need to provide one.
      *
      * @see #write(Object, byte[], int) for a description
      */
     protected int writeImpl(@NonNull T item, @NonNull byte[] output, final int startOffset) {
```
artemananiev, line 389: It would be great to have all write/writeImpl methods consistent: either return void, or return the number of bytes written

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -262,18 +382,18 @@ public final int write(@NonNull T item, @NonNull byte[] output, final int startO
...
-        final BufferedData bufferedData = BufferedData.wrap(output, startOffset, output.length - startOffset);
+        final PbjWriter writer = new PbjWriter(output, startOffset);
         try {
-            write(item, bufferedData);
-        } catch (IOException e) {
-            throw new UncheckedIOException(e);
+            writeImpl(item, writer);
+        } finally {
```
artemananiev, line 393: Why is 'flush()' in the "finally" block? Should the writer be checked for errors?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
```diff
@@ -285,7 +405,31 @@ protected int writeImpl(@NonNull T item, @NonNull byte[] output, final int start
      * @return The length of the data item in the input
      * @throws ParseException If parsing fails
      */
-    public abstract int measure(@NonNull ReadableSequentialData input) throws ParseException;
+    public abstract int measure(@NonNull PbjReader input) throws ParseException;
```
artemananiev, line 408: I'd like to see this method removed altogether. As far as I know, it isn't used anywhere in CN (please, double check)

artemananiev: Not as a part of this PR, of course

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
artemananiev, line 29: Why are ProtoParserTools / ProtoWriterTools included? They add 1200+ lines (pretty straightforward, but still) to the current PR

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
artemananiev: I agree, these methods (both old versions with ReadableSequentialData, and new ones with PbjReader) look redundant. I'd love to have them removed. However:

1. It should be done in a separate PR, since not directly related to PbjReader/Writer

2. Methods like ProtoParserTools.readInt32() are used by generated codecs. This means, codec generators should be updated first to call input.readVarInt(false) instead of readInt32(input), then the methods can be removed

ldintr: I agree with another PR. But perhaps after deciding what to do with test? We'll need to talk about porting/rewriting/removing tests. IMO I should port as many as possible after there's no more suggestions/changes to the interface

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoWriterTools.java
ldintr: My vote is in a new PR after we decide on tests, does that sound like the best time to do it to you?

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: Could we skip it since parseImpl only returns null on error and throwOnError will throw before res is used/looked at? Seems redundant to add it after the throwOnError

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: There's a few Codec implementations in other projects that call parse on an object (in their parseImpl iirc), which seems weird to me. Since the signature doesn't state it throws a parse exception my choices were use this or add a try/catch

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: This I think I can remove. I rather remove it in a commit where we remove parseNoEx for consistency reasons but I think we should have a parseNoEx

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: I was following preexisting code. It looks like it's only doing it here because it's writing to an output buffer. I don't actually remember seeing anything using this. Should I try to remove this instead? (possibly in another PR?) I see this being used in ProtoArrayWriterTools.writeMessageNoChecks,  ProtoArrayWriterTools may need a refactor

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: I never liked that function. PbjWriter writer is created INSIDE the function (without tls too), and isn't returned. The signature is returning int so the only option for error reporting is either change the callers or to throw, which flush does on error

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/Codec.java
ldintr: Looks like you did it already so I'm marking resolved.

pbj-core/pbj-runtime/src/main/java/com/hedera/pbj/runtime/ProtoParserTools.java
ldintr: Generated code uses a lot of those utilities. For example MessageInProtoPackageProtoCodec calls both readBytes and writeBytes. Briefly looking at EverythingProtoCodec, it seems to calling most of those util functions

